PR #207, opened 2019-05-16
Removed Initializer.create_database_indexes() from main.py

rgaudin:
main.py can't start as it is calling a method that has been removed in 8d9c607dc93b


rgaudin:
Ah that explains how this slipped in!

Actually `main.py` is referenced in `uwsgi.py` so it is used by `supervisord` but its `__main__` is not.
Useful for development though.

PR #211, opened 2019-05-19
task revoked events

automactic:
## Rationale

Save revoked events to mongo and update TaskStatus/Events enum on `backend` and `monitor` container.

## Changes

- updated TaskStatus enum, the mongo document validator is updated to include extra values as a side effect
- added revoked event support to monitor


automactic:
Ah, good catch! I should. Knew I was forgetting something

PR #215, opened 2019-05-21
[API] Additional filtering options for schedules 

rgaudin:
## Rationale

We need additional filtering on the schedules list to improve the U.I

## Changes

* `&lang=<code>` to filter by language code
* `&tag=<atag>&tag=<atag>` to filter via tags. Tags are combinative ($all)
* `&name=<astring>` to filter via schedule name. Accepts regex.

---

* All those filters are combinative: the more you set, the more precise the results.
* Added a `count` parameter to the `meta` key with the total (unbound to `limit`) number of results.
* Added some tests for those filters.

⚠️  `name` from the request is passed to mongo for the regex directly. It seems that mongo is doing a good job at escaping its value. I couln't inject anything this way but let me know if this is OK security wise.

### DIFF
--- a/dispatcher/backend/src/routes/schedules/schedule.py
+++ b/dispatcher/backend/src/routes/schedules/schedule.py
@@ -42,8 +42,6 @@ class SchedulesRoute(BaseRoute):
         if name:
             query['name'] = {'$regex': r".*{}.*".format(name), '$options': 'i'}
 
-        print(query)
-
         # get schedules from database
         projection = {
             '_id': 1,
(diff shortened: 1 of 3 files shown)


dispatcher/backend/src/routes/schedules/schedule.py
```diff
@@ -27,11 +27,20 @@ def get(self, *args, **kwargs):
         skip = 0 if skip < 0 else skip
         limit = 20 if limit <= 0 else limit
         categories = request.args.getlist('category')
+        tags = request.args.getlist('tag')
+        lang = request.args.get('lang')
```
automactic: Personally, I would spell out `language`. But it is up to you.

kelson42:
@automactic It is easier if the reviewer - if OK with the PR - just to the following things after:
* Merge the PR
* Delete the git branch
* Close the corresponding ticket if existing (and still not closed).

So, if OK for you, please just merge.

PR #216, opened 2019-05-23
Small improvements to schedules filters

rgaudin:
## Rationale

* We need to be able to filter by queue
* We need to be able to combine languages in filter (union)
* Having a central definition of queues, categories and languages is easier to maintain.

⚠️ the entities part is tied to #217. I've been back and forth about those. Should they each get an endpoint? Should they share an `entities` or `metadata` endpoint? Eventually, I went with this static list that is copied-on-build on the Typescript side: it's simpler, it's faster (less request) and I couldn't get it to work async with angular (my bad). Open to do it another way though.

## Changes

- lang param can be a list (union)
- new queue filter param

Improvements over the `entities` module:
- full list of categories
- list of queues
- list of languages with a dedicated module.
- usage of those entities in validators
- tests for new filters
- additional multi-filters test


dispatcher/backend/src/common/entities.py
```diff
@@ -15,9 +18,62 @@ def all(cls):
...
+    large = 'large'
+    debug = 'debug'
 
     @classmethod
     def all(cls):
-        return [cls.wikipedia, cls.phet]
+        return [cls.small, cls.medium, cls.large, cls.debug]
+
+
+class ScheduleLanguage:
```
automactic: This class is not used. I personally think it is generally a good idea to add then until you need them.

dispatcher/backend/src/common/languages.py
```diff
@@ -0,0 +1,337 @@
+""" static list of languages supported by zimfarm
```
automactic: Ummm... Is it really necessary to list all languages here? Also, once we add this here, we'd have to maintain it to be up to date. 
I think the one and only central place to store languages we support should be the database. We can do a `$group` to get it.

rgaudin: Agrees. As I mentioned, I started this way and came back to this to ease UI integration.
Will update the PR.

rgaudin:
Implemented the requested changes. thanks.

dispatcher/backend/src/common/entities.py
```diff
@@ -15,9 +16,62 @@ def all(cls):
...
+    medium = 'medium'
+    large = 'large'
+    debug = 'debug'
+
+    @classmethod
+    def all(cls):
+        return [cls.small, cls.medium, cls.large, cls.debug]
+
+
+class ScheduleLanguage:
```
automactic: Should this class be removed?

dispatcher/backend/src/routes/languages/language.py
```diff
@@ -0,0 +1,127 @@
...
+from routes.languages.base import LanguageQueryMixin
+
+
+class LanguagesRoute(BaseRoute):
+    rule = '/'
+    name = 'languages'
+    methods = ['POST', 'GET']
+
+    @authenticate
+    def post(self, *args, **kwargs):
```
automactic: Sorry I think language should be a read only thing. Its value should come from the schedules. I will explain more in the main comment of the review.

dispatcher/backend/src/common/entities.py
rgaudin: 🧐 pretty sure I removed it ! my bad.

rgaudin:
Well I wondered about which of those options to go with but figured that with it readonly you can't manage it.
We don't want to end up with users facing a choice of `Malayalam (മലയാളം)` or `Malayalam (മോണക്ക് അറിയാം)` because both were used or even worse offer a unique choice with the wrong one.

People make mistakes, especially on this because it's tricky. the full languages API allows us to build up and fix over time a central list of unique, proper languages.


The `count` property is not language related but it make no sense to omit it when offering a paginated list of items. That's why I added it to schedules as well.

rgaudin:
here you go.

PR #217, opened 2019-05-23
[UI] Schedules filters 

rgaudin:
## Rationale

Simple implementation of angular filters for the schedules view:

- name (to be entered)
- queues (to tick, union)
- languages (to tick, union)
- categories (to tick, union)
- tags (to be entered, intersection)

Also, rows can now be clicked to open a popup with the full schedule JSON. Not pretty nor user friendly but it provides the info until we implement something decent.

(second commit): simplification of `apiRoot` related code

⚠️ This PR needs #216 to work.

## Changes

- script to generate `app/services/entities.ts`
  with static definitions of categories, queues and languages from backend module
- completion of categories filter using entities.ts
- added tags column to schedules view
- added tags filter (manual entry, intersection)
- added language service (to get english name)
- added language filter
- added queue filter
- added name filter
- added on-click popup with schedule JSON as a temporary way to get info (not meant to stay!)
- removed unused `const.service.ts`
- minor styling of filter popups (CSS)

### harmonised apiRoot usage

- single definition in a getApiRoot() function in config.ts
- use of it in both authentication, schedules and tasks.
- getApiRoot() returns `/api` path so its name is relevant.


dispatcher/frontend/gen_entities-ts.py
```diff
@@ -0,0 +1,32 @@
...
+export let queues = {queues};
+export let categories = {categories};
+export let languages = [
+    {languages}
+];
+
+""".format(queues=repr(ScheduleQueue.all()),
+           categories=repr(ScheduleCategory.all()),
+           languages=',\n\t'.join([repr(l) for l in ScheduleLanguage.all()]),
+           )
```
automactic: Sorry I think the language list should be fetched from API

rgaudin:
Thanks. Not a big fan of the huge grid neither.
Will update the PR to push the js stuff and fix/remove the language stuff.

rgaudin:
Made the required changes. thanks.

dispatcher/frontend/build_prod.sh
```diff
@@ -1,5 +1,12 @@
...
 rm -r ./../backend/static/angular/
+# replace entities placeholder with backend values
+python ./gen_entities-ts.py
 ng build --prod
-mv ./dist ./../backend/static/angular
\ No newline at end of file
+mv ./dist ./../backend/static/angular
+# offer to checkout entities placeholder
+read -t 10 -p "Do you want to run git checkout on the entities placeholder? [Y/N]" answer
+if [ "$answer" != "${answer#[Yy]}" ] ;then
```
automactic: I feel like we might simply create `categories.ts` and keep it up to date. I might be totally wrong,  but I don't think we will need to update the categories very often. Or we could fetch it from the API as well. 

Also, if we generate the file when build, it is gonna cause a bit trouble at develop time.

rgaudin: Yes, this call in `build_prod.sh` was totally unnecessary. removed it.

PR #224, opened 2019-05-30
Detecting stuck tasks

automactic:
## Rationale

We want to detect when a container generating zim files got stuck and terminate it. As discussed earlier, we define a container stuck as 10 mins without a stdout update.

Also, I am using asyncio here, because I wanted to use this design to add features like live cpu / memory usage, progress update or live container stream. Let me know if it make sense. 

## Changes

- added stuff to worker to detect container stuck


automactic:
Good point! I will update my PR

PR #226, opened 2019-05-31
Filter tags as list in UI

rgaudin:
## Rationale

* Allow tags filtering without knowing tag names.
* Update UI with latest changes

## Changes

* added /tags/ API (aggregate)
* added tests for /tags/
* updated ui to use /tags/
* generated new build of angular app.


rgaudin:
Well, I though the same thing at first but now that we have a clear definition of the releases we want: `mini`, `nopic`, `maxi` for both selections `top` or full `all`, it is useful (at least in the UI).

But I think that once the discussion on ZIM tags is settled, it might make less sense to have this _duplicate_ tagging system here.

PR #231, opened 2019-06-03
update get schedule to tun tool

automactic:
## Rationale

what I usually do is run this in command line and get a subset of schedules to run. it prints out the son payload a s string.


rgaudin:
Good. could we use that opportunity to make it less-chrisy :)

automactic:
Oops, I thought I committed my password or ssh key for a moment

Issue #233, opened 2019-06-05
Redis should be used via socket (not via tcp)


kelson42:
Using Redis via socket has a significantly lower latency:
https://stackoverflow.com/questions/26934941/unix-sockets-slower-than-tcp-when-connected-to-redis?answertab=votes#tab-top

Here is an article about how to achieve that:
https://medium.com/@jonbaldie/how-to-connect-to-redis-with-unix-sockets-in-docker-9e94e01b7acd

kelson42:
Implemented by @rgaudin

PR #234, opened 2019-06-07
mwoffliner-redis connection via socket

rgaudin:
## Rationale

Request in #233 to improve perf/latency/reliability of mwoffliner-redis communication.

## Changes

- uses a busybox container to _host_ the socket (ran per task)
- socket is postfixed with task_id to prevent conflict on host mapping

### DIFF
--- a/worker/app/operations/__init__.py
+++ b/worker/app/operations/__init__.py
@@ -1,3 +1,4 @@
+from .run_socket import RunSocket
 from .run_redis import RunRedis
 from .run_mwoffliner import RunMWOffliner
 from .run_phet import RunPhet
--- a/worker/app/operations/run_mwoffliner.py
+++ b/worker/app/operations/run_mwoffliner.py
@@ -1,3 +1,4 @@
+import os
 import asyncio
 from pathlib import Path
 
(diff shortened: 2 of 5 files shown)


rgaudin:
You mean having a single socket container and using `volume_from` from it on all mwo containers?

We've discussed this with @kelson42 and it was agreed that we should, as much as possible, not share resources across tasks, even though that means wasting some resource on the wroker (docker overhead).

I have an upcoming PR with the same structure coming…

that said, the socket is a special case as:

* the resources are a single folder on the host so _isolation_ has no real benefit.
* there's no overhead as we need the _container_ but we don't need it running (we run it to create it but it exits right away)

So it's just a design decision. I like it as a host-feature but I'm fine with having it per-task as well. @kelson42 @automactic, suggestion ?

kelson42:
@rgaudin No

> So it's just a design decision. I like it as a host-feature but I'm fine with having it per-task as well. @kelson42 @automactic, suggestion ?

No, you explained it well. I think this is the most robust way to implement it and that for only a little price (in resources).

automactic:
@rgaudin no I mean like following:
```
redis ----|----- host -------|-----mwoffliner
redis_a <--> redis_a.sock <--> mwoffliner_a
redis_b <--> redis_b.sock <--> mwoffliner_b
redis_c <--> redis_c.sock <--> mwoffliner_c
```

so no need for any socket container, simply map the `redis.sock` to different socket files on host, and map those socket files to corresponding mwoffliner container. Will that work?

rgaudin:
Replaced by #236

PR #236, opened 2019-06-13
mwoffliner-redis connection via socket (simple volume version)

rgaudin:
## Rationale

Request in #233 to improve perf/latency/reliability of mwoffliner-redis communication.

## Changes

- mount a shared `/tmp/sockets` on both `mwoffliner_xxx` and `redis_xxx` containers
- redis starts with a socket inside this volume
- socket is postfixed with task_id to prevent conflict on host mapping

### DIFF
--- a/worker/app/utils/settings.py
+++ b/worker/app/utils/settings.py
@@ -23,12 +23,10 @@ class Settings:
 
     working_dir_host: str = os.getenv('WORKING_DIR', None)
     working_dir_container: str = '/zim_files'
+    sockets_dir_container: str = '/tmp/sockets'
     private_key: str = '/usr/src/.ssh/id_rsa'
 
     docker_socket: str = '/var/run/docker.sock'
-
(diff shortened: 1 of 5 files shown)


worker/app/utils/settings.py
```diff
@@ -27,6 +27,9 @@ class Settings:
 
     docker_socket: str = '/var/run/docker.sock'
 
+    sockets_dir_host: str = os.getenv('SOCKETS_DIR', '/tmp/sockets')
+    sockets_dir_container: str = '/tmp/sockets'
```
automactic: Hmm, would it be a good idea to just utilize the `host_working_dir` env? I personally think it is a good idea to keep all files zimfarm needs and generated in one place, instead of scattered around the file system.

rgaudin: OK, I think variables are less clear this way but you're right, the less volumes the better.

PR #241, opened 2019-06-17
Update schedule UI

automactic:
## Rationale

1/N PRs to make schedule page better and more useful.

Screen Shot 2019-06-16 at 10 24 43 PM

## Changes


automactic:
> Should we have something less fancy but dynamic for flags?
Yes, flags are supposed to be based on what docker image user chose. I am just not there yet. 

> What's the circle for?
Yes it is for an icon. Yes each category is supposed to have a static icon. It's basically an another way to display category info.

> as I may not be present
Not sure what this means...

> we're dropping the filtering on schedules completely
No the plan is to add category filters, language filter and task size filters (based on queue)

> there's a small bug on refreshing the UI: 
Yep, like mentioned above, flags are supposed to be dependent on the docker image user chose.

PR #243, opened 2019-06-17
Exit code at the end of the container tab

### LINKED ISSUE #237: in "Container" view, put the exit code at the end.
Currently this is just after the command. I would like to have it at the end to keep the chronological order. It is also not a really important information (at least less than stderr/stdout).

kelson42:
Fix #237


automactic:
I am merging it to rebuild the static site

Issue #174, opened 2019-04-14
Logs should be kept


kelson42:
We need to keep the logs (stdout, stderr, etc.) for investigation reason, even if the scraping process has been (apparently) successful. We should do so, as long, as the scheduling has not run an other time. This logs should be easily available through API and Web UI, probably in a compressed formed.

rgaudin:
Logs should be compressed and uploaded via SFTP to the warehouse as it will be kept only for debugging purposes and doesn't need to interact much with the zimfarm. A way to access/retrieve from the zimfarm should be enough.

## challenges (based on mwo):

* logs are expected to be GBs large (compressed). With about 1,500 schedules (most being smaller ZIMs), we are talking about hundreds of GBs of compressed logs at any given time.
* creation process takes days
* there is almost-constant output during that time
* writing those logs to disk **might** slow down the process? (ie. consume I/Os)
* keeping those logs to RAM might slow down the process (ie. consume precious RAM)
* periodic upload of chunks would solve those two but would mean uploading/appending uncompressed logs to the warehouse: too large to keep on the WH, too large to DL for users and a general waste of BW. Compressing it on the WH seems fragile and CPU-intensive (not the role of the WH).

## Option

Given the scrappers are running in docker, logs are already written to disk (on host at `/var/lib/docker/<cid>/<cid>-json.log`) so there's no additional impact on scrapper performance. It thus seems reasonable to just retrieve and compress the whole docker log.

`xz` compression seems appropriate as our workers are already doing CPU-intensive work and we'll want to save space on WH. `xz` also un-compresses very fast which will make it easy for users/debugging.

We'll want to store it at `/logs/<task_id>.log.xz` for easy retrieval. (ie. `/data/zimfarm-warehouse/logs` on the server).

We are currently sending the last 100 lines of `stdout` and the entirety of `stderr` via rabbitmq. It is expected we'll continue to send both this way and additionally compress/upload the full `stdout`. We might revise this should a scrapper write tons of stuff to `stderr` but it is unlikely and might make more sense to be fixed on the scrapper.

So, on task received, we'll:

* get last task ran for schedule
* remove `<previous_task_id>.log.xz` on warehouse
* [normal worker work]
* on task completion (whatever the reason):
  * get and compress the full log (not just the tail) ⌛️ 
  * upload the full log to WH ⌛️ 
  * still upload tail via rabbitmq

kelson42:
@rgaudin has done this.

PR #244, opened 2019-06-17
Upload logs to Warehouse via SFTP (no-remove)

rgaudin:
## Rationale

#174 we want to keep a copy of the full log of the main container

## Changes

Copy of #238 without the code required to remove the previous log (api endpoint and test, api access from worker and file removal in worker).

### Warehouse

* changed root to `/files/` instead of `/zim_files` (for semantic reason)
* kept zim uploads in `/files/zim` but creates a `/files/logs` folder as well to receive logs.
* added support for the `remove()` command (we'll need that soon to remove logs when task's EOL is implemented)

### Worker

* get full output log (`stdout` + `stderr`) from docker
* save it to a temporary file while compressing it into `xz`: compression is done in batch of 1M lines
* upload the log to `/logs/<container_name>.log.xz`. Chose this so we could one day have multiple logs per task (redis for instance)
* upload log info saved to `ContainerResult`
* slight refactoring of upload code (for files)

### monitor

* on container finish event, store log file info into mogo (filename and size)

### U.I

* added a link in task detail to the compressed log file (via HTTP/s, already deployed on server)
* rebuilt angular project


warehouse/app/warehouse/__init__.py
```diff
@@ -74,13 +74,15 @@ def _configure_root_dir():
...
-                os.makedirs(root)
-        else:
-            os.makedirs(root)
+        root = os.getenv('ROOT_PATH', '/files')
+
+        for subdir_name in ('zim', 'logs'):
+            subdir = os.path.join(root, subdir_name)
+            if os.path.exists(subdir):
+                if not os.path.isdir(subdir):
+                    os.remove(subdir)
```
automactic: I am not sure, but if the subdir has content in it, is `remove` gonna be successful?

automactic: According to the [docs](https://docs.python.org/3.7/library/os.html#os.remove), `rmdir` is needed to remove a dir

warehouse/app/warehouse/sftp.py
```diff
@@ -157,7 +157,23 @@ def readlink(self, path):
...
+        # check permission
+        if not path.startswith(self.root):
+            return sftp.SFTP_PERMISSION_DENIED
+
+        # return error if file doesn't exists
+        if not os.path.exists(path):
+            return sftp.SFTP_FAILURE
+
+        try:
+            os.unlink(path)
```
automactic: should we consider the case if path is a dir?

worker/app/operations/upload.py
```diff
@@ -30,9 +40,52 @@ def execute(self):
...
                         continue
                     client.upload_file(self.remote_working_dir, file)
+
+                    if self.delete:
+                        file.unlink()
         except SSHException as e:
             raise UploadError(code='paramiko.Error', message=str(e))
+
+    @staticmethod
+    def upload(remote_working_dir: str, files: list = None, directory: Path = None, delete: bool = False):
```
automactic: Personally I am not a fan of having `files` and `directory` and ask user to provide one of them. What happens if I provide both, do they / should they all get uploaded? 

Maybe we should have `upload_files` and `upload_directory_content`, one could call the other? What do you think?

worker/app/operations/upload.py
```diff
@@ -30,9 +40,52 @@ def execute(self):
...
+
+        if files is None and directory is None:
+            raise ValueError("upload requires either `files` or `directory`")
+
+        # build list of files if invoked with directory
+        if files is None:
+            files = [file for file in directory.iterdir() if not file.is_dir()]
+
+        stats = [{'name': file.name, 'size': file.stat().st_size} for file in files]
+        files_desc = ','.join(['{name} - {size}'.format(**stat) for stat in stats])
```
automactic: nit: would you consider consolidate these 2 line into one line with python3.7 style string interpolation?

worker/app/operations/upload.py
```diff
@@ -30,9 +40,52 @@ def execute(self):
...
+
+        upload = Upload(remote_working_dir=remote_working_dir, files=files)
+        upload.execute()
+
+        return stats
+
+    @staticmethod
+    def upload_log(container):
+        """fetch, compress and upload the container's stdout to the warehouse"""
+        container.reload()
```
automactic: Why do we need `reload`? 🤔

worker/app/tasks/base.py
```diff
@@ -52,22 +53,13 @@ def clean_up(self):
...
+        shutil.rmtree(working_dir, ignore_errors=True)
 
     def on_success(self, retval, task_id, args, kwargs):
         self.clean_up()
 
     def on_failure(self, exc, task_id, args, kwargs, einfo):
         self.clean_up()
+
+    def upload_zims(self, remote_working_dir: str, files: list = None, directory: Path = None):
+        return Upload.upload(f'/zim{remote_working_dir}', files, directory)
```
automactic: Hmmm if this is just one line, I am not sure it is really beneficial to make it a separate function.

dispatcher/frontend/src/app/task/task-detail/task-detail.ts
```diff
@@ -4,12 +4,16 @@ import { Observable } from 'rxjs';
 import { switchMap } from 'rxjs/operators';
 
 import { Task, TasksService } from '../../services/task.service';
+import { getWareHouseLogsUrl } from '../../services/config';
```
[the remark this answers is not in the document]
rgaudin: Ah indeed you introduced that after. I'll merge master-in and use the base service

worker/app/operations/upload.py
rgaudin: 👍

worker/app/operations/upload.py
rgaudin: you're right, we don't. `logs()` is not using cached data.

worker/app/tasks/base.py
rgaudin: I'd prefer to keep it even though it's simple as it's gonna be used in every scraper and makes sure we keep that `/zim` prefix in place.

warehouse/app/warehouse/sftp.py
rgaudin: right.

worker/app/operations/upload.py
rgaudin: not sure what you mean. I switched to literal string for `files_desc` but don't want to combined them as `stats` is useful and returned while `files_desc` is just for logging.

PR #245, opened 2019-06-18
Fixed X962 error in worker startup

rgaudin:
## Rationale

Worker can't start anymore due to a cryptic X962 error.

Due to an update of paramiko without one for cryptography.

## Changes

* updated all worker requirements to latest versions
* using fixed (`==`) versions to avoid this kind of issue.


automactic:
Oops merged the PR by accident

automactic:
guess I can restart the workers tonight

PR #246, opened 2019-06-19
Check sockets folder permission on start to avoid in-task issue

rgaudin:
## Rationale

`sockets` subfolder needs to be world writable (`0777`) for redis to create the socket (redis runs as a user)

## Changes

* updated doc to mention the required step
* added check in `sanity_check()` so the worker fails upon start if not correct

### DIFF
--- a/worker/README.md
+++ b/worker/README.md
@@ -73,12 +73,7 @@ Please make sure the RSA public key is uploaded to dispatcher using the public A
 
 ## Example
 
-__note__: your local `PATH_WORKING_DIR` must be group writable and `PATH_WORKING_DIR/sockets` must be world writable:
-
-```
-chmod g+rwx PATH_WORKING_DIR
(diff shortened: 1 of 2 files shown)


worker/README.md
```diff
@@ -73,7 +73,12 @@ Please make sure the RSA public key is uploaded to dispatcher using the public A
 
 ## Example
 
-__note__: your local `PATH_WORKING_DIR` must be group writable (chgrp rwx PATH_WORKING_DIR)
+__note__: your local `PATH_WORKING_DIR` must be group writable and `PATH_WORKING_DIR/sockets` must be world writable:
+
+```
+chmod g+rwx PATH_WORKING_DIR
+mkdir -p -m 0777 PATH_WORKING_DIR/sockets
```
automactic: I can understand why we have to do this, but I don't like that user have to perform extra step to get things working

rgaudin: would you prefer the worker to do it itself?

PR #255, opened 2019-06-27
Cancel started Task

### LINKED ISSUE #220: Task cancellation
One should be able to cancel a running task: to free a worker or because there is a mistake in the zim-generation or in the metadata.

For this to be possible, we need a `/tasks/<task-id>/cancel` POST endpoint that will send a cancel message to the assigned worker.

The worker should thus stop the running container(s) and mark the (celery and mongo) tasks as revoked/canceled.

rgaudin:
## Rationale

Fixes #220

## Changes

* Added a `POST /tasks/<tid>/cancel` endpoint
* Added tests for endpoint
* endpoint calls `revoke()` on task which kills the task process in the worker (if started) or dismisses it if not started.
* Added a general handler in worker to remove dnscache container on revoke.

### DIFF
--- a/dispatcher/backend/src/routes/auth/rabbitmq.py
+++ b/dispatcher/backend/src/routes/auth/rabbitmq.py
@@ -29,6 +29,8 @@ def auth(intention: str):
             return Response('allow')
         else:
             user = Users().find_one({'username': username}, {'_id': 0, 'password_hash': 1, 'scope.rabbitmq': 1})
+            if user is None:
+                return Response('deny')
             password_hash = user.get('password_hash', '')
             tags = user.get('scope', {}).get('rabbitmq', [])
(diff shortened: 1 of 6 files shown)


dispatcher/backend/src/routes/auth/rabbitmq.py
```diff
@@ -29,6 +29,8 @@ def auth(intention: str):
             return Response('allow')
         else:
             user = Users().find_one({'username': username}, {'_id': 0, 'password_hash': 1, 'scope.rabbitmq': 1})
+            if user is None:
```
automactic: this check is already done on line 36

rgaudin:
Yes but line 35 fails in this case : can’t do get() on None...

PR #271, opened 2019-07-22
Fixed DNS Cache tear down/set up issue

rgaudin:
## Rationale

mwoffliner tasks have difficulties running one after another as the property might point to the wrong `dnscache_xxx` container.

## Changes

* simplified `stop()` on dnscache to reuse existing code which is also safer (prevents `.stop()` on non-existing container).
* check the `task_id` of the attached `run_dnscache` property and change if required.

### DIFF
--- a/worker/app/tasks/base.py
+++ b/worker/app/tasks/base.py
@@ -9,7 +9,7 @@ from celery.utils.log import get_task_logger
 
 from utils import Settings
 from operations import Upload
-from utils.docker import get_ip_address, remove_existing_container
+from utils.docker import remove_existing_container
 from operations.run_dnscache import RunDNSCache
 
 logger = get_task_logger(__name__)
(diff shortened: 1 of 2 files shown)


worker/app/tasks/base.py
```diff
@@ -43,7 +43,7 @@ def set_up(self):
 
     def get_dns(self):
         """list of DNS IPs to feed `dns` on scrappers with: [<dnscache_ip>]"""
-        if not hasattr(self, 'run_dnscache'):
+        if not hasattr(self, 'run_dnscache') or self.task_id not in self.run_dnscache._container_name:
```
automactic: private property access is bad

worker/app/tasks/base.py
```diff
@@ -43,7 +43,7 @@ def set_up(self):
 
     def get_dns(self):
```
automactic: I didn't go into too much detail, but I feel like `set_up` and `get_dns` has danger to be calling each other.

worker/app/tasks/base.py
rgaudin: right, followed convention when created it but it doesn't make sense to keep this one private. changed.

worker/app/tasks/base.py
rgaudin: right, thanks. even if it's safe and doesn't look like it is, it's wrong. refactored it, I think it's much better now.

Issue #267, opened 2019-07-10
Allow custom timeout via ENV


rgaudin:
Currently we have a strict in-code 10mn timeout for no-log in mwoffliner before we kill it.

It might be too short for less-performing workers to complete long steps like ZIM file creation at the end.

A good addition would be to allow this 10mn variable to be set at run time via a docker env.

PR #273, opened 2019-07-22
Allow custom timeout via ENV

rgaudin:
## Rationale

#267 

## Changes

* Added an `ENV` to Dockerfile, int-casting and no value handling.
* Used the opportunity to fix a typo in log messages on settings
* harmonized `ENV` usage for concurrency as it's similar.

Note: not using default parameter in `_detect_stuck` definition as it is parsed at import time.


rgaudin:
Integrated into #274.

PR #274, opened 2019-07-23
Fixed #267 #261: custom ENV for IDLE_TIMEOUT and container-killed event

### LINKED ISSUE #261: We need a trace (log) if worker is killing scraper container (because of inactivity)
It looks like for now, we have no way to know that. See:
https://github.com/openzim/zimfarm/issues/260#issuecomment-508294636

To efficiently analyse failing scrapings, we need this information.

### LINKED ISSUE #267: Allow custom timeout via ENV

rgaudin:
## Rationale

* Fixes #267
* Fixes #261 

Here's a combination (for review/merge sake) of two fixes: the customization of the idle timeout via an `ENV` and the report of a killed container to the zimfarm.

Currently, the idle timeout is a fixed 10mn hard-coded.

Also, there is no trace of a container being killed except a single line in the worker log (not accessible then). We used to guess a killed due to the mwoffliner exit_code but that's neither wanted nor reliable.

## Changes

__#267__
* Added an `ENV` to Dockerfile, int-casting and no value handling.
* Used the opportunity to fix a typo in log messages on settings
* harmonized `ENV` usage for concurrency as it's similar.

_Note_: not using default parameter in `_detect_stuck` definition as it is parsed at import time.

__#261__
chose to go with a _dumb_ `container-killed` event instead of changing the `container-finished` one to keep it simple. It's just an extra event letting us know that container was killed.

* Added `container_killed` in `TaskEvent` and `TaskStatus` (backend)
* Added a monitor handler for `task-container_killed` event: saves event to mongo (with the passed `timeout`)
* `on_container_killed(timeout)` in `MWOffliner` Task that emits the event.

_note_: not displaying the timeout in the UI. Seems unnecessary (can be retrieved via API).

### DIFF
--- a/dispatcher/backend/src/common/entities.py
+++ b/dispatcher/backend/src/common/entities.py
@@ -6,6 +6,7 @@ class TaskStatus:
     succeeded = 'succeeded'
     container_started = 'container_started'
     container_finished = 'container_finished'
+    container_killed = 'container_killed'
     failed = 'failed'
     retried = 'retried'
     revoked = 'revoked'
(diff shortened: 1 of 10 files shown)


kelson42:
@rgaudin Please merge this PR

PR #295, opened 2019-09-15
schedule config update

automactic:
## Rationale

WIP

## Changes


rgaudin:
Thanks @automactic, is this a WIP? There's no detail (although refactoring is obvious).

I see that you've prepared for allowing fine-grained config updates, which is clearly needed.

Regarding `DockerImageSchema` I don't think we want to restrict the tags to a list. It sounds like too much maintenance work and a useless constraint. mwoffliner is the most used one and I can clearly see us testing a run with a past version of using the `dev` tag.

On the `MWOfflinerConfigFlagsSchema`, we're missing the `--speed` and `--addNamespaces` flags that we're using already. `--requestTimeout` and `--useCache` should also be present. 

There's a `skipCacheCleaning` one though that I think doesn't exist in mwoffliner.

PR #296, opened 2019-09-19
Update UI

automactic:
## Rationale

Screen Shot 2019-09-18 at 8 14 00 PM

## Note

This PR only contains the built code. I have created [another repo](https://github.com/automactic/zimfarm-ui) for UI only stuff.

I think there has been complaints before that the main zimfarm repo is too polluted with compiled js and stuff. And we wanted to explore the possibility of using another docker image to build the site on docker hub. This is a great chance if we want to go that direction.

@kelson42 if you would like to take this approach, feel free to move this new UI repo to openzim org

## Changes

- new design
- update schedule offline config


rgaudin:
OK then Chris, please follow as suggested:
* close this PR or replace it with `static` removal
* submit another one with the source code.

Also please prepare a docker container. I guess we'll have to move the API to a different subdomain if we're to have the UI container serve `farm.openzim.org`. Would `api.farm.openzim.org` work? @kelson42 could you create it?

rgaudin:
here's a sample `Dockerfile` for the UI (using multi-stages):

```
# build web app
FROM node:alpine as builder
RUN apk update && apk add --no-cache git

# install dependencies
RUN cd /zimfarm && npm set progress=false && npm install

# copy and build app
COPY .  /zimfarm
RUN cd /zimfarm && npm run build

# build small nginx from built app
FROM nginx:alpine

# remove default site
RUN rm -rf /usr/share/nginx/html/*

COPY --from=builder /zimfarm/dist/Zimfarm /usr/share/nginx/html

EXPOSE 80
CMD ["nginx", "-g", "daemon off;"]
```

kelson42:
@automactic The JSON should be indented properly like in the current version.

rgaudin:
FYI, a list of things I've noticed:

* Links are now based on schedule IDs and not name (human-friendly). It's way less practical as one can not just guess the URL from its name, forcing user to use the list/search.
* At the moment, only category filtering is there. We definitely need the other ones back.
* I think pagination is either missing or not intuitive
* the fact that there's no scrollbar on the fixed-height width is also disturbing… but that may be due to the pagination stuff.

rgaudin:
closing as not relevant anymore.

PR #397, opened 2020-01-24
Parallel zimcheck

### LINKED ISSUE #385: Allow parallel run in the quarantine
The quarantine system works well but might be quite long to digest the ZIM file throughput. We have time to time a bottleneck at this stage.

Currently the quarantine can analyze only one ZIM file at a time. But the system might be able to support `n` analysis in parallel at a time.

For this reason it would be good if the quarantine would be able to handle many check in parallel and that this level of parallelization might be configuration easily (via a Docker env. variable).

kelson42:
Fixes #385


kelson42:
@rgaudin Please carry on and merge. Thx!

PR #448, opened 2020-03-14
man page and source file in markdown

JensKorte:
pandoc -o zimfarm.1 -s -t man zimfarm.1.md

## Rationale

collected the informations for using zimfarm in one file

## Changes

add the man page and its source file in pandoc's markdown


rgaudin:
Thanks @JensKorte !

* Could you please move it to target the `contrib/` folder, next to the script and example config?
* what's the use of the markdown file? Who would use it and how?
* The `zimfarm` script is dangling in the repo and not installed. We shall probably mention this manpage in the workers/README so that users are informed it exists as it won't be installed automatically.

With all that said, I'm not sure how useful this is. It won't be auto installed, we have a very limited number of users for this and it would require additional maintenance for each modification on the script itself. @kelson42 what do you think?

rgaudin:
Don't mind my markdown comment, I didn't realize it was the source file. Haven't written a man file in ages and clearly remember doing it the hard way then :)

@kelson42, you opinion is requested…

JensKorte:
>     * The `zimfarm` script is dangling in the repo and not installed. We shall probably mention this manpage in the workers/README so that users are informed it exists as it won't be installed automatically.
> 
> 
> With all that said, I'm not sure how useful this is. It won't be auto installed, we have a very limited number of users for this and it would require additional maintenance for each modification on the script itself. @kelson42 what do you think?

I could also provide the man page in my repo and you can link it. Then it should be clear, that it may be not well updated.

kelson42:
Regarding the format, I'm really in favour of proper markdown documentation, in particular for zimfarm parts which are end-user or for worker facing. To the contrary, `man` pages without proper packaging do not seem to be the best approach.

JensKorte:
> Regarding the format, I'm really in favour of proper markdown documentation, in particular for zimfarm parts which are end-user or for worker facing. To the contrary, `man` pages without proper packaging do not seem to be the best approach.

No problem to me. I can host it on my own, since I like man pages.

Which style of documentation do you like, howto, list of options, --help usage? Maybe I can reformat the document to be of use for the project.

Issue #446, opened 2020-03-14
Recipe history (old tasks) get lost by renaming


kelson42:
Steps:
- rename a recipe with tasks in history

I get:
An empty history

I want:
To keep the history of pas tasks

HAKSOAT:
Hello @kelson42 can you help with a pointer to the part of the code base that does this? I've spent some time trying to figure out on my own.

kelson42:
@HAKSOAT Sorry, this is a question for @rgaudin, I'm not a Python dev.

rgaudin:
That would be around https://github.com/openzim/zimfarm/blob/77d2117c8301a0333af50e30fd374dc281cc2ac0/dispatcher/backend/src/routes/schedules/schedule.py#L156

there's no such thing as renaming per say but a schedule's `name` can be updated, given the new name is not taken.
After that check, you would have to make sure that once renamed, you also change the `schedule_name` of `Tasks` and `RequestedTasks` for the previous name.

HAKSOAT:
@rgaudin I just opened a PR #457 It took a while as I had to do a bit of understanding pymongo (not worked with it before). 

Also, is there any documentation that shows how I can setup zimfarm on my machine? I wasn't able to test the modification here, but I think it should work.

rgaudin:
replied on the PR. No, there's no documentation and it's a bit of a pain (been there). We should document that though.

HAKSOAT:
> 
> 
> replied on the PR. 

Seen. 

>No, there's no documentation and it's a bit of a pain (been there). We should document that though.

Yeah. It's been kind of hard to play around.

PR #457, opened 2020-03-18
Updates Tasks and Requested_Tasks with new name

HAKSOAT:
## Rationale

When a schedule is updated with name changes, its equivalent task and requested task are lost. Hence, they need to be updated with the new name.

Issue: #446

## Changes

Modified `dispatcher.backend.src.routes.schedules.schedule` to update `Tasks` and `Request_Tasks` names when a name change is done through a PATCH on `ScheduleRoute`

### DIFF
--- a/dispatcher/backend/src/routes/schedules/schedule.py
+++ b/dispatcher/backend/src/routes/schedules/schedule.py
@@ -3,7 +3,7 @@ from http import HTTPStatus
 from flask import request, jsonify, Response, make_response
 from marshmallow import ValidationError
 
-from common.mongo import Schedules, Tasks, Requested_Tasks
+from common.mongo import Schedules, Tasks, RequestedTasks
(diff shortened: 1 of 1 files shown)


rgaudin:
@HAKSOAT that won't work as you're using `update_one` which, as it sounds, would only update one record of each while we want to change all. Should be a quick fix.

HAKSOAT:
@rgaudin I changed it to `update_many` . I thought that it was one task.

rgaudin:
@HAKSOAT there were a couple mistakes:

* you mispelled `RequestedTasks` as `Requested_Tasks`
* you reused the `query` that we use to update the schedule but since it targets a `name`, it could not work as the Tasks and RequestedTasks reference it using `schedule_name`.

HAKSOAT:
@rgaudin Thanks. Noted.

PR #460, opened 2020-03-23
Fix Recipe Trailing Space Issue

### LINKED ISSUE #458: Don't allow trailing space on recipe name

HAKSOAT:
## Rationale

Recipes with trailing space are not reachable from the UI

Fixes  #458

## Changes

Modified `dispatcher.backend.src.routes.schedules.schedule`. Stripped away spaces from document name when creating a schedule through a GET request on ScheduleRoute.

### DIFF
--- a/dispatcher/backend/src/routes/schedules/schedule.py
+++ b/dispatcher/backend/src/routes/schedules/schedule.py
@@ -73,6 +73,7 @@ class SchedulesRoute(BaseRoute):
 
         try:
             document = ScheduleSchema().load(request.get_json())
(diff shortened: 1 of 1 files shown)


rgaudin:
Thanks, sorry for the delay.

I think we'd want to do that on the UI side: strip the input as that's more certainly a typo and send the stripped version to the API.

The API should remain predictable so either we keep it as it is (ie. if using the API directly, you can add trailing space) or we could change the schema to fail on names with trailing spaces.

kelson42:
@rgaudin I guess this PR is dead...

PR #463, opened 2020-03-26
Allow more SSH connections to the receiver

### LINKED ISSUE #449: Allow more connexions to the receiver?
We are seeing some upload failures from the workers to the receiver with reason like `ssh_exchange_identification: Connection closed by remote host`.

While there might be numerous reasons for this, a good candidate is sshd dropping connexions because it reached its limit.

No options regarding this have been modified from the defaults on the [sshd_config](https://github.com/openzim/zimfarm/blob/master/receiver/sshd_config) that we use.

```
#ClientAliveCountMax 3
#MaxAuthTries 6
#MaxStartups 10:30:100
#MaxSessions 10
```

With the number of workers increasing and the large number of small tasks being run, uploading ZIM is very frequent.
Additionally (and more importantly), **all workers' tasks** are uploading their scraper log constantly. It's basically an always-running file upload for each task (uploads a chunk, disconnects then uploads a chunk, etc).

I don't think that there is a global connection limit in sshd_config so the nb of workers or tasks in general should not be a concern but I can imagine we might reach the `MaxSessions` on a worker running multiple task ?

Also, all foundation workers are on a virtual network which looks like a single IP from the outside. [manual](https://man.openbsd.org/sshd_config#MaxSessions) is not clear on what is a _network connection_ in the `MaxSessions` context.

``` man
MaxSessions
     Specifies the maximum number of open shell, login or subsystem (e.g. sftp) sessions permitted per network connection.  Multiple sessions may be established by clients
     that support connection multiplexing.  Setting MaxSessions to 1 will effectively disable session multiplexing, whereas setting it to 0 will prevent all shell, login and
     subsystem sessions while still permitting forwarding.  The default is 10.
```

So we might want to experiment with `MaxSessions` at least to see if that might be an issue.

kelson42:
Fixes #449

### DIFF
--- a/receiver/sshd_config
+++ b/receiver/sshd_config
@@ -33,6 +33,7 @@
 #StrictModes yes
 #MaxAuthTries 6
 #MaxSessions 10
+MaxSessions 20
 
 PubkeyAuthentication yes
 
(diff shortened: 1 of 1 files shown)


kelson42:
@rgaudin OK, will do. Thx for the review

PR #506, opened 2020-06-25
Don't allow trailing and leading spaces in recipe name

### LINKED ISSUE #458: Don't allow trailing space on recipe name

satyamtg:
This fixes #458 by doing the following changes -
- trim() names before passing to payload in the frontend (CloneSchedule.vue, ScheduleEditor.vue)
- raise bad request on backend if recipe name is strippable (schedule.py)

Changes are incorporated for creating/updating/cloning a recipe (schedule)


dispatcher/backend/src/routes/schedules/schedule.py
```diff
@@ -76,6 +76,10 @@ def post(self, token: AccessToken.Payload):
         except ValidationError as e:
             raise InvalidRequestJSON(e.messages)
 
+        # make sure name doesn't contain whitespaces around it
```
rgaudin, line 79: this should be part of schema validation

dispatcher/backend/src/routes/schedules/schedule.py
```diff
@@ -154,6 +158,8 @@ def patch(self, schedule_name: str, token: AccessToken.Payload):
             raise InvalidRequestJSON(e.messages)
 
         if "name" in update:
+            if len(update["name"]) != len(update["name"].strip()):
```
rgaudin, line 161: Idem, should be in schema validation

dispatcher/backend/src/routes/schedules/schedule.py
```diff
@@ -221,6 +227,8 @@ def post(self, schedule_name: str, token: AccessToken.Payload):
 
         request_json = CloneSchema().load(request.get_json())
         new_schedule_name = request_json["name"]
+        if len(new_schedule_name) != len(new_schedule_name.strip()):
```
rgaudin, line 230: Idem in CloneSchema

dispatcher/frontend-ui/src/components/CloneSchedule.vue
```diff
@@ -32,7 +32,7 @@
         if (!this.ready)
           return;
         let parent = this;
-        let payload = {name: this.form_name};
+        let payload = {name: this.form_name.trim()};
```
rgaudin, line 35: I think `ready()` should do the `trim()` as well so that a new name such as ` ` doesn't trigger the clone button.
Also, we should maybe update `this.form_name` so that in case of error, the user input is updated as well.

dispatcher/frontend-ui/src/components/ScheduleEditor.vue
```diff
@@ -377,8 +377,13 @@
         let parent = this;
 
         ["name", "category", "enabled", "periodicity"].forEach(function (key) {
-          if (parent.edit_schedule[key] != parent.schedule[key])
-            payload[key] = parent.edit_schedule[key];
+          if (parent.edit_schedule[key] != parent.schedule[key]) {
```
rgaudin, line 380: Wouldn't it be easier to do `parent.edit_schedule[key].trim();` instead? We'd get the UI updated as for clone and that would remove the need for this special treatment here.

dispatcher/backend/src/routes/schedules/schedule.py
satyamtg: Done

dispatcher/backend/src/routes/schedules/schedule.py
satyamtg: Done

dispatcher/backend/src/routes/schedules/schedule.py
satyamtg: Done

kelson42:
@satyamtg @rgaudin Any chance we could close/merge finally this PR without much additional work?

rgaudin:
Eventually merged in #576

PR #523, opened 2020-08-21
Add support for --no-identicons option in sotoki

satyamtg:
This adds support for the `--no-identicons` option in sotoki


dispatcher/backend/src/common/schemas/offliners/sotoki.py
```diff
@@ -87,6 +87,15 @@ class Meta:
         },
     )
 
+    no_identicons = fields.Boolean(
+        required=False,
+        data_key="no-identicons",
+        metadata={
+            "label": "No Identicons",
+            "description": "Don't download user profile pictuures",
```
rgaudin, line 95: typo

PR #527, opened 2020-09-01
Add new black formatting

satyamtg:
Adds new formatting from black 20.8

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -21,9 +21,8 @@ jobs:
         pip install pytest
     - name: black code formatting check
       run: |
-        pip install black
-        black --version
-        black .
+        pip install -U "black>=20.8b1"
+        black --check .
     - name: Test with pytest
       run: |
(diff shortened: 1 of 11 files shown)


rgaudin (changes requested):
black in CI must be updated as well (using updated action?)

satyamtg:
> black in CI must be updated as well (using updated action?)

I browsed the marketplace and there doesn't seem to be an action using black 20.8b1 . Have created https://github.com/jpetrucciani/black-check/issues/1 for the same.

rgaudin:
> > black in CI must be updated as well (using updated action?)
> 
> I browsed the marketplace and there doesn't seem to be an action using black 20.8b1 . Have created [jpetrucciani/black-check#1](https://github.com/jpetrucciani/black-check/issues/1) for the same.

Then we can remove the action and add a regular step install it via pip

rgaudin:
thanks ; specified min version and used `black --check`

rgaudin (approved):

PR #533, opened 2020-09-14
Support new --threads parameter for ted and openedx

satyamtg:
This adds support for the `--threads` parameter in ted and openedx (yet to be merged to master. See - https://github.com/openzim/openedx/pull/144)


rgaudin (approved):
Removed unnecessary `data_key` as the variable is the same already.

Issue #487, opened 2020-05-26
Durations rounding is misleading


kelson42:
See:
![image](https://user-images.githubusercontent.com/1029718/82871577-1a800480-9f32-11ea-9835-ab0c257e4113.png)

Duration should be: 3 hours + 17 mn + 2 hours... but it is 3 hours. Something is wrong here because ZIM files are here not build in paralell.

rgaudin:
OK so the total computation time (`Duration: 3 hours`) appears to be less than the sum of all the ZIM files creation (which are run one after another).

This is not a computation error. This duration is just the duration between `started` and `scraper_completed`.

The issue is with the rounding of the value done by [moment.js](https://momentjs.com/docs/#/durations/humanize/) which is not precise enough.

Maybe we should not use `humanize()` for durations and use a short string like `02:17:36` ?

mahakporwal02:
I am sorry about directly making this PR without mentioning anything earlier here....kindly assist me to go ahead with this PR

rgaudin:
That's OK ; will look at it tomorrow. thank you

mahakporwal02:
@rgaudin any updates??

mahakporwal02:
@rgaudin So does this mean the duration has to be shown without rounding off for instance if the duration is 91mn it should display "01:31"??

rgaudin:
@mahakporwal02, I'm not the one who opened the ticket. @kelson42 did.

kelson42:
My ticket complains about two things:
* rough rounding (legitimate)
* values, but I misunderstood how it was working. After @rgaudin’s comment I unerstand better (not legitimate)

Concerning the rounding, I want to precise my request. The values given here need:
* to easily give a good sense of the duration
* has to be easily compareable to previous runs

Therefore, I believe we should print durations with « two decimals », which gives in the hour system something like:
* 2d 2h
* 5h 30m
* 30m 2s
* 10s

So no rounding anymore, but scientific precision truncating.

Implementing https://github.com/openzim/zimfarm/issues/486 would help as well to better understand these numbers.

@mahakporwal02 Hope this helps

mahakporwal02:
![Screenshot from 2020-11-01 00-39-23](https://user-images.githubusercontent.com/56486682/97787875-c19c5280-1bda-11eb-8911-5325e5e2a751.png)
@rgaudin @kelson42 Is this fine?
Can I make a pull request?

rgaudin:
Looks good

[...]

PR #544, opened 2020-10-26
added duration rounding to taskView

mahakporwal02:
## Rationale
@rgaudin This solves  #487 
I have made changes in `task_duration` function in TaskView.vue and I have made a new function in constants.js to calculate duration between two timestamps and have added the durations as required 

## Changes
Added a new function `duration_between` in constants.js and modified `task_duration` in TaskView.vue

### DIFF
--- a/dispatcher/frontend-ui/src/views/TaskView.vue
+++ b/dispatcher/frontend-ui/src/views/TaskView.vue
@@ -157,6 +157,16 @@
         if (this.is_running) {
           return Constants.format_duration_between(first, Constants.now());
         }
+        var dur=0;
+        if(this.task.files&&Object.keys(this.task.files).length>0)
+        {
+            for (var file in this.task.files)
+            {
+              dur=dur+Constants.duration_between(this.task.timestamp.scraper_started, this.task.files[file].created_timestamp);
+            }
+            
+            return Constants.format_duration(dur);
+        }
 
         // if task is not running, it's started to last status
         let last = this.task.updated_at;
--- a/dispatcher/frontend-ui/src/constants.js
+++ b/dispatcher/frontend-ui/src/constants.js
@@ -18,6 +18,11 @@ function format_duration_between(start, end) { // display a duration between two
   return format_duration(diff);
 }
 
+function duration_between(start,end) {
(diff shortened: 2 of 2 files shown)


mahakporwal02:
@rgaudin any updates??

mahakporwal02:
@rgaudin any updates?

rgaudin:
@mahakporwal02 since this is a UI issue, can you please post an example of what it looks like with your code so that @kelson42 can confirm if it solves his issue or not.

dispatcher/frontend-ui/src/views/TaskView.vue
```diff
@@ -157,6 +157,16 @@
         if (this.is_running) {
           return Constants.format_duration_between(first, Constants.now());
         }
+        var dur=0;
+        if(this.task.files&&Object.keys(this.task.files).length>0)
```
rgaudin, line 161: Note that we bundle sugarjs so you can use it's Array.sum

mahakporwal02:
Screenshot from 2020-10-29 12-10-48
Screenshot from 2020-10-29 12-12-07
@rgaudin 
as we can see duration is shown correctly as the sum of 22mins+10mins=32mins and 5hrs+3hrs=8hrs respectively

dispatcher/frontend-ui/src/views/TaskView.vue
mahakporwal02: I'll make the required changes

rgaudin:
Ok then I think tou misunderstood the issue.
What we want to fix is the rounding, not how durations are calculated.
In current view, if the duration is 91mn (1h31) it will be rounded to "2 hours" by the humanize function. Same goes for days which are rounded probably after 13h (not sure).
that's what needs a fix

mahakporwal02:
@rgaudin So does this mean the duration has to be shown without rounding off for instance if the duration is 91mn it should display "01:31"??

kelson42:
@mahakporwal02 If @rgaudin agrees, I believe to have explicited quite clearly what is expected in my last comment on the issue ticket.

rgaudin:
Closing this as this is not what the ticket it about

Issue #461, opened 2020-03-23
feature request: support farm.o.o/worker#<workername>


JensKorte:
I don't know, if this is possible with dynamicly generated pages.

It would be nice, if I could jump to or highlight the worker(s) that match the anchor in the URL.

e.g.: https://farm.openzim.org/workers#hum would jump to the worker hum (exact hit) or it would highlight all worker containing "hum".

Nearly the same can be done with the search function in the browser, but with #hum it would be a permanent URL that can be bookmarked and shared.

mahakporwal02:
@rgaudin Can I work on this issue??

kelson42:
Looks good!

rgaudin:
Sure! Please let us know how/what you plan to do first as this is not clear what's exactly expected from this ticket.

mahakporwal02:
I plan to extract the string after hash('#') in url after '/workers' then find that substring in the worker's name, if the substring is present then a css class will be applied on that substring which will change the background behind that substring to a different color

kelson42:
@mahakporwal02 This isnot how itshould work IMO, read thi for example http://www.echoecho.com/htmllinks08.htm

mahakporwal02:
@kelson42 should the webpage just scroll down to worker name or should the worker name be highlighted as well?

kelson42:
@mahakporwal02 Jumping to seems good enough. If you can do high,ight as well, that might be nice... But I don't know how this would be doable in a simple manner (with a simple CSS rule). Therefore, probably not worse it.

PR #550, opened 2020-11-06
jump to a worker in anchor in url and highlight it

mahakporwal02:
## Rationale

#461 
## Changes
I have added `scrollTo` function in `updated` in workersView as it was not scrolling by normal archoring in url as it was jumping to the worker even before the page was completely loaded,I have also added a bootstrap class to the worker name to show it highlighted

Now after the page is loaded it will jump to the worker with id same as anchor of url and will highlight it @kelson42 @rgaudin


rgaudin:
Thank you @mahakporwal02 ; works I expected (I guess). Two remarks:

- shouldn't you add links to worker names so that there are clickable? Having to type this manually kind of defeat the purpose I think.
- Can we do it without another external dependency? Dependencies are a nightmare to track and maintain and that feature seems both simple and relatively futile so I'd prefer we don't add another dep for it.

mahakporwal02:
@rgaudin 
1)I have made the required changes about clickable worker name
2)We have to use dep for scrolling in vue as the function has to be performed after the page being completed loaded

rgaudin:
@mahakporwal02 sorry your explanation is not enough. Look at what the dep does and check whether we can implement the part that interest us in our code.

Also, don't rewrite commits until your PR is accepted as it's impossible to review what's changed ; so I must re-review the whole thing

mahakporwal02:
@rgaudin pls review the changes

rgaudin:
@mahakporwal02 thank you for the update but it's not working for me. Highlighting works fine but it doesn't scroll…

mahakporwal02:
@rgaudin pls review it

rgaudin:
Thanks ; it's now working!
Can you please adjust code formatting so it's similar to the rest of the code base: spaces around operators ; no space before trailing `;` ; same indentation, comment on top of line.

Will merge as soon as it's done.

rgaudin (approved):

Issue #556, opened 2020-11-20
Delete a user of type worker, should clean the entry in the worker table


kelson42:
The problem currently is that we delete the user, but we still see the worker on the workers board.

mahakporwal02:
@kelson42  Can I work on this issue?

kelson42:
Good for me.

PR #565, opened 2020-11-27
deleting worker from workers list on deleting a user

mahakporwal02:
## Rationale

The worker does not delete from workers list even when its corresponding user is deleted

#556 
## Changes
Added code to delete worker when user is being deleted

### DIFF
--- a/dispatcher/backend/src/routes/users/user.py
+++ b/dispatcher/backend/src/routes/users/user.py
@@ -6,6 +6,7 @@ from werkzeug.security import generate_password_hash
 from marshmallow import ValidationError
 
 from common.mongo import Users
+from common.mongo import Workers
 from common.roles import get_role_for, ROLES
 from routes import authenticate, url_object_id, errors, require_perm
 from utils.token import AccessToken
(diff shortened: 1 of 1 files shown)


dispatcher/backend/src/routes/users/user.py
```diff
@@ -129,6 +130,8 @@ def patch(self, token: AccessToken.Payload, username: str):
     @url_object_id("username")
     def delete(self, token: AccessToken.Payload, username: str):
         # delete user
+        if get_role_for(Users().find_one({"username":username}).get("scope")) == "worker":
```
rgaudin, line 133: Inserting new code in between a comment and the code that's related to it is misleading.

dispatcher/backend/src/routes/users/user.py
```diff
@@ -6,6 +6,7 @@
 from marshmallow import ValidationError
 
 from common.mongo import Users
+from common.mongo import Workers
```
rgaudin, line 9: You should share that import with the previous one

dispatcher/backend/src/routes/users/user.py
rgaudin: You should use two steps so that the workers delete is only happening after the user is deleted

rgaudin: This will crash if there is no User with that username as `find_one` will return `None`

rgaudin: You didn't respect the black formatting (`pip install black` then `black <yourfile`)

rgaudin (changes requested):
Thanks ; strategy's good but there are a number of stuff to fix

dispatcher/backend/src/routes/users/user.py
mahakporwal02: @rgaudin I have done the specified changes

rgaudin (approved):

Issue #510, opened 2020-06-26
Web Recipe list filters should be in URL


kelson42:
This would allow to share URLs, create bookmarks, etc...

Abuhmadh:
Good

DanielMK01:
Hi @kelson42 ! I'm interested in working on this issue, but I would like clarification on a couple of things.

1) Which file(s) should be updated for this issue?
2) Is there any other set-up for the project required besides the procedure shown in the Contributing.md file?

Thanks!

kelson42:
@DanielMK01 Thank you for proposal but I'm not the lead developer here and the lead developer is away for a few days, but you will get a feedback lately Monday next week.

DanielMK01:
@kelson42 I see. Who is the lead developer?

mahakporwal02:
@kelson42 @rgaudin Can I work on this issue?

kelson42:
@mahakporwal02 I believe this is slightly more complex than what you have done so far. Please explain here your plan to fix it before starting the coding.

mahakporwal02:
@kelson42  First I am planning to add filters to url using `this.$router.replace` and when the page is reloaded I am accessing the queries from previous url using `this.$route.query.<filter>` and pushing the filters into `selected<filters>` function in `schedulesList.vue`

mahakporwal02:
@rgaudin This issue might have been left open

PR #572, opened 2020-12-05
web recipe filters added in url

mahakporwal02:
## Rationale

Web recipe filters added in url which will allow to store and copy the url and reload the page 

#510 
## Changes
I have pushed the selected categories in url and once the page is reloaded url retains the filters but selected categories become empty so I have extracted filters to url and added to selected categories in each of the four category

Now one can add the filters and reload the page and still the page will retain all the information about the filters being added

### DIFF
--- a/dispatcher/frontend-ui/src/components/SchedulesList.vue
+++ b/dispatcher/frontend-ui/src/components/SchedulesList.vue
@@ -98,12 +98,10 @@
 
 <script type="text/javascript">
   import Multiselect from 'vue-multiselect'
-
   import Constants from '../constants.js'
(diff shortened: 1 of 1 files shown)


mahakporwal02:
Screenshot from 2020-12-11 23-39-25
@kelson @rgaudin the url looks like this when filters are added

rgaudin (changes requested):
Thanks but there's too much duplicated code. Please try and find a way to factorize all this.
Also, please respect the code formatting (again)

rgaudin:
Closed in favor of #587

PR #576, opened 2020-12-08
Don't allow trailing and leading spaces in recipe name

### LINKED ISSUE #458: Don't allow trailing space on recipe name

mahakporwal02:
## Rationale
#458 

Fixes #458 

#458 
## Changes
Changed validation format in schema for recipe name so that there are no trailing spaces added, it also makes sure the recipe name is found even if there is a typo(of blank space) in filters(name)

Now even if recipe name contains trailing or preceding spacing in a new recipe it will trim the spaces


rgaudin (approved):
thanks ; simplified it a bit

Issue #420, opened 2020-02-21
feature request: zimfarm.o.o: make workers table sortable


JensKorte:
It would be nice if the tables at <https://farm.openzim.org/workers> could be sorted ascending/descending by clicking on the header. The worker names have no header, maybe this could be added.

mahakporwal02:
![Screenshot from 2020-12-11 23-28-13](https://user-images.githubusercontent.com/56486682/101937941-93ae2300-3c08-11eb-98c1-469849bbcd25.png)
@rgaudin @JensKorte

kelson42:
@mahakporwal02 I would recommend to use a library which does that out of th box to avoid reinventing the wheel. Something like https://mottie.github.io/tablesorter/docs/

kelson42:
After discussion with @mahakporwal02 and @rgaudin it seem quite difficult to implement this properly and the added value is really limited (for the moment. Closing the issue as WONTFIX.

PR #580, opened 2020-12-11
made workers list sortable

mahakporwal02:
## Rationale
#420 
Workers list has been made sortable by clicking on header(worker name)

#420 
## Changes
added `sort_workers` to `workers view` to sort in ascending/descending order

### DIFF
--- a/dispatcher/frontend-ui/src/views/WorkersView.vue
+++ b/dispatcher/frontend-ui/src/views/WorkersView.vue
@@ -25,7 +25,20 @@
                   @click.prevent="toggle_workerslist">{{ toggle_text }}</b-button>
       </div>
     </div>
+    <div style="padding:15px">
+    </div>
     <table v-if="!error && workers" class="table table-responsive-md table-sm table-striped table-hover">
+      <thead>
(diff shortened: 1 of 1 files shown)


mahakporwal02:
@kelson42 The external library you recommended was unable to sort two columns (worker and its respective tasks), also @rgaudin suggested not to use external library as far as possible, so I have made changes that would work in the same way

rgaudin:
Thank you for the PR but I don't think it fixes the mentioned issue. It only allows to sort workers by name which has limited value to me.
Actually, I don't understand the rationale behind the issue. What problem should it solve?

As you've noticed, this table is composed of multiple tables and sorting by resources is gonna be tedious and is not defined.

Before adding more complexity to this code, I'd prefer we define proper use cases and maybe we redo that page altogether. It hasn't changed since we introduced it back when the zimfarm was launched. So there's probably a lot of room for improvements.

@kelson42 you probably are the biggest (if not the only) user of that page. What do you think?

kelson42:
@rgaudin Think twice about the ticket, I tend to agree. Sorry for the lost of time, but will close the issue as won't fix.

Issue #588, opened 2020-12-26
“Hide offline workers” should last longer


kelson42:
If I click on this button, it applies properly, ie. offline workers disappear.

Unfortunately after around a day, the offline workers reappear automatically.

It should not, as long as the button is not clicked again (to display offline workers).

I believe this is handled by a cookie which has an expiration time.... and it should not.

kelson42:
@mahakporwal02 Would you volunteer to fix that one?

mahakporwal02:
@kelson42 Sure

PR #590, opened 2020-12-29
cookie should last longer

mahakporwal02:
#588 

Offline workers reappear automatically

#588 
## Changes
Added expiration time set as 10 years as expiration time cannot be set as infinity.

Now the workers will not appear automatically as expiration time is far long.

### DIFF
--- a/dispatcher/frontend-ui/src/components/Mixins.js
+++ b/dispatcher/frontend-ui/src/components/Mixins.js
@@ -56,7 +56,7 @@ export default {
       return this.$cookie.get('pref-limit') || Constants.DEFAULT_LIMIT;
     },
     saveLimitPreference(value) {  // save nb of items to display in a cookie
-      this.$cookie.set('pref-limit', value);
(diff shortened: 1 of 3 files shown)


kelson42:
@mahakporwal02 Thank you for your quick fix.
@rgaudin would that make sense to have manybe a kind of coherent strategy regrading cookies expiration time?

rgaudin:
> @rgaudin would that make sense to have manybe a kind of coherent strategy regrading cookies expiration time?

We have 3 cookies:
- pref-limit, for your limit on number of items in pipeline/recipes
- token, for the authentication. this expires after 6months
- onlines-only, for the workers view. With this fix would *never expire* (10y)

Token one is though-through but if you have suggestions for the other two, this PR would be a good opportunity to set them.
We could also use a single preference cookie for both values but that would require some additional changes.

kelson42:
@mahakporwal02 Would you able to add a cookie lifetime configuration (or constant) and rely on it for the two cookies @rgaudin is talking about?

mahakporwal02:
@rgaudin @kelson42 OK, I'll fix it

rgaudin (approved):

Issue #358, opened 2019-12-31
Selectbox for recipes Docker image tag


kelson42:
It is currently a free text imput. It should propose/allow only the tags available on Docker Hub. This should be doable through a simple API call.

rgaudin:
It's on purpose so we can set it to whatever we want, just as for the image.

What would be the added value of restricting it? Users would only change that knowing what they're doing or cloning from an existing working recipe.

kelson42:
> What would be the added value of restricting it? Users would only change that knowing what they're doing or cloning from an existing working recipe.

Avoiding to make typos in tags. Such a solution for the image does not sound a bad idea either (does Sloppy works that way already?)

kelson42:
Here is a typical API call:
https://hub.docker.com/v2/repositories/openzim/mwoffliner/tags/?page=1&page_size=250

Could we create a vue.js selectbox component which would print the current value, and if we click on it, dynamically populate the list based on this kind of call?

kelson42:
@mahakporwal02 This might be an other potential task interesting for you.

PR #605, opened 2021-01-28
added selectbox for image tags

mahakporwal02:
## Rationale

To avoid typos in image tag.

#358 
## Changes
Free text input changed to selectbox input in image tags in edit tab.

### DIFF
--- a/dispatcher/backend/docs/openapi_v1.yaml
+++ b/dispatcher/backend/docs/openapi_v1.yaml
@@ -432,23 +432,17 @@ paths:
             application/json:
               schema:
                 type: object
-                example:
-                  {
-                    "result":
+                properties:
+                  meta:
+                    $ref: '#/components/schemas/MetaProperty'
+                  items:
+                    type: array
+                    example:
                       [
                         "latest",
-                        "dev",
                         "1.11.3",
                         "1.11.2",
-                        "1.11.1",
-                        "1.11.0",
-                        "1.10.12",
-                        "1.10.11",
-                        "1.10.10",
-                        "1.10.9"
                       ]
-                  }
-                description: Image Tags for corresponding Image Name
   /schedules/{scheduleName}/clone:
     post:
       tags:
--- a/dispatcher/backend/docs/openapi_v1.yaml
+++ b/dispatcher/backend/docs/openapi_v1.yaml
@@ -434,7 +434,7 @@ paths:
                 type: object
                 example:
                   {
-                    "result": 
+                    "result":
                       [
                         "latest",
                         "dev",
--- a/dispatcher/backend/docs/openapi_v1.yaml
+++ b/dispatcher/backend/docs/openapi_v1.yaml
@@ -406,6 +406,26 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/Error'
+  /schedules/{scheduleName}/image-names:
+    get:
+      tags:
+      - restricted
+      summary: get image tags  
+      operationId: getImageTags
+      description: Retrieve Image Tags
+      security:
+        - token: []
+        - oauth: []
+      parameters:
+      - $ref: '#/components/parameters/ScheduleNameParameter'
+      - $ref: '#/components/parameters/ImageNameParameter'
+      responses:
+        200:
+          description: List of Image Tags
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/ImageTags'          
   /schedules/{scheduleName}/clone:
     post:
       tags:
@@ -1554,6 +1574,14 @@ components:
       name: scheduleName
       schema:
         $ref: '#/components/schemas/ScheduleName'
+    ImageNameParameter:
+      in: query
+      name: hub_name
+      description: Docker Hub image name
+      required: true
+      schema:
+        type: string
+        example: openzim/mwoffliner
     ScheduleNameQueryParameter:
       in: query
       required: false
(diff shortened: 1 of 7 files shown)


mahakporwal02:
Screenshot from 2021-01-28 16-58-33

rgaudin (changes requested):
Unfortunately, this doesn't work as hub.docker.com doesn't respond to OPTIONS request, making the GET request fail.

I tried to get the information from Github as well but this is not available via the REST API and the GraphQL one seems buggy:

```graphql
{
  organization(login: "openzim") {
    packages(packageType: DOCKER) {
      totalCount
    }
  }
}
```

returns nothing

```json
{
  "data": {
    "organization": {
      "packages": {
        "totalCount": 0
      }
    }
  }
}
```

Not sure what to do without proxying this.

mahakporwal02:
@kelson42 I have requested the image tags as suggested by you using an API request but it turns out that docker doesn't respond to OPTIONS request and those tags are sent by browsers as a part of CORS.
@rgaudin suggested altenatives 
1)serve those tags by our python API (proxying dockerhub as we can do a regular GET in python)
2) request the tags in JS outside an XHR request: add a tag in the html document with the docker hub url as source, register an event for when this loads and then read its content.
Is this something that can be applied or pls suggest what else can be done?

mahakporwal02:
@kelson42 @rgaudin  any updates, please?

kelson42:
I propose:
* First report the bug to Docker and wait the feedback
* Otherwise the call should be relayed without CORS in Python

rgaudin:
@kelson42 I've found tickets in GH Community mentioning this issue. GHCR containers not appearing on the GraphQL API.
GH [is aware of that and is working on it](https://github.community/t/github-container-registry-ghcr-io-packages-not-appearing-in-api/130133/12) although there is no ETA and the *DELETE* REST endpoint they mentioned as the first one that would arrive is already late.

So that leaves us with the python proxy. @mahakporwal02 you'll have to expose a new API endpoint on the backend and retrieve it from there.

kelson42:
@rgaudin what abour Dovker hub?

rgaudin:
> @rgaudin what abour Docker hub?

Sorry, there was a misunderstanding. The docker API endpoint is not *buggy* but intentionally doesn't support CORS.
It can easily be enabled as documented in this [previous version doc](https://github.com/docker-archive/docker-registry/blob/master/ADVANCED.md#cors) and mentioned in https://github.com/distribution/distribution/issues/39 but what transpires is that they won't enable it on docker hub.

mahakporwal02:
@rgaudin Sorry I had to rebase the code and thats why I have force-pushed.
And sorry for the late PR

rgaudin:
That's good, thanks!

Could you please:
- add the new endpoint to the `openapi_v1.yaml` for the documentation
- add a test for the new endpoint so we don't decrease code coverage

Let me know if you need any assistance for those

dispatcher/backend/docs/openapi_v1.yaml
```diff
@@ -1,7 +1,7 @@
 openapi: 3.0.0
 servers:
   - description: production server
-    url: https://api.farm.openzim.org/v1
+    url: http://192.168.29.252:8000/v1
```
rgaudin, line 4: You left your debug url

dispatcher/backend/docs/openapi_v1.yaml
```diff
@@ -1531,8 +1551,8 @@ components:
       description: OAuth2
       flows:
         password:
-          tokenUrl: https://api.farm.openzim.org/v1/auth/oauth2
-          refreshUrl: https://api.farm.openzim.org/v1/auth/oauth2
+          tokenUrl: http://192.168.29.252:8000/v1/auth/oauth2
```
rgaudin, line 1554: again

dispatcher/backend/docs/openapi_v1.yaml
```diff
@@ -1531,8 +1551,8 @@ components:
...
+          refreshUrl: http://192.168.29.252:8000/v1/auth/oauth2
```
rgaudin, line 1555: again

dispatcher/backend/docs/openapi_v1.yaml
```diff
@@ -1554,6 +1574,14 @@ components:
       name: scheduleName
       schema:
         $ref: '#/components/schemas/ScheduleName'
+    ImageNameParameter:
```
rgaudin, line 1577: This won't be reused so it's better to keep it in the route definition

dispatcher/backend/docs/openapi_v1.yaml
```diff
@@ -1860,6 +1888,49 @@ components:
       type: integer
       format: int32
       description: Disk space (in bytes) required to run it
+    ImageTags:
```
rgaudin, line 1891: This type `object` is useless if it's not defined. Having this with an overly complex example prevents the developer from making any assumption about the returned data… while it's exactly the objective of the documentation.

This description should be accurate which means we also have to maintain it so that's why I suggest the API returns only what we need (the name)

dispatcher/backend/src/routes/schedules/schedule.py
```diff
@@ -204,6 +205,26 @@ def delete(self, schedule_name: str, token: AccessToken.Payload):
...
+    methods = ["GET"]
+
+    @authenticate
+    @require_perm("schedules", "update")
+    def get(self, schedule_name: str, token: AccessToken.Payload):
+        """ proxy list of tags from docker hub """
+        request_args = request.args.to_dict()
+        hub_name = request_args.get("hub_name")
+        try:
+            data = requests.get(
```
rgaudin, line 220: We should get that and just return a list of the tag names. There is no point in forwarding all that

dispatcher/backend/src/tests/integration/routes/schedules/test_schedule.py
```diff
@@ -279,6 +280,19 @@ def test_create_schedule_missing_config_keys(self, client, access_token, key):
...
+        response = client.get(
+            url,
+            headers={"Authorization": access_token},
+            query_string={"hub_name": "openzim/mwoffliner"},
+        )
+        assert response.status_code == 200
+        response = json.loads(response.data)
+        assert len(response["results"]) > 0
+        for item in response["results"]:
+            assert isinstance(item["name"], str)
```
rgaudin, line 294: might change with simplication of responde

dispatcher/frontend-ui/src/components/ScheduleEditor.vue
```diff
@@ -475,6 +479,23 @@
...
+        const parent = this;
+        const image_name = parent.edit_schedule.config.image.name;
+        const parts = image_name.split("/");
+        if (parts.length < 2) {
+           return; // invalid image_name
+        }
+        const hub_name = parts.splice(parts.length - 2, parts.length).join("/");
+        parent.queryAPI('get', '/schedules/' + this.schedule_name + '/image-names', {params: {hub_name: hub_name}})
+        .then(function (response) {
+          parent.image_tags = response.data.results.map(function (tag) { return tag.name; });
```
rgaudin, line 492: would change with simplification of response

rgaudin (changes requested):
Thank you ; documentation made me realize we don't want to just forward the whole docker hub response but just return the list of tag names that we got.

rgaudin (changes requested):
Great!

Could you please just use a format that matches the other responses?

```json
{
    "meta": {"skip": 0, "limit": 30, "count": 2},
    "items": ["latest", "1.10.1"],
}
```

It's just formatting as this `result` key doesn't match the rest.
As we don't really care about pagination, just sets the `skip` to `0` and have a `limit = 30`. Pass `?n=30` to the docker hub request so that it's coherent though.

rgaudin (approved):
Thank you @mahakporwal02 !

PR #647, opened 2021-10-14
Use Zim tools 3.0.0

kelson42:
This is necessary to benefit from the better/faster latest zimcheck.


rgaudin:
Ah sorry, I saw the github notification and though that was a request issue and not a PR ; I already had the changes in my source tree so I commit and pushed

Issue #757, opened 2023-03-30
Backend: Support JWT signing secret as environment variable 


benoit74:
The Flask backend generates a new JWT signing secret at every startup.

It means that at every restart, all existing access tokens are immediately invalidated.

This is particularly painful in development, but probably makes little sense in production as well.

This signing secret should be initialized from an (optional) environment variable.

kelson42:
As a user, I confirme this... and that this is really annoying :)

PR #758, opened 2023-03-30
Backend: Support JWT signing secret as environment variable

benoit74:
## Rationale

Fixes: #757 

## Changes

- added the `JWT_SECRET` as optional environment variable to initialize the secret used to sign JWT access token

### DIFF
--- a/dispatcher/backend/src/utils/token.py
+++ b/dispatcher/backend/src/utils/token.py
@@ -1,5 +1,6 @@
 import datetime
 import json
+import os
 import random
 import string
 import uuid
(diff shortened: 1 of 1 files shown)


rgaudin:
- Link the issue with the PR using proper Github wording (`Fixes` or `Fixed`) or via the UI (development icon on the right)
- Don't include the issue number in the PR title. Has little value (it's already in the commit) and makes titles too long

rgaudin (dismissed):
Awaiting green checks 😅

benoit74:
I'm not sure that adding the .flake8 file is the preferred way to go, but I had issues with flake8 and black competing about the appropriate line length

PR #759, opened 2023-03-30
Code formating

benoit74:
## Rationale

Deploy same code formating guidelines as other projects + fix some issue linked to unusual isort behavior not detected beforehand.

## Changes

- Fix formating issues (mainly linked to `isort`)
- Create missing `requirements.txt` for few projects to make dependencies clearer
- Fix Github Actions to run isort + flake8 + black


rgaudin:
> NOTA: I failed to configured .pre-commit-config.yaml, it fails with very strange errors, I propose to start without it for now, it will have little impact to include it afterwards (while all these code changes should be merged quickly IMO)

Ok 👍

.github/workflows/qa.yml
```diff
@@ -0,0 +1,49 @@
...
+
+on: [push, pull_request]
+
+env:
+  # black default
+  MAX_LINE_LENGTH: 88
+
+jobs:
+  check-qa:
+    runs-on: ubuntu-20.04
```
rgaudin, line 11:
```suggestion
    runs-on: ubuntu-22.04
```

.github/workflows/qa.yml
```diff
@@ -0,0 +1,49 @@
...
+    steps:
+      - uses: actions/checkout@v2
```
rgaudin, line 13:
```suggestion
      - uses: actions/checkout@v3
```

.github/workflows/qa.yml
```diff
@@ -0,0 +1,49 @@
...
+      - name: Set up Python ${{ matrix.python }}
+        uses: actions/setup-python@v1
```
rgaudin, line 15:
```suggestion
        uses: actions/setup-python@v3
```

.github/workflows/qa.yml
```diff
@@ -0,0 +1,49 @@
...
+        with:
+          python-version: 3.8
```
rgaudin, line 17:
```suggestion
          python-version: "3.11"
```

I think our images ship with py3.8 but I think we want to default to 3.11 soon

dispatcher/backend/.flake8
```diff
@@ -0,0 +1,5 @@
+# .flake8
+
+[flake8]
+max-line-length = 88
+extend-ignore = E203
```
rgaudin, line 5: We can add the extend-ignore to the flake8 command (` --extend-ignore=E203`) and thus not need this file at all

rgaudin (changes requested):
Great a couple of suggestions mostly to use latest versions of actions everytime we introduce a new workflow

dispatcher/backend/.flake8
rgaudin: Hum, I have a similar setup in `~/.flake8` but that's because those are what we use everywhere. Would that work for you?
I am not against having the file in the repo but then it hides this behavior from within the files that call the flake8 binary… Can be solved with a comment maybe

benoit74: Yes, good point, I will remove these files from the repo

rgaudin (approved):
Perfect ; thanks

PR #774, opened 2023-05-08
Allow disabling users and workers

### LINKED ISSUE #763: Allow disabling users and workers
For archive reason, we don't want to ever delete user accounts as those can be linked to data that is still valuable.
Should a user not be allowed to authenticate anymore, we should be able to _disable_ it.

Same goes for Workers which are versatile by nature. We should be able to mark them disabled and keep them in the database.

Shall be implemented along #317 

**Note**: this ticket is only about technical capability (database column), not about UI to handle those capabilities

benoit74:
## Rationale

Fix #763 

## Changes

- add a `deleted` column for `worker` and `user` DB table
- fix APIs to not display `deleted` users and workers, as explained in the issue  + block `deleted` users authentication
- create tests for users `passwords` and `keys`
- modify the `workers` test fixture so that 2 workers have a duplicate IP and 1 has no IP yet, so that `test_build_workers_whitelist` test is a bit more realistic / complete
- added automatic creation of test DB in dev container


dispatcher/backend/src/tests/integration/routes/workers/test_worker.py
```diff
@@ -6,7 +6,7 @@
 class TestWorkersCommon:
     def test_build_workers_whitelist(self, workers):
         whitelist = build_workers_whitelist()
-        assert len(whitelist) == len(workers)
+        assert len(whitelist) == 36
```
rgaudin, line 9: - ideally, we'd have a variable with the expected number
- static value in a test should be explained. How can next developer looking be sure about this or update it should fixture changed?
- don't change a test without mentioning is clearly; otherwise it looks like you broke a test and entered whatever value made the test pass witout fixing the issue 😅

dev/backend-tools-tests/Dockerfile
```diff
@@ -2,7 +2,7 @@ FROM python:3.8-alpine
 LABEL zimfarm=true
 LABEL org.opencontainers.image.source https://github.com/openzim/zimfarm
 
-RUN apk add openssl gcc libffi-dev libc-dev
+RUN apk add openssl gcc libffi-dev libc-dev openssh-keygen
```
rgaudin, line 5: I don't see the value of having this in this container…

rgaudin (changes requested):
Thank you ; it looks good.

It seems a frequent pattern is to retrieve the User (or worker) and then test whether it is `deleted` and raise should it be.

We'd benefit from expanding `dbm.User.get_or_none` (same for worker) with a param like `deleted_ok: Optional[bool] = False`.

Please also fix the QA

dev/backend-tools-tests/Dockerfile
benoit74: This is needed when setting a user SSH key to generate its PKCS8 : https://github.com/openzim/zimfarm/blob/ac18dd3738e0d386f80c52c10979319776746dc6/dispatcher/backend/src/routes/users/keys.py#L93

dispatcher/backend/src/tests/integration/routes/workers/test_worker.py
benoit74: Agreed. Just to confirm my understanding, where did you expect me to mention it? In the PR, right?

dev/backend-tools-tests/Dockerfile
rgaudin: OK ; I don't recall seeing tests for the SSH key endpoint.

benoit74:
FYI, I finally expanded the "checking" logic to all classes, so that it is more straightforward.
The QA issue is kinda fun, my editor (Visual Studio Code) is not apply the same parameters for flake + isort when the file is named `ssh.py`. Still didn't found the cause ...

dispatcher/backend/src/db/models.py
```diff
@@ -243,10 +304,30 @@ class Task(Base):
...
+        exception_class: Type[Exception] = Exception,
+        *exception_args: object,
+    ) -> "Task":
+        """Search DB for a task by UUID
+
+        If the check of the task is not ok, raise thes exception passed in parameters"""
         stmt = select(Task).where(Task.id == id)
-        return session.execute(stmt).scalar_one_or_none()
+        task = session.execute(stmt).scalar_one_or_none()
+        cls.check(task, exception_class, *exception_args)
```
rgaudin, line 329: If this is ran unconditionally, I see no reason to have a separate method. It seems that we don't call this `.check()` anywhere else so I think we can simplify here

rgaudin (changes requested):
Thank you ; I think moving the checks to the getter is a good move and the dedicated method makes sense for `User` but for the others I only see extra glue.

benoit74:
Agreed 👍

rgaudin (approved):
All Good 👍

Issue #317, opened 2019-11-15
Switch MongoDB to RDMS for storage


rgaudin:
* MongoDB consumes too much RAM
* our use case doesn't need/benefit from this data being in RAM
* RAM is expensive on VPS and sloppy
* disk storage is cheap

for all those reason, I believe we shall switch from mongo to mariadb. once done, memory usage should be lower and more controllable.

Just make sure that task creation is atomic to ensure only a single worker is assigned to a task.

`pymysql` and `peewee` should be sufficient.

kelson42:
Needed to proper backup.

kelson42:
Backup is secured, but we still need to move the MongoDB content to the MariaDB instance.

rgaudin:
Refreshing this ticket more than 3 years later:

- zimfarm has a single DB, for its API (dispatcher)
- it's a MongoDB so it's fully in RAM and synced to disk
- At this very moment, mongo consumes 6GiB of RAM and periodically gets restarted by k8s/kernel due to OOM
- DB takes up 1GiB on disk

That's for the issue we face.

Now MongoDB was chosen by previous ZF developer and we never really questioned it until its RAM usage became problematic.

It's important to note that we have no MongoDB expertise. Our instance is not [configured](https://www.mongodb.com/docs/manual/reference/configuration-options/#storage-options). We use the official image (4.2.9) and its defaults.

It's also important to note that we don't really need a schemaless DB. Most of what we store is code-driven and the main (only?) flexible part is the flags in the config. We could store that in a JSON field.

Using Mongo also means a different code, different tools, different backup scenarios, etc. And writing aggreation queries is difficult and not self-explanatory. So it doesn't make sense to use it if we don't benefit from its main feature and it's the only one in a large collection of projects (actually the cardshop API also uses mongo but will switch as well).

For maintenance sake, we want to use a single type of database across our projects (this doen't exclude KV store like redis where see fits) and we've decided to use PostgreSQL because of its wide support and reputation for saefty and performances.

Key tasks:
- identify a db stack. We've tried thing with the cms but we're not happy with it.
- we want an ORM because that's more convenient and we want to abstract SQL so we may switch to a different engine in the future
- we want one that can play nicely with async frameworks such as fastapi
- we want to be able to input SQL when needed
- we want a schema migration tools
- ORM schema
- code must be ported to the new DB
- update backup
- migration plan
- deployment

The most important task here being choosing the python stack because we will use those in other projects and we want this sorted once and for a good while.

[...]

PR #780, opened 2023-05-12
Issue 317 step 5

benoit74:
## Rationale

Part of #317, migrating to PG

## Changes

- replace `url_object_id` by `url_uuid` + fail if parsing is not ok + remove all useless usage


rgaudin (approved):
👍

PR #782, opened 2023-05-12
Cleanup all mongo dependencies

### LINKED ISSUE #781: Cleanup after RDBMS migration to PG
Some cleanup has to be performed after some time spent with the database migrated to PostgreSQL

- [x] fix documentation in CONTRIBUTING.md (remove reference to Mongo)
- [x] remove mongo_val + mongo_id columns in database
- [x] remove mongo and mongoexpress from dev docker-compose
- [x] remove .github/workflows/db-migration-docker.yaml workflow

benoit74:
## Rationale

Fix #781 

## Changes

- everything described in #781 
- we still keep a requirement on pymongo because we need it to have bson utils which are used for serialization / deserialization of JSON columns in PG ; these utils are used to properly handle serialization/deserialization of naive datetimes

⚠️ **DO NOT MERGE NOW**


dispatcher/backend/requirements.txt
```diff
@@ -2,7 +2,8 @@ Flask>=2.3.2,<2.4
 flask-cors>=3.0,<3.1
 paramiko>=3.1,<3.2
 PyJWT>=2.6,<2.7
-pymongo==3.12.0
+# pymongo is still a requirement because used for bson serializer/deserializer
+pymongo==3.12.0
```
rgaudin, line 6: Then use `bson` directly

rgaudin (approved):
Good but please use only `bson` dep and take a look at my comment regarding the fixture

dispatcher/backend/requirements.txt
benoit74: If you speak about https://pypi.org/project/bson/, I have to warned you that this is not the official package maintained by Mongo. It is a fork by some folk http://github.com/py-bson/bson and looks not maintained at all.
I saw this package but decided to not use it due to the associated risks (no maintenance, potential risk of takeover by someone that would introduce a security flaw).
Are you aware of all this and do you still consider this is the way to go ?

dispatcher/backend/src/tests/integration/conftest.py
```diff
@@ -177,3 +171,12 @@ def pytest_sessionstart(session):
     before performing collection and entering the run test loop.
     """
     Initializer.check_if_schema_is_up_to_date()
+    with Session.begin() as session:
```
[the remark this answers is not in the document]
benoit74: It is true that not being able to detect left overs is an issue.
I added this because I used the DB for something else and it hit me in the face.
But you are true that this is probably not a good idea. Thank you.

benoit74:
This is still a draft only to avoid accidental merge. This must be merged only once we are so confident that the DB migration is OK / completed that we can delete backup data and other left-overs.

dispatcher/backend/requirements.txt
rgaudin: I forgot! Thanks for the reminder. Can you mention it in the comment here so that we don't accidentally replace it later ; it's uncommon that a single package installs multiple modules.

Also, there are traces of `ObjectId` in `utils/json.py` and `utils/token.py` that can be removed

PR #800, opened 2023-07-06
Fix remove_secrets_from_response with a more robust approach

### LINKED ISSUE #799: Secrets are not hidden properly in API responses
## Problem

The API is hiding secrets in `config.flags` but not anymore in `config.command` or `container.command`.

Probably a regression linked to the move to PG database + the hot fix https://github.com/openzim/zimfarm/commit/c68f06b7ef5ea06ee1ed91deef66f3d4c6b3d2a2

The API is also not hiding secrets in `upload.logs.upload_uri`, but we cannot completely remove this parameter from responses since it is used to build the download URL in the UI. We will have to assume it is an S3 URL and that we have to hide `keyid` and `secretaccesskey`.

benoit74:
## Rationale

Fix #799 

## Changes

- more robust / generic code to cope with many situations
- add code to also clean `upload.logs.upload_uri`
- add tests

### DIFF
--- a/dispatcher/backend/src/routes/utils.py
+++ b/dispatcher/backend/src/routes/utils.py
@@ -1,34 +1,85 @@
+import logging
+from urllib.parse import urlparse
+
 from common.constants import SECRET_REPLACEMENT
 from common.schemas.models import ScheduleConfigSchema
 from utils.offliners import build_str_command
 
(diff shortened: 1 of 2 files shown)


rgaudin (approved):
LGTM ; thanks!

PR #807, opened 2023-07-25
Adapt maint scripts to PG migration - continued

benoit74:
- adapt command to add `PYTHON_PATH` so that dispatcher classes are accessible
- adapt requirements + sample environment variable to Postgres migration


rgaudin (approved):
LGTM

PR #808, opened 2023-07-27
Fix secrets removal in UI + add support for long-description in kolibri

### LINKED ISSUE #803: Secrets are still not hided properly
- Location: __API__

## Problem

Secrets are still not hided properly by #799, at least not in `container.command`.

We had doubts regarding `config.command` that must not be hidden when you are logged in with sufficient rights, so we reverted the change to ensure there is no impact for workers.

### LINKED ISSUE #805: Add support for `--long-description` parameter for kolibri
`kolibri` `1.1.0` has just been [released](https://github.com/openzim/kolibri/releases/tag/v1.1.0) and has added support for the  optional `--long-description` CLI parameter which must be configured in Zimfarm now.

benoit74:
## Rationale

Fix #805 

Fix #803 

## Changes

- added support for `--long-description` CLI parameter for `kolibri`
- fixed the removal of secret in commands to support more cases
  - flags where OK
  - issue was linked to the fact tasks are returning command with `param=secret_value`  instead of `param="secret_value"`
  - secret removal now supports commands with `param="secret_value"` (as before) but also (new) `param=secret_value` and `param='secret_value'` 
  - tests have been added accordingly to exercise these "new" situations
  - no issue related to authenticated users has been observed, they have access to the plain secret value

### DIFF
--- a/dispatcher/backend/src/common/schemas/offliners/kolibri.py
+++ b/dispatcher/backend/src/common/schemas/offliners/kolibri.py
@@ -41,12 +41,23 @@ class KolibriFlagsSchema(SerializableSchema):
             "description": "Custom title for your ZIM. Kolibri channel name otherwise",
         }
     )
+
     description = fields.String(
         metadata={
             "label": "Description",
             "description": "Custom description for your ZIM. "
             "Kolibri channel description otherwise",
         }
+    )
+
+    long_description = fields.String(
+        metadata={
+            "label": "Long description",
+            "description": "Custom long description for your ZIM. "
+            "If not provided, either not set or Kolibri channel description if it was "
+            "too long to fit entirely in ZIM description",
+        },
+        data_key="long-description",
     )
 
     favicon = fields.Url(
--- a/dispatcher/backend/src/routes/utils.py
+++ b/dispatcher/backend/src/routes/utils.py
@@ -10,12 +10,12 @@ logger = logging.getLogger(__name__)
 
 def remove_secrets_from_response(response: dict):
     """replaces or removes (in-place) all occurences of secrets"""
-    remove_upload_secrets_from_flags_and_commands(response)
-    remove_upload_secrets_from_response(response)
+    remove_secret_flags(response)
+    remove_upload_secrets(response)
 
 
-def remove_upload_secrets_from_flags_and_commands(response: dict):
-    """replaces (in-place) all occurences of secrets in flags/commands with stars"""
+def remove_secret_flags(response: dict):
+    """replaces (in-place) all occurences of secret flags with stars"""
 
     # we need flags to retrieve the secret value and replace it where we find it
     # in commands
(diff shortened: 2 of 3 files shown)


dispatcher/backend/src/common/schemas/offliners/kolibri.py
```diff
@@ -49,6 +50,16 @@ class Meta:
         }
     )
 
+    long_description = fields.String(
```
rgaudin, line 53: We should probably open a ticket to switch all long-description fields to a textarea with 4K chars limit

dispatcher/backend/src/routes/utils.py
```diff
@@ -1,34 +1,92 @@
...
+    response: dict, root_key_value: str, field_name: str, secret_value: str
+):
+    if "command" not in response[root_key_value]:
+        return
+    if secret_value not in response[root_key_value]["command"]:
+        return
+    index = response[root_key_value]["command"].index(secret_value)
+    response[root_key_value]["command"][
+        index
+    ] = f'--{field_name}="{SECRET_REPLACEMENT}"'
```
rgaudin, line 71: If `index` is the position in list and contains `secret_value`, then assigning `--{field-name}="secret"` to `command[index]` will result in a list with the following I think

```py
command [ 
    ...
    '--field-name',
    '--field-name="secret"',
    …
]
```

rgaudin (changes requested):
I'm not sure 🤔

dispatcher/backend/src/common/schemas/offliners/kolibri.py
benoit74: I do not see any long_description field in any other scraper. 

I suggest that we create a ticket to:
- move the logic regarding truncating / defaults of description / long_description that we developed in `kolibri` to `pythonscraperlib` so that it is shared across scrapers
- add support for both description / long_description CLI flags in all python scrapers
- add support for textarea in the UI for flags fields
- limit field maximum length in the UI (I'm not sure about this one, is there any validation logic implemented in the UI ? maybe sufficient to do this in the backend when validating the payload received)

WDYT?

rgaudin: > I do not see any long_description field in any other scraper.

Not sure if it's in zimfarm but I remember it being supported elsewhere (nautilus at least). `mwoffliner` (not python) does support it in the zimfarm.
 
> I suggest that we create multiple issues to:
> 
>     * move the logic regarding truncating / defaults of description / long_description that we developed in `kolibri` to `pythonscraperlib` so that it is shared across scrapers

Should support both cases: with or without a potential long description.

>     * add support for both description / long_description CLI flags in all python scrapers

Description is already everywhere ; but needs to be updated. Would require updating scraperlib to latest version.

>     * add support for textarea in the UI for flags fields

👍 

>     * limit field maximum length in the UI (I'm not sure about this one, is there any validation logic implemented in the UI ? maybe sufficient to do this in the backend when validating the payload received)

We'd check in the backend for sure. For frontend, textarea/input's `maxlength` is good enough.

benoit74: Opened :
- https://github.com/openzim/zimfarm/issues/810
- https://github.com/openzim/python-scraperlib/issues/110
- https://github.com/openzim/overview/issues/21

dispatcher/backend/src/routes/utils.py
rgaudin: You're right ; it only applies to command which has this format.

rgaudin (approved):
LGTM

Issue #817, opened 2023-09-17
Multiple tasks have not reached a final state `failed`, `canceled` or `succeeded`


benoit74:
- Location: API

## Problem
In [worker view](https://farm.openzim.org/workers), failed / cancelled tasks are shown as still there, while they are not running anymore. Looks like it does not have an impact on scheduling, it is more a UI issue. This is probably a regression.

![image](https://github.com/openzim/zimfarm/assets/7102089/6810391d-169e-46a5-8f81-1eabf7de9d26)

Please do not cleanup the situation until we have a look into it to fully understand the situation

benoit74:
I confirm this does **not** affect scheduling, worker performance or workload.

What happens is that:
- to display running tasks, the UI is requesting the `GET /tasks` endpoint filtering for tasks with a status among `reserved`, `started`, `scraper_started`, `scraper_completed`, `scraper_killed` or `cancel_requested` 
  - i.e. anything not `requested`, `failed`, `canceled` or `succeeded`
  - `scraper_running` is also a potential task status which is sent by worker but it is never persisted to the DB)
- we have many tasks which did not reached those final states `failed`, `canceled` or `succeeded` in the DB and are hence incorrectly reported in the view

I will continue to investigate the exact list of tasks with improper status and the underlying reason.

benoit74:
This is the list of problematic tasks. Many kind of scrapers, many workers, many time slots. Zimfarm is still operating properly, so it looks a lot like "lost messages".

| Task ID | worker | kind | DB Status | last update at | requested | reserved | started | scraper_started | scraper_completed |
|--|--|--|--|--|--|--|--|--|--|
| [e54a4a26-1273-4608-92a7-bac9d826813b](https://farm.openzim.org/pipeline/e54a4a26-1273-4608-92a7-bac9d826813b) | athena18 | sotoki | scraper_started | 2023-08-29T16:02:41 | 2023-08-29T16:00:13 | 2023-08-29T16:02:21 | 2023-08-29T16:02:24 | 2023-08-29T16:02:41 | - |
| [59f8ef9b-c608-47ef-84ec-7641a46c6fe8](https://farm.openzim.org/pipeline/59f8ef9b-c608-47ef-84ec-7641a46c6fe8) | chance | zimit | scraper_started | 2023-08-30T16:01:40 | 2023-08-30T16:00:01 | 2023-08-30T16:01:33 | 2023-08-30T16:01:38 | 2023-08-30T16:01:40 | - |
[...]

PR #820, opened 2023-09-19
Add timeouts for GHCR + CMS requests

benoit74:
## Rationale

See #817 for details

## Changes

- add timeout to requests to GHCR + CMS
- timeout values have a default of 10 secs, which can be override by an environment variable
- more homogeneity in environment variables / constants naming across all requests timeouts


rgaudin (approved):
Great ; can you please update zimfarm config on k8s to specify all those (to make tweaking config easier) and also (in that config) specify a 60s timeout for cms to balance it's current perf issue

PR #827, opened 2023-09-22
dispatcher/backend: Remove zmq broadcaster + enhance CMS error handling

benoit74:
## Rationale / Change

- Do not try to decode JSON if HTTP status code is an error
- Remove unused zmq broadcaster from code path (potential source of issues + totally unused)


rgaudin (changes requested):
Nope, that's not what we want. This will prevent storing the actual status code in the DB which we want to retry the request later.

You could add an `if` before the `try` block to stop it from adding logs on something we know is not useful

benoit74:
I fixed the original change as you suggested (sorry, I read the code too fast).
I removed all zmq stuff as discussed.

rgaudin (approved):
LGTM 👍

PR #833, opened 2023-09-28
Fix periodic scheduling + most_recent_task update

### LINKED ISSUE #813: multiple parralel tasks for dp.la_en_all
Many things are not understandable to me:
* We have a task of it running since a month https://farm.openzim.org/pipeline/60faed9f-264a-490c-9919-e33b96637623, so we have a Zimit bug here?
* when inlook to the recipe https://farm.openzim.org/recipes/dp.la_en_all, it seems we have 3 times the same scrape running in parallel, but this has not been scheduled manually?!
* I see we had 3 times success within the same hours. But the schedule is monthly?!

benoit74:
## Rationale

Fix #813 

## Changes
- schedules `most_recent_task` property is set only when the task is created (`reserved` event)
- periodic scheduler does not request a new task when the `most_recent_task` is not completed

NB: as discussed in the issue, setting schedules `most_recent_task` property when the task is requested is not feasible or at least a very significant change, including in terms of DB schema adaptation ; so let's update it only when the task is requested, this is in line with current behavior + this event is never lost (it is automatically triggered when the requested task is transformed into a task, and the worker won't start the task without a success on this operation).


dispatcher/backend/src/tests/integration/routes/business_logic/test_periodic_scheduling.py
```diff
@@ -0,0 +1,101 @@
...
+            schedule_obj = dbm.Schedule.get(session, temp_schedule["name"])
+            schedule_obj.most_recent_task = task
+        self.assert_number_of_requested_tasks_for_schedule(temp_schedule, 0)
+        request_tasks_using_schedule()
+        self.assert_number_of_requested_tasks_for_schedule(temp_schedule, 0)
+
+    def test_periodic_scheduling_most_recent_overdue_complete(
+        self, temp_schedule, make_task
+    ):
+        """most recent task is overdue but still running => request a new execution"""
```
rgaudin, line 62: I believe the docstring is incorrect (**not** still running)

dispatcher/backend/src/tests/integration/routes/business_logic/test_task_event_handler.py
```diff
@@ -0,0 +1,24 @@
+import db.models as dbm
+from common.enum import TaskStatus
+from common.utils import task_event_handler
+from db import Session
+
+
+class TestTaskEvents:
```
rgaudin, line 7: I think both test names are not precise enough about what's being tested

dispatcher/backend/src/tests/integration/routes/conftest.py
```diff
@@ -93,6 +96,31 @@ def schedule(make_schedule):
     return make_schedule()
 
 
+@pytest.fixture
+def temp_schedule(make_schedule):
```
rgaudin, line 100: How is this specific? Tests data are not supposed to be persisted so all of them are _temporary_.

rgaudin (changes requested):
Thank you ; please see inline comments (minor precisions)

dispatcher/backend/src/tests/integration/routes/conftest.py
benoit74: Other test data are issued by a fixture with scope "module" ; this one has a scope "function", meaning it will be recreated for each test case (avoiding side-effects from other test cases since we are modifying the data quite heavily).

dispatcher/backend/src/tests/integration/routes/business_logic/test_task_event_handler.py
benoit74: Fixed, is it better?
I find it's a bit long now, but I'm ok with it
FYI, I chose those vague names because I expected those tests to be enriched in the future, but it's true that it might never happen

dispatcher/backend/src/tests/integration/routes/conftest.py
rgaudin: OK, I guess it's helpful in this way.

dispatcher/backend/src/tests/integration/routes/business_logic/test_task_event_handler.py
rgaudin: Yep

rgaudin (approved):
👍

PR #840, opened 2023-09-29
Add is_requested property to schedule endpoints

### LINKED ISSUE #720: Schedule column always empty under certain conditions
Recipe view looks good with pagination at 10
A123B034-FB26-4D87-B05A-332077BA2A8A

But with pagination at 50, the schedule column is empty
51A8F65F-7DF9-4496-9405-40CD13856C23

benoit74:
## Rationale

Fix #720 

## Changes

- new `is_requested` property on `GET /schedules/` and `GET /schedules/{schedule_name}` endpoints
- for `GET /schedules/`, property is computed based on an addition to the complex SQL Query retrieving schedules and their properties
- for `GET /schedules/{schedule_name}`, property is computed based on an additional SQL Query (transparent) to retrieve requested_tasks of current schedule (not an issue since we have only one schedule and rarely a requested_task)
- UI does not use the `GET /requested_tasks` endpoint anymore, column is sourced directly from `is_requested` property

### DIFF
--- a/dispatcher/backend/docs/openapi_v1.yaml
+++ b/dispatcher/backend/docs/openapi_v1.yaml
@@ -1722,6 +1722,7 @@ components:
         - language
         - name
         - category
+        - is_requested
       properties:
         config:
           type: object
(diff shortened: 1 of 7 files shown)


rgaudin (approved):
LGTM ; thanks 👍

PR #843, opened 2023-10-06
Enhance youtube flags and enhance StringEnum UI handling

benoit74:
## Changes
- In the schedule config UI, when the flag type is "StringEnum", the "Not set" option is not shown in the combobox when the flag is required
- Youtube flags are re-order so that:
  - important parameter are displayed first in the list
  - flags which are specific to one mode (normal or playlists) are grouped together
  - expert flags are grouped at the end of the list
- Youtube flags descriptions are enhanced:
  - flags specific to one mode (normal or playlists) are clearly indicated
  - no technical flag identifier is used in descritpion, only flag display names are used
  - subtle wording changes, more precisions, ...
- API key is set a secret

### DIFF
--- a/dispatcher/backend/src/common/schemas/offliners/youtube.py
+++ b/dispatcher/backend/src/common/schemas/offliners/youtube.py
@@ -12,6 +12,26 @@ class YoutubeFlagsSchema(SerializableSchema):
     class Meta:
         ordered = True
 
+    optimization_cache = fields.Url(
+        metadata={
+            "label": "Optimization Cache URL",
+            "description": "Technical Flag: S3 Storage URL including credentials and "
+            "bucket",
+            "secret": True,
+        },
+        data_key="optimization-cache",
+    )
+
+    api_key = String(
+        metadata={
+            "label": "API Key",
+            "description": "Technical flag: Youtube API Token",
+            "secret": True,
+        },
+        data_key="api-key",
+        required=True,
+    )
+
     indiv_playlists = fields.Boolean(
         truthy=[True],
         falsy=[False],
@@ -54,7 +74,7 @@ class YoutubeFlagsSchema(SerializableSchema):
     name = String(
         metadata={
             "label": "ZIM Name",
-            "description": "Normal mode: Used as identifier and filename (date will "
+            "description": "Single mode: Used as identifier and filename (date will "
             "be appended)",
             "placeholder": "mychannel_eng_all",
         },
@@ -63,7 +83,7 @@ class YoutubeFlagsSchema(SerializableSchema):
     zim_file = String(
         metadata={
             "label": "ZIM Filename",
-            "description": "Normal mode: ZIM file name (optional, based on ZIM Name "
+            "description": "Single mode: ZIM file name (optional, based on ZIM Name "
             "if not provided). Include {period} to insert date period dynamically",
         },
         data_key="zim-file",
@@ -73,7 +93,7 @@ class YoutubeFlagsSchema(SerializableSchema):
     title = String(
         metadata={
             "label": "ZIM Title",
-            "description": "Normal mode: Custom title for your project and ZIM. "
+            "description": "Single mode: Custom title for your ZIM. "
             "Default to Channel name (of first video if playlists)",
         }
     )
@@ -81,7 +101,7 @@ class YoutubeFlagsSchema(SerializableSchema):
     description = String(
         metadata={
             "label": "ZIM Description",
-            "description": "Normal mode: Description for ZIM",
+            "description": "Single mode: Description for ZIM",
         },
         validate=validate_zim_description,
     )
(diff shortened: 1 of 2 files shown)


dispatcher/backend/src/common/schemas/offliners/youtube.py
```diff
@@ -17,74 +17,136 @@ class Meta:
...
+            "description": "Normal mode: ZIM file name (optional, based on ZIM Name "
+            "if not provided). Include {period} to insert date period dynamically",
+        },
+        data_key="zim-file",
+        validate=validate_zim_filename,
+    )
+
+    title = String(
+        metadata={
+            "label": "ZIM Title",
```
rgaudin, line 75: What's the project? Is it used anywhere else beside the metadata?

dispatcher/backend/src/common/schemas/offliners/youtube.py
```diff
@@ -17,74 +17,136 @@ class Meta:
...
+            "description": "Normal mode: Used as identifier and filename (date will "
+            "be appended)",
             "placeholder": "mychannel_eng_all",
         },
     )
+
+    zim_file = String(
+        metadata={
+            "label": "ZIM Filename",
+            "description": "Normal mode: ZIM file name (optional, based on ZIM Name "
```
rgaudin, line 66: I think “normal mode” might be misleading. Can we change it to “Single mode” instead?

dispatcher/backend/src/common/schemas/offliners/youtube.py
```diff
@@ -139,146 +213,105 @@ class Meta:
...
 
-    profile = fields.Url(
+    api_key = String(
         metadata={
-            "label": "Profile Image",
-            "description": "Custom profile image. Squared. "
-            "Will be resized to 100x100px",
-        }
+            "label": "API Key",
+            "description": "Expert flag: Youtube API Token",
```
rgaudin, line 268: This is not an expert flag. It's mandatory and must be chosen carefully (otherwise it might break tons of tasks). I'm afraid marking this as _Expert_ will have Content People not touch it and we'll just duplicate over and over while they should manage those keys

dispatcher/backend/src/common/schemas/offliners/youtube.py
```diff
@@ -139,146 +213,105 @@ class Meta:
...
     )
-    banner = fields.Url(
+
+    optimization_cache = fields.Url(
         metadata={
-            "label": "Banner Image",
-            "description": "Custom banner image. Will be resized to 1060x175px",
-        }
+            "label": "Optimization Cache URL",
+            "description": "Expert Flag: S3 Storage URL including credentials and "
```
rgaudin, line 278: Not an expert field. Content people hopefully knows what this is and why it should always be set and according to the scraper. It's not edited often though and is a technical flag but should be checked.

dispatcher/frontend-ui/src/components/ScheduleEditor.vue
```diff
@@ -329,7 +329,9 @@
           if (field.type == "string-enum") {
             component = "b-form-select";
             options = field.choices.map(function (option) { return {text: option, value: option}; });
-            options.push({text: "Not set", value: undefined});
+            if (field.required != true) {
+              options.push({text: "Not set", value: undefined});
```
rgaudin, line 333: Please make sure to check zimit-frontend about this. I'm think this part is duplicated so it might need to be backported there as well

rgaudin (changes requested):
Thank you ; it looks good but there's a few questions

dispatcher/backend/src/common/schemas/offliners/youtube.py
benoit74: I wasn't aware that we are using multiple API keys. If so, then we must not mark it as Expert.

dispatcher/frontend-ui/src/components/ScheduleEditor.vue
benoit74: https://github.com/openzim/zimit-frontend/pull/38

dispatcher/backend/src/common/schemas/offliners/youtube.py
benoit74: No idea, it was here before I came. I assumed it had some meaning for you, but obviously it doesn't so I removed it.

dispatcher/backend/src/common/schemas/offliners/youtube.py
benoit74: Moved up front, it is important indeed you are right

dispatcher/backend/src/common/schemas/offliners/youtube.py
benoit74: Moved up front, it is too important indeed

rgaudin (approved):
Great!

PR #858, opened 2023-10-30
Fix worker IP update to use one single DB session

### LINKED ISSUE #857: Worker IP change issue
`michaelblob` had tons of tasks fail very early because its IP was not in the Wasabi whitelist.

- we should not let a _failing worker_ (considering that all its tasks are failing early) to continue to get new tasks.
- we should harden the IP change process.

On that second point:
- We've had an issue in the past, after the change to postgres which was linked to the new IP not being recorded to DB at time of generation of policy.
- `michaelblob`'s IP seems dynamic. 4d ago, it was already failing due to IP not in whitelist with `173.73.128.55`. Now it's failing for same reason with IP `70.108.9.176`.
- New IP was already recorded in DB when I checked (17:30) but not in whitelist.
- Whitelist had been modified on Oct 25, 2023, 2:09 AM. Don't know which worker triggered it. We should log that.
- We limit each worker to 4 IP changes per day but I can't find any trace of this happening in grafana
- Only Call to policy update was this one so it succeeded.

From those initial information it looks like a similar bug to the previous one (assuming michaelblob triggered that Oct25 change): we saw the IP change, recorded it but for some reason the IP is not in the updated list.
This matches with my `record_ip_change("michaelblob")` fixing it.

benoit74:
## Rationale

Fix #857

## Changes (original ones, see below for updated list)

- use on single DB session instead of two which obviously leads to inconsistent results
- add log information about which worker has changed IP, including previous and new IP
- refresh `USES_WORKERS_IPS_WHITELIST` dynamically so that it can be manipulated in tests
- removed previous fix on this topic, since it was not working / needed
- simplified code paths when comparing / setting IPs

## Remarks
- I'm not a huge fan of the dynamic refresh of `USES_WORKERS_IPS_WHITELIST` but this is the less intrusive change I found to keep this in `constants.py` since this file is useful to have the list of available environment variables
- I've done a quick and dirty change with the `IpUpdater` class ; this should not be done like this, we should have a cleaner code architecture + use a mocking library, but again, this is not a small change
- as mentioned in the issue, I fixed only the bug, the other part should be moved to a specific issue from my PoV

### DIFF
--- a/dispatcher/backend/src/common/external.py
+++ b/dispatcher/backend/src/common/external.py
@@ -29,7 +29,7 @@ logger = logging.getLogger(__name__)
 
 def update_workers_whitelist(session: so.Session):
     """update whitelist of workers on external services"""
-    ExternalIpUpdater.update_fn(build_workers_whitelist(session=session))
+    ExternalIpUpdater.update(build_workers_whitelist(session=session))
 
 
 def build_workers_whitelist(session: so.Session) -> typing.List[str]:
@@ -150,7 +150,13 @@ def update_wasabi_whitelist(ip_addresses: typing.List):
 
 
 class ExternalIpUpdater:
-    update_fn = update_wasabi_whitelist
+    """Class responsible to push IP updates to external system(s)
+
+    `update` is called with the new list of all workers IPs everytime
+    a change is detected.
+    By default, this class update our IPs whitelist in Wasabi"""
+
+    update = update_wasabi_whitelist
 
 
 @dbsession
--- a/dispatcher/backend/src/common/external.py
+++ b/dispatcher/backend/src/common/external.py
@@ -27,12 +27,11 @@ logging.basicConfig(level=logging.DEBUG)
 logger = logging.getLogger(__name__)
 
 
-def update_workers_whitelist():
+def update_workers_whitelist(session: so.Session):
     """update whitelist of workers on external services"""
-    update_wasabi_whitelist(build_workers_whitelist())
+    IpUpdater.update_fn(build_workers_whitelist(session=session))
 
 
-@dbsession
 def build_workers_whitelist(session: so.Session) -> typing.List[str]:
     """list of worker IP adresses and networks (text) to use as whitelist"""
     wl_networks = []
@@ -148,6 +147,10 @@ def update_wasabi_whitelist(ip_addresses: typing.List):
         PolicyDocument=new_policy,
         SetAsDefault=True,
     )
+
+
+class IpUpdater:
+    update_fn = update_wasabi_whitelist
 
 
 @dbsession
--- a/dispatcher/backend/src/common/external.py
+++ b/dispatcher/backend/src/common/external.py
@@ -29,7 +29,7 @@ logger = logging.getLogger(__name__)
 
 def update_workers_whitelist(session: so.Session):
     """update whitelist of workers on external services"""
-    IpUpdater.update_fn(build_workers_whitelist(session=session))
+    ExternalIpUpdater.update_fn(build_workers_whitelist(session=session))
 
 
 def build_workers_whitelist(session: so.Session) -> typing.List[str]:
@@ -149,7 +149,7 @@ def update_wasabi_whitelist(ip_addresses: typing.List):
     )
 
 
-class IpUpdater:
+class ExternalIpUpdater:
     update_fn = update_wasabi_whitelist
 
 
--- a/dispatcher/backend/src/routes/requested_tasks/requested_task.py
+++ b/dispatcher/backend/src/routes/requested_tasks/requested_task.py
@@ -24,8 +24,8 @@ from common.schemas.parameters import (
     WorkerRequestedTaskSchema,
 )
 from common.utils import task_event_handler
-from db import count_from_stmt, dbsession
-from errors.http import InvalidRequestJSON, TaskNotFound, WorkerNotFound
+from db import count_from_stmt, dbsession, dbsession_manual
+from errors.http import InvalidRequestJSON, TaskNotFound, WorkerNotFound, HTTPBase
 from routes import auth_info_if_supplied, authenticate, require_perm, url_uuid
 from routes.base import BaseRoute
 from routes.errors import NotFound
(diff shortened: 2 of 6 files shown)


dispatcher/backend/src/tests/integration/routes/workers/test_worker.py
```diff
@@ -206,3 +209,59 @@ def test_checkin_another_user(
...
+                "worker": worker["name"],
+                "avail_cpu": 4,
+                "avail_memory": 2048,
+                "avail_disk": 4096,
+            },
+            headers={"Authorization": make_access_token(worker["username"], "worker")},
+        )
+        assert response.status_code == 200
+
+    new_ip_address = "88.88.88.88"
```
rgaudin, line 241: That's unexpected. Place assignements on top of class for it's easily identifiable

dispatcher/backend/src/tests/integration/routes/workers/test_worker.py
```diff
@@ -206,3 +209,59 @@ def test_checkin_another_user(
...
+
+    def custom_ip_update(self, ip_addresses: List):
+        self.ip_updated = True
+        assert TestWorkerRequestedTasks.new_ip_address in ip_addresses
+
+    def test_requested_task_worker_update_ip_whitelist(
+        self, client, make_access_token, worker
+    ):
+        self.ip_updated = False
```
rgaudin, line 250: As always, confusing to declare instance variables inside a method. Makes reading difficult. Especially in this context of self-contained tests

rgaudin (changes requested):
I don't like this implementation neither.

First `refreshable_constant` does noting. It simply returns what is passed. This is confusing and should be avoided.
You may want to look at functools.partial I think.

Next is this faux-dynamic refresh of environ. It's not possible to change the environment of a running process so this brings additional confusion and only _works_ in your tests because pytest fires processes.
Testing should mostly adapt to codebase or require refactoring but here it's a detrimental change of codebase to accomodate tests.

Another thing is I think we wont write the IP to the DB should the policy update fail (it's possible, it's a remote call). Are we OK with this? Maybe a comment would be helpful.

benoit74:
I honestly don't know how to fix code to match your expectations. As stated, existing code architecture prevented me to find a clean and simple way to run tests. And I didn't knew about the impossibility to update a process environment, my bad.

The only solution I see to move forward would be to delete the tests, since it looks they are detrimental to code quality. It makes me very sad because I really prefer working and tested code to clean code. But that's life and I don't have time to refactor the Zimfarm codebase, at least I'm not supposed to do it as far as I've understood.

I agree about the fact that we wont write the IP to the DB should the policy update fail. Was already the case before, still the case now. Rather than a comment, I think we should try/catch the remote IP update to ensure data is persisted or commit the transaction before making the remote IP update.

Feel free to take over this PR and finish the work if it is simpler.

rgaudin:
Can you describe the problem/issue? Maybe we can think of a better way to solve it

benoit74:
This is my issue (only for one test):
- I want to ensure IPs sent to code handling Wasabi policies is correct (assuming this code is correct)
- for this I need to get result of `build_workers_whitelist` to check its value
- I would like to not call the method directly to stay in same DB session (otherwise bug is not triggered btw)
- so what I found is that I need to
  - set `USES_WORKERS_IPS_WHITELIST` environment variable to 1 so that IP whitelist code is triggered
  - capture result of IP computation to check its value
  - ensure that tests won't try to call Wasabi

benoit74:
https://github.com/openzim/zimfarm/issues/863 should probably be tackled at the same time

benoit74:
I almost completely rewrote the tests + added a short-term fix to persist IP changes asap in DB, before we can tackle #863 which is indeed not straightforward)

## Changes

- use on single DB session instead of two which leads to inconsistent results
- use a new kind of dbsession decorator which does not manage transaction for us ; this allows to call commit ourselves whenever we see fit
- commit IP changes to DB before calling Wasabi so that this is persisted no matter what happens later
- add log information about which worker has changed IP, including previous and new IP
- remove default value when computing `USES_WORKERS_IPS_WHITELIST`, useless
- simplified code paths when comparing / setting IPs

dispatcher/backend/src/routes/requested_tasks/requested_task.py
```diff
@@ -229,15 +225,26 @@ def get(self, session: so.Session, token: AccessToken.Payload):
...
+                    # quickly and transaction is not blocking
+                    session.commit()
+                    if constants.USES_WORKERS_IPS_WHITELIST:
+                        try:
+                            record_ip_change(session=session, worker_name=worker_name)
+                        except Exception:
+                            raise HTTPBase(
+                                status_code=HTTPStatus.SERVICE_UNAVAILABLE,
+                                error="Recording IP changes failed",
+                            )
```
rgaudin, line 247: Shouldn't we commit again after `fond_requested_task_for()` ?

dispatcher/backend/src/common/external.py
```diff
@@ -150,6 +149,10 @@ def get_statement():
     )
 
 
+class ExternalIpUpdater:
+    update_fn = update_wasabi_whitelist
```
rgaudin, line 153:
```suggestion
    update = update_wasabi_whitelist
```

Calling an `update_fn` function makes little sense.
Also please include a doctstring for this class

rgaudin (changes requested):
I'd also like my two comments from previous revision to be addressed ; either in discussion or in code.

dispatcher/backend/src/routes/requested_tasks/requested_task.py
benoit74: I do not expect `find_requested_task_for` to write any data to the DB, so I would prefer not to commit. This is a good safeguard to avoid making unexpected changes to the DB.

rgaudin: Indeed ; I tend to forgort that it just offers task and workers initiate the status change 👍

benoit74:
I rebased the commits on latest `main` changes maybe a bit too soon, I should have wait the end of the review, sorry.

I fixed the last comment which wasn't yet fixed, the code was dead indeed, I just forgot to remove it from the tests ... my bad, sorry again. 

AFAIK the other comment has already been fixed, I hope I'm not mistaken.

rgaudin (approved):
Thank you

Issue #844, opened 2023-10-12
Detect duplicate schedule names when cloning a schedule


benoit74:
- Location: API / UI
 
## Problem
When cloning a schedule, if the new chosen name is already in used, this is only detected at the DB level, raising a 500 error.

We should detect this at least at the API level, and maybe also at the UI level (effort to be investigated).

## Reproducing steps
Clone a schedule and choose a name which is already in use.

kelson42:
Duplicate of #740

benoit74:
This is not really a duplicate, I was only speaking about duplicated names. Not about duplicated configuration as in #740. This issue is probably way easier to solve than #740 and would avoid 500 errors

kelson42:
@benoit74 Today, I lost around 15 mn because of that issue. Considering we have a PR we should anyway be treated/improved/... and the impact is very "vicious" (old ZIM disappear from library.kiwix.org) Could you please fix this in priority?

benoit74:
@kelson42 I don't think your problem is the same as this issue. This issue (and corresponding PR) is speaking about duplicated **schedule** name. You seem to have an issue about duplicate **ZIM** name. This is a distinct problem far more complex to solve (please open an issue if you confirm your problem is a different problem). As indicated here, this issue is only about better handling of duplicate schedule name which are currently raising 500 errors. And this is seemed not a very important thing since it is **very rare** to happen in our history (we are being alarmed on every 500 error since months now).

benoit74:
@kelson42 I'm still waiting on your feedback about your recent issue. Can you clarify what made you loose 15 mn of your precious time so that we know what needs to be enhanced to not do it again? I'm quite sure your problem is duplicate ZIM name and not duplicate recipe name, but I would like to have your confirmation.

kelson42:
@benoit74 You are 100% right, this is a duplicate ZIM metadata "Name". My bad.

benoit74:
I isolated your specific issue here: https://github.com/openzim/zimfarm/issues/1037

elfkuzco:
Is this not resolved by #862 ? Or is it the pending issue of not differentiating the exceptions?

benoit74:
Yes it is going to be solved by #862. 

But we never achieved to merge this (not because it is complex, but because I was expecting Renaud to complete the effort, then Renaud handed over but I did not find time to complete the effort, and now the PR is so old it will not merged).

@elfkuzco can you please help finish this PR so that we can finally merge?

[...]

PR #862, opened 2023-11-02
raise HTTP 400 on IntegrityError during Clone

rgaudin:
Fixeds #844

### DIFF
--- a/dispatcher/backend/src/routes/schedules/schedule.py
+++ b/dispatcher/backend/src/routes/schedules/schedule.py
@@ -382,6 +382,9 @@ class ScheduleCloneRoute(BaseRoute):
         )
         clone.durations.append(duration)
 
-        session.flush()
+        try:
+            session.flush()
+        except IntegrityError:
+            raise BadRequest("Schedule name already exists")
 
         return make_response(jsonify({"_id": str(clone.id)}), HTTPStatus.CREATED)


dispatcher/backend/src/routes/schedules/schedule.py
```diff
@@ -382,6 +382,9 @@ def post(self, schedule_name: str, token: AccessToken.Payload, session: so.Sessi
         )
         clone.durations.append(duration)
 
-        session.flush()
+        try:
+            session.flush()
+        except IntegrityError:
+            raise BadRequest("Schedule name already exists")
```
benoit74, line 388: I find this a bit risky, I'm afraid we might have other integrity constraints in the future and might forget to alter this message.

At least I would log a warning with the exact integrity issue so that we can discover this in the log, otherwise it will be difficult to debug.

Or I would analyze IntegrityError, check the issue is about the schedule name, raise this BadRequest only in this situation ;  and raise a more generic IntegrityError for other situations (which do not exist today).

benoit74 (changes requested):
A test case is probably easy to add + would be beneficial.

kelson42:
@benoit74 This is a very old PR, it would be good to consoder to treat it

rgaudin:
We should probably close it. I think it was open in reaction to something that's not that important now

rgaudin:
Issue is sill valid

benoit74:
I think issue is still valid, and feedback has been provided in the review (quite minor, but still important). At this point I expected @rgaudin to finish the PR since he started the job. I don't mind to finish it if it is more appropriate.

rgaudin:
@benoit74 please take over

benoit74:
Superseeded by https://github.com/openzim/zimfarm/pull/1245

PR #865, opened 2023-11-06
Check schedule flags validity also when creating a schedule from scratch

### LINKED ISSUE #864: Schedule configuration flags must also be validated at schedule creation
- Location: API

## Problem

Schedule configuration flags have various constraints which must be checked. These constraints are check when the schedule configuration is updated (`PATCH /schedule/{schedule_id}`) but not when the schedule is created from scratch (`POST /schedule`). 

This has almost zero impact on farm.openzim.org since the UI does not allows to create schedules from scratch but it has a very significant one on youzim.it where the UI is always creating schedules from scratch (and most probably, same issue impact WP1, but I'm less familiar with it).

Once fixed, this will solve (first part of) https://github.com/openzim/zimit-frontend/issues/40

## Reproducing steps

Create (not clone and patch) a new schedule with a very long description (longer than 80 chars). It will succeed while it should fails.

benoit74:
## Rationale

Fix #864 

## Changes

- schedule flags are also checked when creating a recipe
- freecodecamp test case has been adapted to:
  - fix "normal" test case which was indeed not ok since mandatory flags were not provided at schedule creation
  - create an abnormal test case to exercise new check / avoid future regressions (this has been done on freecodecamp for simplicity, any scraper could be used indeed)
- an abnormal test case has also been created for the more generic schedule creation tests
- generic schedule creation tests have been fixed to ensure they are failing for the expected reason

### DIFF
--- a/dispatcher/backend/src/tests/integration/routes/schedules/test_freecodecamp.py
+++ b/dispatcher/backend/src/tests/integration/routes/schedules/test_freecodecamp.py
@@ -31,7 +31,6 @@ class TestFreeCodeCamp:
             url, json=schedule, headers={"Authorization": access_token}
         )
         response_data = response.get_json()
-        print(response_data)
         if "_id" in response_data:
             garbage_collector.add_schedule_id(response_data["_id"])
         assert response.status_code == 201
@@ -94,7 +93,6 @@ class TestFreeCodeCamp:
             url, json=schedule, headers={"Authorization": access_token}
         )
         response_data = response.get_json()
-        print(response_data)
         if "_id" in response_data:
             garbage_collector.add_schedule_id(response_data["_id"])
         assert response.status_code == 400
--- a/dispatcher/backend/src/tests/integration/routes/schedules/test_freecodecamp.py
+++ b/dispatcher/backend/src/tests/integration/routes/schedules/test_freecodecamp.py
@@ -1,9 +1,9 @@
 class TestFreeCodeCamp:
-    def test_create_freecodecamp_schedule(
+    def test_create_freecodecamp_schedule_ok(
         self, client, access_token, garbage_collector
     ):
         schedule = {
-            "name": "fcc_javascript_test",
+            "name": "fcc_javascript_test_ok",
             "category": "freecodecamp",
             "enabled": False,
             "tags": [],
@@ -14,7 +14,13 @@ class TestFreeCodeCamp:
                 "image": {"name": "openzim/freecodecamp", "tag": "1.0.0"},
                 "monitor": False,
                 "platform": None,
-                "flags": {},
+                "flags": {
+                    "course": ("somecourse"),
+                    "language": "eng",
+                    "name": "acourse",
+                    "title": "a title",
+                    "description": "a description",
+                },
                 "resources": {"cpu": 3, "memory": 1024, "disk": 0},
             },
             "periodicity": "quarterly",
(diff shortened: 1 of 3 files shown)


benoit74:
Indeed there is maybe a thing to enhance in this PR, we validate schemas twice now (once for global schema and once for offliner schema). 

This has the disadvantage that if we have two errors, one in the global schema and one in the offliner schema, the error message will only contain the error in the global schema at first. And once this is fixed, it will display the second error in the offliner schema. Maybe we should trigger both checks in all situations and merge the two lists of error messages before returning the errors.

dispatcher/backend/src/tests/integration/routes/schedules/test_freecodecamp.py
```diff
@@ -24,9 +30,11 @@ def test_create_freecodecamp_schedule(
         response = client.post(
             url, json=schedule, headers={"Authorization": access_token}
         )
-        assert response.status_code == 201
         response_data = response.get_json()
-        garbage_collector.add_schedule_id(response_data["_id"])
+        print(response_data)
```
rgaudin, line 34: debug print

dispatcher/backend/src/tests/integration/routes/schedules/test_freecodecamp.py
```diff
@@ -50,3 +58,49 @@ def test_create_freecodecamp_schedule(
...
+            },
+            "periodicity": "quarterly",
+        }
+
+        url = "/schedules/"
+        response = client.post(
+            url, json=schedule, headers={"Authorization": access_token}
+        )
+        response_data = response.get_json()
+        print(response_data)
```
rgaudin, line 97: Another one

dispatcher/backend/src/tests/integration/routes/schedules/test_schedule.py
```diff
@@ -265,15 +265,26 @@ def test_create_schedule_missing_keys(self, client, access_token, key):
...
+            "periodicity": "quarterly",
         }
 
         del schedule[key]
         url = "/schedules/"
         response = client.post(
             url, json=schedule, headers={"Authorization": access_token}
         )
+        response_data = response.get_json()
+        print(response_data)
```
rgaudin, line 280: :)

dispatcher/backend/src/tests/integration/routes/schedules/test_schedule.py
```diff
@@ -293,15 +304,64 @@ def test_create_schedule_missing_config_keys(self, client, access_token, key):
...
+            "periodicity": "quarterly",
         }
 
         del schedule["config"][key]
         url = "/schedules/"
         response = client.post(
             url, json=schedule, headers={"Authorization": access_token}
         )
+        response_data = response.get_json()
+        print(response_data)
```
rgaudin, line 319: :0

dispatcher/backend/src/tests/integration/routes/schedules/test_schedule.py
```diff
@@ -293,15 +304,64 @@ def test_create_schedule_missing_config_keys(self, client, access_token, key):
...
+            },
+            "periodicity": "quarterly",
+        }
+
+        url = "/schedules/"
+        response = client.post(
+            url, json=schedule, headers={"Authorization": access_token}
+        )
+        response_data = response.get_json()
+        print(response_data)
```
rgaudin, line 357: !

rgaudin (changes requested):
LGTM but you left a few debug prints

benoit74:
I left those debug print intentionally (to help should the test fail at some point), but I don't mind to remove them, I have no real PoV of whether this is a good idea or not. I find it useful to avoid to have to debug the test locally to understand what went wrong ... but in fact this is usually needed anyway. I will remove them for now.

rgaudin:
> I left those debug print intentionally (to help should the test fail at some point), but I don't mind to remove them, I have no real PoV of whether this is a good idea or not. I find it useful to avoid to have to debug the test locally to understand what went wrong ... but in fact this is usually needed anyway. I will remove them for now.

If those are useful, we can left them in. We shall document somewhere (bootstrap) that we allow this and add corresponding relaxing rule for tests (in other projects we lint the tests)

benoit74:
I'm not sure they are really useful. My point was only to stress this might be intentional ^^

rgaudin (approved):
👍

PR #875, opened 2023-11-27
Fix update of worker last_seen property

benoit74:
## Rationale
- since #857, the `last_seen` property of workers is not updated correctly anymore

## Changes
- always commit changes to DB, not only if worker IP has changed
- in tests, move stuff to fixtures + perform tests around `last_seen` and `last_ip` updates


dispatcher/backend/src/tests/integration/routes/workers/test_worker.py
```diff
@@ -212,32 +213,130 @@ def test_checkin_another_user(
...
+            return make_headers(access_token=access_token, client_ip=client_ip)
+
+        return _worker_headers
+
+    @pytest.fixture
+    def default_ip(self):
+        return "192.168.1.1"
+
+    @pytest.fixture
+    def increase_ip(self):
```
rgaudin, line 259: Why not using functions for fixture? Is it to keep this together with the tests that use them? Maybe it's time to split the file 😉

dispatcher/backend/src/routes/requested_tasks/requested_task.py
```diff
@@ -223,20 +223,27 @@ def get(self, session: so.Session, token: AccessToken.Payload):
...
                         f"Worker IP changed detected for {worker_name}: "
                         f"IP changed from {worker.last_ip} to {worker_ip}"
                     )
                     worker.last_ip = worker_ip
-                    # commit explicitely since we are not using an explicit transaction,
-                    # and do it before calling Wasabi so that changes are propagated
-                    # quickly and transaction is not blocking
-                    session.commit()
+
+                # commit explicitely since we are not using an explicit transaction,
```
rgaudin, line 241: I think this comment should mention that we commit in order for push the `last_seen` update. There's a reason we missed it and the fact the it's preceded by the IP code change (that has no reason to be committed unless changed) doesn't help

rgaudin (approved):
Good ; please update the comment and merge

dispatcher/backend/src/tests/integration/routes/workers/test_worker.py
benoit74: Thank you for confirming the idea of splitting the file was a good one 🤣

PR #884, opened 2023-12-12
Do not request all schedules details at once, proceed in small batches

### LINKED ISSUE #882: Stop overwhelming the API with hundreds of requests in parallel
- Location: U.I

## Problem

When displaying the failed pipeline, the UI has to request details about each schedule to display the last task. Currently the UI does all requests at once, meaning up to 200 requests in parallel (or even 400 with the preflight `OPTIONS`). This is probably the main root cause of https://github.com/kiwix/k8s/issues/149

## Reproducing steps

Display 200 failed recipes in the UI

## Change requested

Perform the requests in small batches of for instance 5 requests.

benoit74:
## Rationale

Fix #882 

## Changes

- when getting schedules information for the tasks details, we proceed in batches of 5 requests every 500 ms
- when the data is refreshed (scheduled refresh every minute, or when the number of tasks to display is changed), details about schedules are hidden until all data is loaded
-  Bump @babel/traverse from 7.12.12 to 7.23.2 in /dispatcher/frontend-ui
- Bump browserify-sign from 4.2.1 to 4.2.2 in /dispatcher/frontend-ui
- Bump axios from 0.21.2 to 1.6.0 in /dispatcher/frontend-ui

### DIFF
--- a/dispatcher/frontend-ui/src/components/PipelineTable.vue
+++ b/dispatcher/frontend-ui/src/components/PipelineTable.vue
@@ -180,12 +180,20 @@
       },
       async loadLastRuns() {
         let parent = this;
+        parent.last_runs_loaded = false;
         let schedule_names = parent.tasks.map(function (item) { return item.schedule_name; }).unique();
 
         parent.toggleLoader("fetching last runs…");
-        let requests = schedule_names.map(function (schedule_name) {
-            return parent.queryAPI('get', "/schedules/" + schedule_name);
-          });
+        const chunkSize = 5;
+        let requests = []
+        for (let i = 0; i < schedule_names.length; i += chunkSize) {
+            const chunk = schedule_names.slice(i, i + chunkSize);
+            for (let iChunk in chunk) {
+              requests.push(parent.queryAPI('get', "/schedules/" + chunk[iChunk]))
+            }
+            await Constants.delay(500)
+        }
+
         const results = await Promise.all(requests.map(p => p.catch(e => e)));
 
         results.forEach(function (response, index) {
--- a/dispatcher/frontend-ui/src/components/PipelineTable.vue
+++ b/dispatcher/frontend-ui/src/components/PipelineTable.vue
@@ -184,14 +184,14 @@
         let schedule_names = parent.tasks.map(function (item) { return item.schedule_name; }).unique();
 
         parent.toggleLoader("fetching last runs…");
-        const chunkSize = 5;
+        const chunkSize = Constants.TASKS_LOAD_SCHEDULES_CHUNK_SIZE;
         let requests = []
         for (let i = 0; i < schedule_names.length; i += chunkSize) {
             const chunk = schedule_names.slice(i, i + chunkSize);
             for (let iChunk in chunk) {
               requests.push(parent.queryAPI('get', "/schedules/" + chunk[iChunk]))
             }
-            await Constants.delay(500)
+            await Constants.getDelay(Constants.TASKS_LOAD_SCHEDULES_DELAY)
         }
 
         const results = await Promise.all(requests.map(p => p.catch(e => e)));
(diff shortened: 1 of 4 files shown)


dispatcher/frontend-ui/src/components/PipelineTable.vue
```diff
@@ -180,12 +180,20 @@
...
 
         parent.toggleLoader("fetching last runs…");
-        let requests = schedule_names.map(function (schedule_name) {
-            return parent.queryAPI('get', "/schedules/" + schedule_name);
-          });
+        const chunkSize = 5;
+        let requests = []
+        for (let i = 0; i < schedule_names.length; i += chunkSize) {
+            const chunk = schedule_names.slice(i, i + chunkSize);
+            for (let iChunk in chunk) {
```
rgaudin, line 191:
```suggestion
            for (let schedule_name in chunk) {
```

dispatcher/frontend-ui/src/constants.js
```diff
@@ -295,6 +295,10 @@ function get_timezone_details() {
   return {tz: dt.zoneName, offset: dt.o, offsetstr: offsetstr}
 }
 
+function delay(time) {
```
rgaudin, line 298: This is unclear. `time` means nothing. Is this milliseconds?

dispatcher/frontend-ui/src/constants.js
```diff
@@ -527,4 +531,5 @@ export default {
   fromSeconds: DateTime.fromSeconds,
   to_timestamp: to_timestamp,
   is_ios_firefox: isFirefoxOnIOS(),
+  delay: delay,
```
rgaudin, line 534: I think `getDelay` might be less ambiguous. WDYT?

dispatcher/frontend-ui/src/components/PipelineTable.vue
```diff
@@ -180,12 +180,20 @@
...
+              requests.push(parent.queryAPI('get', "/schedules/" + chunk[iChunk]))
+            }
+            await Constants.delay(500)
```
rgaudin, line 194: Is this 500milliseconds? Would that mean that it would spread the 200 requests over 20seconds ?
200 / 5 = 40 iterations
40 * 500ms = 20Kms = 20s
That seems crazy long

rgaudin (changes requested):

dispatcher/frontend-ui/src/components/PipelineTable.vue
benoit74: Just as crazy as making 200 requests for displaying one screen. What would you suggest?

dispatcher/frontend-ui/src/components/PipelineTable.vue
benoit74: It is not yet a schedule name, it is only an index in the chunk

dispatcher/frontend-ui/src/constants.js
benoit74: Fixed

dispatcher/frontend-ui/src/components/PipelineTable.vue
benoit74: I find it's clearer with the `iChunk` name

dispatcher/frontend-ui/src/components/PipelineTable.vue
rgaudin: If the value can't obviously set, it should be changeable with an env. I'd start with 100 and we'll see how it goes.

rgaudin: Making the nb of schedules in chunk changeable would help as well

dispatcher/frontend-ui/src/components/PipelineTable.vue
rgaudin: My bad, I thought it was a schedule name

dispatcher/frontend-ui/src/components/PipelineTable.vue
benoit74: Done

rgaudin (approved):
Good ; I'll let you merge once you've updated the ui configs so those values are explicit and can be changed easily

PR #901, opened 2024-01-17
Harmonize `Publisher` ZIM metadata

### LINKED ISSUE #900: Set default Publisher to `openZIM` in all offliners
As requested in https://github.com/openzim/zim-requests/issues/779, we must harmonize the default Publisher for all our ZIMs.

benoit74:
## Rationale

Fix #900 

## Changes

- new environment variable to activate the setting of a default `--publisher` flag in all offliners supporting it
- offliners definition defines which offliners are supporting the override (and potentially what is the proper flag name)
- if `--publisher` flag is manually set in the configuration, then the default value is not set


dispatcher/backend/src/utils/offliners.py
```diff
@@ -73,9 +76,35 @@ def command_for(offliner, flags, mount_point):
...
+
+    The "publisher" flag is set if a default is provided in the local environment, if
+    the scraper supports it, and if it is not already set manually.
+
+    The "publisher" flag might have a different name, configured in the offliner
+    definition.
+    """
+
+    flag_name = (
+        offliner_def.publisher_flag
```
rgaudin, line 96: There is none at the moment right?

dispatcher/backend/src/utils/offliners.py
```diff
@@ -8,22 +8,25 @@
...
-    Offliner.ifixit: od("ifixit2zim", True, True),
-    Offliner.mwoffliner: od("mwoffliner", "outputDirectory", False),
-    Offliner.youtube: od("youtube2zim-playlists", True, False),
-    Offliner.ted: od("ted2zim-multi", True, False),
-    Offliner.openedx: od("openedx2zim", True, False),
-    Offliner.nautilus: od("nautiluszim", True, False),
-    Offliner.zimit: od("zimit", True, "statsFilename"),
-    Offliner.kolibri: od("kolibri2zim", True, False),
+    Offliner.freecodecamp: od("fcc2zim", True, False, True),
+    Offliner.gutenberg: od("gutenberg2zim", False, False, False),
```
rgaudin, line 19: I guess that's because gutenberg has no publisher flag and this will change once it is released

rgaudin (approved):
LGTM

dispatcher/backend/src/utils/offliners.py
benoit74: Yes

dispatcher/backend/src/utils/offliners.py
benoit74: Yes

PR #917, opened 2024-02-22
Artifacts upload

### LINKED ISSUE #913: On-demand archiving of temporary files produced by scrapers
The reported issue is that it is very difficult to debug `zimit` issues on the Zimfarm because the WARC files that are produced by Browsertrix crawler are lost once the recipe is finished. 

This causes:
- long time and resources lost crawling again the same website
- uncertainties about the fact that the WARC files we use in local tests are really the same than the ones which caused the issue on the Zimfarm

On a normal basis, we do not need / want to store these WARC files, just like we do not store gutenberg temporary DB and other build assets. But it would be great that we could do it when we know there is a high probability that we will need the WARC files (e.g. when debugging Zimit 2).

As discussed live in our last weekly dev meeting, there are basically three places where this could be implemented:
- in warc2zim
- in zimit
- in zimfarm

Since the WARCs are produced by Browsertrix crawler / Zimit, it feels odd to implement this in warc2zim since it would mean the scraper is archiving its input files which are not modified at all by the scraper.

Implementing this in `zimit` is probably more straightforward: we do not take the risk to modify shared code base and it is quickly deployed on all workers

Implementing this in `zimfarm` is not expected to be very difficult (we already have an upload "function" for logs and ZIMs) and it will be valuable for other scrapers as well. But since it will be done in the worker, it will be more "complex" to deploy and might have a bigger impact in case of issue.

I'm now quite convinced that `zimfarm` is the right place to implement this.

WDYT?

benoit74:
## Rationale

- Fix #913 
- Fix what has been found to not be relevant around this change

Remark: 
- there are many changes, I suggest to review them commit by commit to gain a better understanding since commits are meant to provide meaning to the change made
- I did not included any upgrade of other frontend or backend dependencies, I prefer to do it in a separate PR (but I did not omit this is needed as well)

## Changes

- Always add `--keep` to zimit offliner command
  - Needed if we want to archive these artifacts (e.g. warc.gz files)
  - We do not mind since the container is deleted almost right after anyway
  - Exposing the parameter in Zimfarm adds no value for the end user and there is significant risk of mistake (very frustrating to restart a recipe with artifacts globs configured ... only to realize after multiple hours you forgot to also activate the `--keep` setting)
- Update QA tools and fix code accordingly
- Add artifacts functionality
  - Add a `artifacts_globs` configuration which is a list of strings
  - Modify API and UI to manipulate this new configuration
  - Modify task worker to check if `artifacts_globs` is configured after scraper has completed, and if so:
    - tar the files matching any of the globs in a single tar file
      - in-tar hierarchy match the original filesystem hierarchy to avoid to have to deal with conflicting file names
    - as files are added to the tar archive, delete them from the filesystem
    - start an upload container to upload the tar file
    - configuration is done through two new environment variables (needed to be set at task request, values are saved in task afterwards to be pushed to the worker): `ARTIFACTS_UPLOAD_URI` and `ARTIFACTS_EXPIRATION`
- Enhance checks at recipe creation to avoid inconsistencies between selected offliner and image name
- Enhance recipe config patching to delete keys which can be unset (`platform` and new `artifacts_globs`)
- Enhance secrets removal: always remove what looks like an S3 secret in what looks like a URL, in any key of the response, no matter where it is in the structure
- Fix tests around recipe patching:
  - in good patch tests, the assertions made were not checking the configuration content ... which is the main part modified by the patch
  - in bad patch tests fixture, many data was not needed and confusing, patch request was reduced to the strict minimum to make the test fail (i.e. changing a single dict value in test fixture make the test fail because data becomes valid)
  - add lots of comment to ensure we know what we are testing (so that it is easier to fix when an issue in encountered)
- Small cosmetic changes to recipe configuration UI:
  - Enhance descriptions of Docker image name and tag, and do not suggest "latest" as default value (this is not always a good idea, and not our usage anymore in none of our instances)
  - Add missing semicolons

## Screenshots

Recipe configuration (view)

image

Recipe configuration (edit)

image

New button in task debug screen to download artifacts:

image

Result tar content:

image

### DIFF
--- a/dispatcher/backend/src/common/constants.py
+++ b/dispatcher/backend/src/common/constants.py
@@ -33,6 +33,13 @@ try:
     LOGS_EXPIRATION = int(os.getenv("LOGS_EXPIRATION", "30"))
 except Exception:
     LOGS_EXPIRATION = 30
+ARTIFACTS_UPLOAD_URI = os.getenv(
+    "ARTIFACTS_UPLOAD_URI", "sftp://uploader@warehouse.farm.openzim.org:1522/artifacts"
+)
+try:
+    ARTIFACTS_EXPIRATION = int(os.getenv("ARTIFACTS_EXPIRATION", "30"))
+except Exception:
+    ARTIFACTS_EXPIRATION = 30
 
 # empty ZIMCHECK_OPTION means no zimcheck
 ZIMCHECK_OPTION = os.getenv("ZIMCHECK_OPTION", "")
--- a/dispatcher/backend/src/common/constants.py
+++ b/dispatcher/backend/src/common/constants.py
@@ -35,9 +35,7 @@ try:
     LOGS_EXPIRATION = int(os.getenv("LOGS_EXPIRATION", "30"))
 except Exception:
     LOGS_EXPIRATION = 30
-ARTIFACTS_UPLOAD_URI = os.getenv(
-    "ARTIFACTS_UPLOAD_URI", "sftp://uploader@warehouse.farm.openzim.org:1522/artifacts"
-)
+ARTIFACTS_UPLOAD_URI = os.getenv("ARTIFACTS_UPLOAD_URI", None)
 try:
     # artifact files expiration, 0 to disable expiration
     ARTIFACTS_EXPIRATION = int(os.getenv("ARTIFACTS_EXPIRATION", "30"))
--- a/dispatcher/backend/src/common/constants.py
+++ b/dispatcher/backend/src/common/constants.py
@@ -23,6 +23,7 @@ ZIM_UPLOAD_URI = os.getenv(
     "ZIM_UPLOAD_URI", "sftp://uploader@warehouse.farm.openzim.org:1522/zim"
 )
 try:
+    # ZIM files expiration, 0 to disable expiration
     ZIM_EXPIRATION = int(os.getenv("ZIM_EXPIRATION", "0"))
 except Exception:
     ZIM_EXPIRATION = 0
@@ -30,6 +31,7 @@ LOGS_UPLOAD_URI = os.getenv(
     "LOGS_UPLOAD_URI", "sftp://uploader@warehouse.farm.openzim.org:1522/logs"
 )
 try:
+    # log files expiration, 0 to disable expiration
     LOGS_EXPIRATION = int(os.getenv("LOGS_EXPIRATION", "30"))
 except Exception:
     LOGS_EXPIRATION = 30
@@ -37,6 +39,7 @@ ARTIFACTS_UPLOAD_URI = os.getenv(
     "ARTIFACTS_UPLOAD_URI", "sftp://uploader@warehouse.farm.openzim.org:1522/artifacts"
 )
 try:
+    # artifact files expiration, 0 to disable expiration
     ARTIFACTS_EXPIRATION = int(os.getenv("ARTIFACTS_EXPIRATION", "30"))
 except Exception:
     ARTIFACTS_EXPIRATION = 30
--- a/dispatcher/backend/src/routes/schedules/schedule.py
+++ b/dispatcher/backend/src/routes/schedules/schedule.py
@@ -291,12 +291,18 @@ class ScheduleRoute(BaseRoute):
 
         for key, value in update.items():
             if key in config_keys:
-                schedule.config[key] = value
+                if value is None:
+                    if key in schedule.config:
+                        del schedule.config[key]
+                else:
+                    schedule.config[key] = value
             elif key == "language":
                 schedule.language_code = value["code"]
                 schedule.language_name_en = value["name_en"]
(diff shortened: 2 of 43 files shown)


dispatcher/backend/src/routes/utils.py
```diff
@@ -87,19 +87,24 @@ def remove_secret_flag_from_command(
...
+            remove_s3_secrets(response[key])
+        else:
+            if not isinstance(response[key], str) or "://" not in response[key]:
+                continue
+            url = urlparse(response[key])
+            response[key] = url._replace(
+                query="&".join(
+                    [
+                        param
+                        for param in url.query.split("&")
```
rgaudin, line 105: 😳 use `parse_qs`

dispatcher/frontend-ui/src/components/ScheduleEditor.vue
```diff
@@ -131,7 +131,7 @@
         </b-form-group>
       </b-col>
       <b-col>
-         <b-form-group label="Image Tag" label-for="es_imagetag" description="Just the tag name. `latest` usually.">
+         <b-form-group label="Image Tag" label-for="es_imagetag" description="Input image name first to have proper values">
```
rgaudin:
```suggestion
         <b-form-group label="Image Tag" label-for="es_imagetag" description="Set image name first to get existing values">
```

dispatcher/backend/src/routes/schedules/schedule.py
```diff
@@ -306,6 +302,23 @@ def patch(
...
+    @dbsession
+    def patch(
+        self, schedule_name: str, token: AccessToken.Payload, session: so.Session
+    ):
+        """Update all properties of a schedule but _id and most_recent_task"""
+
+        schedule = dbm.Schedule.get(session, schedule_name, ScheduleNotFound)
+
+        try:
+            update = self._get_and_validate_patch(schedule=schedule, request=request)
```
rgaudin, line 316: Not a fan of this kind of things. I guess Codefactor complained about complexity?

rgaudin (approved):
Thank you ; please look at inline suggestions ; doesn't require re-review

dispatcher/backend/src/routes/schedules/schedule.py
benoit74: Yes ... I always feel cheating when doing this ... 
But once I've done the modification I had to admit that it makes the code probably easier to read and definitely easier to test (we could test individual functions in unit tests instead of doing this in integration tests)

dispatcher/backend/src/routes/utils.py
benoit74: 🤦🏻 this is already running in production, what a shame ; every time I read this code I had a bad feeling but always said in my mind "this has been reviewed, there was probably nothing best to do, do not loose your time on checking this again". I'll fix it right now!

workers/app/task/worker.py
```diff
@@ -461,6 +464,87 @@ def check_scraper_log_upload(self):
...
+        if not artifacts_globs:
+            logger.debug("No artifacts configured for upload")
+            return
+        else:
+            logger.debug(f"Archiving files matching {artifacts_globs}")
+
+        logger.debug("Creating a tar of requested artifacts")
+        filename = f"{self.task['_id']}_{self.task['config']['task_name']}.tar"
+        try:
+            files_to_tar = [
```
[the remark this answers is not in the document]
benoit74: Done

workers/app/task/worker.py
```diff
@@ -461,6 +464,87 @@ def check_scraper_log_upload(self):
...
+            if len(files_to_tar) == 0:
+                logger.debug("No files found to archive")
+                return
+
+            fpath = self.task_workdir / filename
+            with tarfile.open(fpath, "w") as tar:
+                for file in files_to_tar:
+                    tar.add(file, arcname=file.relative_to(self.task_workdir))
+                    try:
+                        file.unlink()
```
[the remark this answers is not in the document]
benoit74: I doubt it could be missing if we just added it to the archive, but OK

dispatcher/backend/src/common/constants.py
```diff
@@ -33,6 +33,13 @@
     LOGS_EXPIRATION = int(os.getenv("LOGS_EXPIRATION", "30"))
 except Exception:
     LOGS_EXPIRATION = 30
+ARTIFACTS_UPLOAD_URI = os.getenv(
+    "ARTIFACTS_UPLOAD_URI", "sftp://uploader@warehouse.farm.openzim.org:1522/artifacts"
+)
+try:
+    ARTIFACTS_EXPIRATION = int(os.getenv("ARTIFACTS_EXPIRATION", "30"))
```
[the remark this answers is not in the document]
rgaudin: I meant that setting this value is not completely up to the variable-setter since it **must** be higher than the configured bucket autodelete.

PR #918, opened 2024-02-22
Activate dependabot with advanced configuration

benoit74:
Currently dependabot is configured on this repo but:
- it creates PR only for critical version updates
- it does not group PRs (i.e. we have one PR per dependency to update)

The first point is not convenient because it means we still have to update all other dependencies manually, we do not have an overview of what upgrades we are missing, and we have many things to update (backend, frontend, taskmanager, workermanager, uploader, ...).

The second point is not convenient because it means we have a lot of PRs in // since we do not have sufficient bandwidth to check these PRs on a regular basis and we prefer to avoid updating systems on a fast pace when we are not focused on them.

This PR proposes to:
- activate dependabots for all kind of updates
- activate dependabots only for npm, pip and github-actions ecosystems
- to group PRs, or more precisely to have one PR per package-ecosystem and per dependency usage (production vs development)


rgaudin (approved):
Given we look at dependabot PRs weekly in the routine ; shouldn't we set a weekly period here?

LGTM otherwise

PR #927, opened 2024-02-26
Upgrade uploader

benoit74:
## Rationale

Maintenance

## Changes

- add a `ssh-host` dev container so that we can test uploader
  - also configured as ZIMS, LOGS and ARTIFACTS target for convenience during tests, no need to rely on external S3 
provider
  - including SSH material (public/private keys) for the server and for the client ; client key is in PEM format so that it can be used with the zimfarm as worker identity as well
- upgrade dependencies: Python 3.12 and Debian bookworm 
  - move to the slim Debian image, it seems sufficient according to tests
- pin Python dependency according to convention
- add `CONTRIBUTING.md` guide including instructions on how to test the uploader
- allow to pass a new `DOCKER_NETWORK` environment variable to the worker task manager, which is used to create child containers (e.g. uploaders)


rgaudin (approved):
Good! Can you add a comment somewhere to mention that real zimfarm uploads (except S3) targets a zimfarm receiver which authenticates differently

benoit74:
Hum, you are right ... 

Maybe the idea to add the `ssh-host` in dev stack is indeed a wrong one, I should have used the receiver directly. Here I'm not really testing what needs to be tested. Maybe I even broken the uploader without noticing. WDYT?

benoit74:
I ran tests locally, and can confirm it's in fact, new uploader is working fine with receiver and won't be deployed since we do not (yet) use latest image, so we can merge safely.

I will remove `ssh-host` dev container in next PR (tomorrow) and hence add a clear comment about which target we use for SSH and how authentication is handled.

PR #929, opened 2024-02-27
Upgrade receiver and workers

benoit74:
## Rationale

Maintenance

## Changes

- Upgrade receiver: 
  - Upgrade to Python 3.12 + Debian slim bookworm
  - Upgrade to Jailkit 2.23
- Upgrade task worker:
  - Upgrade to Python 3.12 + Debian slim bookworm
  - Upgrade and pin Python dependencies
- Upgrade worker manager:
  - Upgrade to Python 3.12 + Debian slim bookworm
  - Upgrade and pin Python dependencies
- Fix small bug in receiver `get_zimfarm_key.py` usage
- Adapt development docker-compose:
  - remove ssh-host: this should have been the receiver from the beginning
  - add receiver
    - including helper script to create all required warehouse paths
  - add commented worker manager and task worker
  - add instructions in `dev/README.md`
  - add `contrib` script to create a worker / start first requested task
  - fix backend configuration to upload all artifacts to receiver
  - fix uploader tests to use receiver instead of ssh-host
- trigger a rebuild of dnscache
- modify task worker to use latest images of dnscache, uploader
- remove `zmq` for worker dependencies + corresponding files

Open point: is it normal that we need to create all warehouse paths in the receiver manually ? This is not part of the procedure to create a new warehouse path, so I'm a bit surprised


workers/manager-Dockerfile
```diff
@@ -1,7 +1,11 @@
-FROM python:3.8-buster
+FROM python:3.12-slim-bookworm
 LABEL zimfarm=true
 LABEL org.opencontainers.image.source https://github.com/openzim/zimfarm
 
+# install libmagic for zimscraperlib
+RUN apt-get update -y \
+    && apt-get install -y --no-install-recommends libmagic1
```
rgaudin, line 7: Why would we want scraperlib in manager??

workers/task-requirements.txt
```diff
@@ -1,9 +1,9 @@
...
+requests==2.31.0
+docker==7.0.0
+psutil==5.9.8
+humanfriendly==10.0
+PyJWT==2.8.0
 kiwixstorage==0.6
-ujson>=5.1.0,<5.2
-zimscraperlib>=1.4.1,<1.5
+ujson==5.9.0
+zimscraperlib==3.3.0
```
rgaudin, line 8: Given how limited the scraperlib usage in task worker, might be worth looking into replacing it with pylibzim here.

dev/contrib/create_worker.sh
```diff
@@ -0,0 +1,62 @@
+#!/bin/bash
+
```
rgaudin, line 2: A comment describing what it does and why

dev/contrib/start_first_req_task.sh
```diff
@@ -0,0 +1,35 @@
+#!/bin/bash
+
```
rgaudin, line 2: A comment describing what it does and why

rgaudin (changes requested):
Good

workers/manager-Dockerfile
benoit74: Copy-paste issue, my bad.

workers/task-requirements.txt
benoit74: Done in https://github.com/openzim/zimfarm/pull/929/commits/ff103262721c44561e7204e4114a5ccd0a6c3184 

I had to duplicate some code from scraperlib but it was pretty limited, I do not expect this code to change a lot in the future and it is probably worth it to depend only on pylibzim instead of the whole scraperlib and all its dependencies.

It also helped to realize that this get_zim_info code was in fact broken. I did not realized it because `zimcheck` were not activated on the `dev` setup. It is activated now.

rgaudin (approved):
Thank you

PR #935, opened 2024-03-05
Zimit keep flag was set but not declared in Offliner definition

benoit74:
## Rationale

Since the introduction of a default `keep` flag defaulting "True" for Zimit, we can't update a recipe anymore.

image

## Changes

- declare `keep` flag in Zimit
- force the default `True` value only if not already set (after all, a developer might need to disable this for some reason)


benoit74:
Nota: tested locally with UI and worked well

rgaudin (approved):
LGTM ; there are indeed several ways to achieve this

PR #939, opened 2024-03-10
Fix #936 adjust css for std to show config value

### LINKED ISSUE #936: Config not displayed properly in not logged-in
See for example:
grafik

I see only each parameter name, but not the value.

I should be able to see the full configuration

poju3185:
## Rationale

The row `Scraper stdout` can sometimes be too long, pushing the value of `Config` to the far right side of the table, which makes user think they are invisible. So I add max width constrain to the stdout row.

## Changes

Add max width constrain to the stdout row

### DIFF
--- a/dispatcher/frontend-ui/src/views/TaskView.vue
+++ b/dispatcher/frontend-ui/src/views/TaskView.vue
@@ -306,6 +306,6 @@
   .stdout, .stderr {
     max-height: 9rem;
     overflow: scroll;
-    max-width: 120rem;
+    max-width: 40rem;
   }
 </style>
--- a/dispatcher/frontend-ui/src/views/TaskView.vue
+++ b/dispatcher/frontend-ui/src/views/TaskView.vue
@@ -306,5 +306,6 @@
   .stdout, .stderr {
     max-height: 9rem;
     overflow: scroll;
+    max-width: 120rem;
   }
 </style>


benoit74 (changes requested):
Where does the `120rem` comes from, is it in line with other components of the page, are you sure it works in all screen widths? Is the result responsive (responsiveness of Zimfarm UI is already limited but we should try to not make it worse)

Do you have some screenshot / video to share the result? 

Isn't there also an issue with the `command` which might be too long?

poju3185:
@benoit74 Thank you for your valuable feedback.

After reviewing the situation, I still believe that setting a maximum width is the best solution to address this issue. We're aware that the stdout can be quite lengthy, which is why overflow: scroll; was initially implemented. However, this approach didn't achieve the desired effect due to the response-table's tendency to expand rows as much as possible, preventing stdout from ever overflowing.

I acknowledge that setting the max-width to 120rem may not be the ideal solution, as it could force users to scroll horizontally on smaller screens to view the entire table. The primary goal of introducing a max-width was to prevent stdout from being the cause of table width expansion, given its potential length.

Considering the limitations, I propose adjusting the max-width to 40rem. This adjustment would allow users to see up to the first 40 characters at a glance, with the option to scroll horizontally for more information. This method ensures that even on smaller screens, the overall table width remains manageable.

I am, of course, open to exploring other solutions and would appreciate any suggestions you might have.

Please let me know your thoughts or if there are any alternative approaches you believe we should consider.


https://github.com/openzim/zimfarm/assets/81625625/8205924d-2254-43bd-9d14-827e6f7209b6

benoit74:
This is not working + we do not want to force screen width, it is already controlled elsewhere

Superseeded by #943

PR #941, opened 2024-03-11
Issue #936. Fix for long horizontal scrolling in table rows to show config values.

jairajmahadev:
## Rationale

Sometimes table rows can be too long and a lot of horizontal scrolling is required to see config values.

Issue: #939 

## Changes

Added CSS class ```truncate``` in ```TaskView.vue``` file. Class includes the following attributes:
```
white-space: nowrap;  /* disable wrap around text */
overflow-x: auto;     /* enable horizontal scroll */
max-width: 200px      /* limit width */
```


benoit74:
This is not working + we do not want to force screen width, it is already controlled elsewhere

Superseeded by #943

PR #942, opened 2024-03-12
Fix secret removal which was also removing carriage returns and not supporting multiple URLs in single string

### LINKED ISSUE #924: Missing carriage returns in log output
Not sure this is for zimfarm but in recent youzim runs (like [this one](https://farm.youzim.it/pipeline/4ad054a0-4015-42c1-be36-931aa2916ba3/debug)), I notice missing `\n` in the zimfarm UI log 

Screenshot 2024-02-26 at 09 56 59

benoit74:
## Rationale

Fix #924 (and enabler for #936)

## Changes

The logic to remove S3 secrets was not taking into account the fact that some dictionary keys (e.g. logs) might contain long strings with many S3 secrets to remove. It was also considering that the string is a URL, hence removing all carriage returns.

Logic has been fixed to take all this into account and properly continue to remove secrets.

Test cases have been supplemented with new test cases and assertion logic has be supplemented with a check of the whole response.

Password in network location is removed as well.

Secret replacement is now "--------" instead of "********" so that it is not replaced by url encoding / is part of the allowed chars in a URL/URI.

### DIFF
--- a/dispatcher/backend/src/routes/utils.py
+++ b/dispatcher/backend/src/routes/utils.py
@@ -97,14 +97,16 @@ def remove_s3_secrets(response: dict):
         else:
             if not isinstance(response[key], str) or "://" not in response[key]:
                 continue
-            url = urlparse(response[key])
-            response[key] = url._replace(
-                query="&".join(
-                    [
-                        f"{key}={value}"
-                        for key, values in parse_qs(url.query).items()
-                        if str(key).lower() not in ["keyid", "secretaccesskey"]
-                        for value in values
-                    ]
-                )
-            ).geturl()
+            for part in [part for part in response[key].split() if "://" in part]:
+                url = urlparse(part)
+                url = url._replace(
+                    query="&".join(
+                        [
+                            f"{key}={value}"
+                            for key, values in parse_qs(url.query).items()
+                            if str(key).lower() not in ["keyid", "secretaccesskey"]
+                            for value in values
+                        ]
+                    )
+                ).geturl()
+                response[key] = response[key].replace(part, url)
--- a/dispatcher/backend/src/routes/utils.py
+++ b/dispatcher/backend/src/routes/utils.py
@@ -1,6 +1,6 @@
 import logging
 from typing import Any, Dict, List
-from urllib.parse import parse_qs, urlparse
+from urllib.parse import parse_qs, urlencode, urlsplit, urlunsplit
 
 from common.constants import SECRET_REPLACEMENT
 from common.schemas.models import ScheduleConfigSchema
@@ -26,7 +26,7 @@ def has_dict_sub_key(data: Dict[str, Any], keys: List[str]):
 def remove_secrets_from_response(response: dict):
     """replaces or removes (in-place) all occurences of secrets"""
     remove_secret_flags(response)
-    remove_s3_secrets(response)
+    remove_url_secrets(response)
 
 
 def remove_secret_flags(response: dict):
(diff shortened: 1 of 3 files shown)


dispatcher/backend/src/routes/utils.py
```diff
@@ -97,14 +97,16 @@ def remove_s3_secrets(response: dict):
...
-                query="&".join(
-                    [
-                        f"{key}={value}"
-                        for key, values in parse_qs(url.query).items()
-                        if str(key).lower() not in ["keyid", "secretaccesskey"]
-                        for value in values
-                    ]
-                )
-            ).geturl()
+            for part in [part for part in response[key].split() if "://" in part]:
```
rgaudin, line 100: What's this supposed to be? Why splitting on space? Why not mame it `urlpart` to make it clearer? This will be hard to maintain

dispatcher/backend/src/routes/utils.py
```diff
@@ -97,14 +97,16 @@ def remove_s3_secrets(response: dict):
...
+                url = urlparse(part)
+                url = url._replace(
+                    query="&".join(
```
rgaudin, line 103: Again, why not using `urlencode(doseq=True)` to reconstruct?

dispatcher/backend/src/routes/utils.py
```diff
@@ -97,14 +97,16 @@ def remove_s3_secrets(response: dict):
...
+                        [
+                            f"{key}={value}"
+                            for key, values in parse_qs(url.query).items()
+                            if str(key).lower() not in ["keyid", "secretaccesskey"]
```
rgaudin, line 107: This does not only remove the secret but the params altogether right? This might mislead one into thinking the URL lacked the secret info. We should use our `SECRET_REPLACEMENT` as elsewhere

rgaudin (changes requested):

dispatcher/backend/src/routes/utils.py
benoit74: The `split` is splitting on space and carriage returns and tabs.

I nevertheless agree that naming is bad. I would suggest to use this:

```
for urlpart in [word for word in response[key].split() if "://" in word]:
```

WDYT?

dispatcher/backend/src/routes/utils.py
benoit74: Correct

dispatcher/backend/src/routes/utils.py
rgaudin: Yes that's better. What source data could trigger this?

dispatcher/backend/src/routes/utils.py
rgaudin: You deserialize with `parse_qs` so you should serialize back with `urlencode` as [the doc recommends](https://docs.python.org/3/library/urllib.parse.html#urllib.parse.parse_qs). I know the function name is misleading…

> Use the [urllib.parse.urlencode()](https://docs.python.org/3/library/urllib.parse.html#urllib.parse.urlencode) function (with the doseq parameter set to True) to convert such dictionaries into query strings.

dispatcher/backend/src/routes/utils.py
benoit74: This is typically useful for all logs, which might contain secrets if some scraper is not redacting them properly. We could even decide to not redact them anymore in the scraper so that Zimfarm admins have access to the plain text of the secret.

rgaudin (approved):

PR #952, opened 2024-04-18
Remove ability to create multiple ZIMs per TED recipe

benoit74:
## Rationale

We do not want to support anymore the creation of multiple ZIMs per TED recipe since this has proven to cause many metadata quality issues.

## Changes

- remove unneeded parameters
- use normal command instead of multi

NOTA : production DB has already been cleaned from any recipe using the multi mode + corresponding flags have been removed from recipes which might have had them at some point


rgaudin (approved):
LGTM

PR #967, opened 2024-05-14
StackExchange watcher: sync dumps whenever they are modified

### LINKED ISSUE #963: Watcher is not detecting updated files when SE is fixing issues
On April 2, SE pushed the March 2024 dumps to archive.org ; the watcher retrieved them and pushed them to S3.

SE quickly realized that the dumps were "corrupted/buggy".

They pushed an update on April 7.

See https://meta.stackexchange.com/a/398606 for details.

The watcher never retrieved these updated files and this caused issues on our side:
- the scraper was not working anymore
- I failed to realize this was just a corrupted dump issue and already fixed upstream

This already happened in the past.

The watcher should probably be smarter at detecting changes, so that we have higher chances to process the last available dump. The last available dump might still be corrupted, but at least it is the "best dump so far", and I won't say "this dump is 1 month old, it is obviously used by many users, it should be ok or issues would have been reported upstream".

WDYT?

benoit74:
## Rationale

Fix #963 

## Changes

- do not parse anymore the "Last-Modified" header of the 7z dump, store it as-is in S3 metadata
- consider the "Last-Modified" date of every individual 7z file instead of the one of "Site.xml"
  - probably a bit slower since we need to do a `HEAD` request on every file, but who cares, this is supposed to run only once a day
- S3 metadata is renamed from `version` to `lastmodified` (all lower-case, no dash because upper-case with dash wasn't working properly)
- Logs have been adapted and we now log systematically the new "Last-Modified" on a file-by-file basis
- Add a docker-compose dev container for the watcher + instructions

## Tests

This has been tested "locally" with the docker-compose container + a newly created org-kiwix-dev-stackexchange S3 bucket in Wasabi:
- run previous code from `main` branch
  - file is uploaded since bucket was empty
- run again previous code from `main` branch
  - nothing to do
- run new code from `watcher_last_modified`
  - previous file is removed, new file is uploaded since metadata has changed
- run new code from `watcher_last_modified`
  - nothing to do
- delete file from S3 bucket and run new code from `watcher_last_modified`
  - file is uploaded since bucket was empty
- tweak last-modified value in source code and run new code from `watcher_last_modified`
  - file is uploaded since last-modified value has changed
- run real new code from `watcher_last_modified`
  - file is uploaded since last-modified value has changed
- run real new code from `watcher_last_modified`
  - nothing to do

### DIFF
--- a/dev/README.md
+++ b/dev/README.md
@@ -41,6 +41,10 @@ This container is the main worker container, responsible to start tasks. It is c
 
 This container is a sample task executor. It is commented by default.
 
+### watcher
+
+This container is a StackExchange dumps watcher.
+
 ## Instructions
 
 First start the Docker-Compose stack:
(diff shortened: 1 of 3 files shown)


rgaudin (approved):
Looks great

PR #984, opened 2024-06-18
Add new CLI arguments for Zimit/Warc2zim 2.0.2

benoit74:
## Rationale

Add new arguments in Zimit offliner definition, following the release of Zimit/Warc2zim 2.0.2

Nota: this will also be deployed on farm.zimit.kiwix.org where we still do not use 2.x, and on recipes of farm.openzim.org which are still in 1.x ; this is assumed to be mostly harmless since it has no impact if options are not specified, and none of them are required.


benoit74:
Sorry, I just spotted a silly mistake

benoit74:
This is ready for review again

rgaudin (approved):
LGTM

PR #999, opened 2024-07-22
When scheduling, also consider task age to sort requested tasks

### LINKED ISSUE #946: Task scheduling should also take into account requested task "age"
For now, when deciding which is the next task a worker should start, the scheduling algorithm order requested task first by priority, then by expected duration.

It would be great to also take into account the requested task "age" (i.e. duration since requested date), so that tasks requested a long time ago are started first. This is especially important for the situation (like youzim.it) where all tasks have the same priority and expected/empty duration.

I suggest to simply add the age as a third sorting criteria.

Another alternative would be to create a weighted approach between priority, expected duration and age, but it seems a daunting task (weigths would be difficult to decide) for a limited added value.

benoit74:
## Rationale

Fix #946

## Changes
- Also consider task age (based on `updated_at` which is a "proxy") when sorting requested tasks, so that older tasks are ran first

### DIFF
--- a/dispatcher/backend/src/utils/scheduling.py
+++ b/dispatcher/backend/src/utils/scheduling.py
@@ -301,7 +301,9 @@ def get_reqs_doable_by(session: so.Session, worker: dbm.Worker) -> List[Dict[str
     """cursor of RequestedTasks() doable by a worker using all its resources
 
     - sorted by priority
-    - sorted by duration (longest first)"""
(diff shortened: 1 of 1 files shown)


benoit74:
I will merge this right away, it is too important for zimit, I completely forgot about this bug since weeks ...

@rgaudin feedback is still welcomed of course.

PR #1004, opened 2024-07-29
Relaxed nautilus schema

rgaudin:
As for zimit.kiwix.org, we want the ZIM as a service to produce ZIM file via the
api.farm.zimit.kiwix.org backend without the zim-file constraint of the period.

Duplicated the zimit tests to prevent regression but it serves no other purpose.


dispatcher/backend/src/tests/integration/routes/schedules/test_nautilus.py
```diff
@@ -0,0 +1,109 @@
+from collections import namedtuple
+from typing import List
+
+import pytest
+
+from common import constants
+
+
+def update_dict(dict: dict, key_path: str, new_value: any):
```
benoit74, line 9: Should we share this across tests? I assume this is not specific to Nautilus, but if it is stored here then we will not reuse it for sure.

benoit74 (approved):
LGTM, few minor comments, I let you decide what to do with them then merge.

PR #1014, opened 2024-09-12
Rework/fixes for zimit-frontend

### LINKED ISSUE #1013: Add query parameter to hide schedule / task secret flags on-demand
For zimit-frontend, we query the /tasks/xxx endpoint to retrieve details about the task.

While the zimit-frontend internal API does it best to hide unnecessary details (including potential secrets e.g. notifications webhooks), it is impossible to ensure that flags with secret values are not exposed.

Currently zimit scraper does not have any secret flag, but this might change at some point in the future.

zimit-frontend needs to call the /tasks/xxx endpoint with authentification so that it retrieves details about the notification in order to know if an email has been configured by the user. All flags are hence exposed in clear text. 

I recommend to add an optional query parameter `hide_secrets` so that zimit-frontend can explicitely require to hide the secrets.

benoit74:
- Fix #1013 
- Also hide secret flags on `GET /requested-tasks/{id}` endpoint (was forgotten previously, probably because it is not used in the UI)
- Add `notification` property on `GET /requested-tasks/{id}` endpoint

Remark: 
- `?hide_secrets=` tested locally, works well, complex to test in automated end-to-end

### DIFF
--- a/dispatcher/backend/docs/openapi_v1.yaml
+++ b/dispatcher/backend/docs/openapi_v1.yaml
@@ -261,7 +261,9 @@ paths:
       - public
       summary: dump schedules
       operationId: dumpSchedules
-      description: Dump all Schedules (without `most_recent_task`) in a single request. (Notfication is removed and Secrets in config are replaced by ****** for unauthenticated requests and authenticated requests from users without schedules.update permission)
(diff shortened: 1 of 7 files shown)


benoit74:
Edit: update first message with some details

benoit74:
Edit: added `notification` property on requested-tasks endpoint, it is needed / used by zimit-frontend

PR #1020, opened 2024-09-24
Add Devdocs offliner, category and warehouse path

benoit74:
## Rationale

Add new devdocs offliner, including new category (since we do not yet know which category list we want to use) and new warehouse path.

Also add new platform since we do not want to run more than one task per worker (but without global limit on the farm). See setting for our prod in https://github.com/kiwix/operations/pull/263

Nota: while we have not yet released a 0.1, this is coming soon and this code should be ok to deploy since anyway we already have a `dev` Docker image at https://github.com/openzim/devdocs/pkgs/container/devdocs

### DIFF
--- a/dispatcher/backend/src/common/schemas/offliners/devdocs.py
+++ b/dispatcher/backend/src/common/schemas/offliners/devdocs.py
@@ -1,0 +1,162 @@
+from marshmallow import fields
+
+from common.schemas import SerializableSchema, String
+from common.schemas.fields import (
+    validate_output,
+    validate_zim_description,
+    validate_zim_longdescription,
+)
+
+
+class DevDocsFlagsSchema(SerializableSchema):
+    class Meta:
+        ordered = True
+
+    all_flag = fields.Boolean(
+        truthy=[True],
+        falsy=[False],
+        metadata={
+            "label": "All",
+            "description": "Fetch all Devdocs resources, and produce one ZIM "
+            "per resource.",
+        },
+        data_key="all",
+    )
+
+    slug = String(  # should be ListOfString but not yet supported by Zimfarm
+        metadata={
+            "label": "Slug",
+            "description": "Fetch the provided Devdocs resource. "
+            "Slugs are the first path entry in the Devdocs URL. "
+            "For example, the slug for: `https://devdocs.io/gcc~12/` is `gcc~12`.",
+        },
+    )
+
+    first = fields.Integer(
+        metadata={
+            "label": "Number of first items",
+            "description": "Fetch only the first N items per slug as shown "
+            "in the DevDocs UI. Do not set to fetch all items.",
+        },
+    )
+
+    skip_slug_regex = String(
+        metadata={
+            "label": "Skip slugs regex",
+            "description": "Skips slugs matching the given regular expression."
+            "Do not set to fetch all slugs",
+        },
+        data_key="skip-slug-regex",
+    )
+
+    file_name_format = String(
+        metadata={
+            "label": "ZIM filename",
+            "description": "ZIM filename. Do not input trailing `.zim`, it "
+            "will be automatically added. You can use placeholders, see "
+            "https://github.com/openzim/devdocs/blob/main/README.md. Defaults "
+            "to devdocs.io_en_{clean_slug}_{period}",
+        },
+        data_key="file-name-format",
+    )
+
+    name_format = String(
+        metadata={
+            "label": "ZIM name",
+            "description": "ZIM name. You can use placeholders, see "
+            "https://github.com/openzim/devdocs/blob/main/README.md. Defaults "
+            "to devdocs.io_en_{clean_slug}",
+        },
+        data_key="name-format",
+    )
+
+    title_format = String(
+        metadata={
+            "label": "ZIM title",
+            "description": "ZIM title. You can use placeholders, see "
+            "https://github.com/openzim/devdocs/blob/main/README.md. Defaults "
+            "to `{full_name} Docs`",
+        },
+        data_key="title-format",
+    )
+
+    description_format = String(
+        metadata={
+            "label": "ZIM description",
+            "description": "ZIM description. You can use placeholders, see "
+            "https://github.com/openzim/devdocs/blob/main/README.md. Defaults "
+            "to `{full_name} docs by DevDocs`",
+        },
+        data_key="description-format",
+        validate=validate_zim_description,
+    )
+
+    long_description_format = String(
+        metadata={
+            "label": "ZIM long description",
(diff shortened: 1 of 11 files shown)


dispatcher/backend/src/common/schemas/offliners/devdocs.py
```diff
@@ -0,0 +1,162 @@
...
+    devdocs_frontend_url = String(
+        metadata={
+            "label": "DevDocs frontend URL",
+            "description": "Scheme and hostname for the devdocs frontend."
+            "Defaults to https://devdocs.io",
+        },
+        data_key="devdocs-frontend-url",
+    )
+
+    devdocs_documents_url = String(
```
rgaudin, line 155: Should we expose this dev-only option?

dispatcher/backend/src/common/schemas/offliners/devdocs.py
```diff
@@ -0,0 +1,162 @@
...
+        validate=validate_output,
+    )
+
+    debug = fields.Boolean(
+        truthy=[True],
+        falsy=[False],
+        metadata={"label": "Debug", "description": "Enable verbose output"},
+    )
+
+    devdocs_frontend_url = String(
```
rgaudin, line 146: Should we expose this dev-only option?

dispatcher/backend/src/common/schemas/offliners/devdocs.py
```diff
@@ -0,0 +1,162 @@
...
+            "label": "ZIM filename",
+            "description": "ZIM filename. Do not input trailing `.zim`, it "
+            "will be automatically added. You can use placeholders, see "
+            "https://github.com/openzim/devdocs/blob/main/README.md. Defaults "
+            "to devdocs.io_en_{clean_slug}_{period}",
+        },
+        data_key="file-name-format",
+    )
+
+    name_format = String(
```
rgaudin, line 63: I thought we wanted to get away from those auto-generated metadata?

dispatcher/backend/src/common/schemas/offliners/devdocs.py
```diff
@@ -0,0 +1,162 @@
...
+    validate_zim_description,
+    validate_zim_longdescription,
+)
+
+
+class DevDocsFlagsSchema(SerializableSchema):
+    class Meta:
+        ordered = True
+
+    all_flag = fields.Boolean(
```
rgaudin, line 15: Is the `skip_slig_regex` applied when using this? That's unclear

dispatcher/backend/src/common/schemas/offliners/devdocs.py
```diff
@@ -0,0 +1,162 @@
...
+        falsy=[False],
+        metadata={
+            "label": "All",
+            "description": "Fetch all Devdocs resources, and produce one ZIM "
+            "per resource.",
+        },
+        data_key="all",
+    )
+
+    slug = String(  # should be ListOfString but not yet supported by Zimfarm
```
rgaudin, line 26: Cant the scrapper support a split syntax as we generally do?

dispatcher/backend/src/common/schemas/offliners/devdocs.py
```diff
@@ -0,0 +1,162 @@
...
+        metadata={
+            "label": "Slug",
+            "description": "Fetch the provided Devdocs resource. "
+            "Slugs are the first path entry in the Devdocs URL. "
+            "For example, the slug for: `https://devdocs.io/gcc~12/` is `gcc~12`.",
+        },
+    )
+
+    first = fields.Integer(
```
rgaudin, line 35: I see no such option on the website. Or does `DevDocs UI` here means the ZIM UI? Feels like a dev option that shouldn't be on ZF

rgaudin (changes requested):
Category and warehouse paths require a decision. I haven't heard we're creating either for devdocs and so until then this should be `other` for both I believe.

Excuse my questions if some doesn't make sense ; I haven't used the scraper yet.

dispatcher/backend/src/common/schemas/offliners/devdocs.py
benoit74: I think we want to avoid auto-generated metadata which might contain only garbage, be too long, ...

Here we know (and have tested) that these metadata are compliant with all existing slugs on devdocs.io, and it allows to ensure some homogeneity (avoiding typos typically) so I think they make much more sense than description automatically grabbed from Youtube channel typically.

That being said, I can hide (not document) them on the Zimfarm if it seems better.

dispatcher/backend/src/common/schemas/offliners/devdocs.py
benoit74: I think it might avoid to have to release the scraper should we have to update this. But it is not like this is going to change, I agree.

dispatcher/backend/src/common/schemas/offliners/devdocs.py
benoit74: Good point

dispatcher/backend/src/common/schemas/offliners/devdocs.py
benoit74: Yes it could, let's open an issue for that

dispatcher/backend/src/common/schemas/offliners/devdocs.py
benoit74: I agree I see no usage for this in production.

kelson42:
> Category and warehouse paths require a decision. I haven't heard we're creating either for devdocs and so until then this should be `other` for both I believe.

Supportive of a dedicated Category and storage path.

dispatcher/backend/src/common/schemas/offliners/devdocs.py
benoit74: Removed

dispatcher/backend/src/common/schemas/offliners/devdocs.py
benoit74: Same remark, removed

dispatcher/backend/src/common/schemas/offliners/devdocs.py
benoit74: Removed

benoit74:
@rgaudin do not hesitate to ask for live call to settle on open discussions

rgaudin (approved):
I think everything's settled but you did not re-request… approved anyway

benoit74:
Of course I did ^^
image

rgaudin:
Do you have a timestamped attorney-certified copy of that? 🙃

PR #1024, opened 2024-10-08
Enforce that ZIM title must not be longer than 30 chars on our ZIMs

### LINKED ISSUE #777: Clarify that Title should be max 30 characters long
Creating a zim file whose title is longer than 30 characters will cause the job to crash. This should either be explicitely indicated in the `Title` field (with a small counter to indicate where one is) or if there is no particular reason for this limit then extend to something along the lines of 60 characters.

benoit74:
## Rationale

Fix #777 

## Changes

- add new validation method for ZIM titles (no longer than 30 chars)
- setup this validation method on title metadata on all offliners


rgaudin (approved):
LGTM

PR #1032, opened 2024-10-14
Add a platform for shamela.ws zimit recipes

### LINKED ISSUE #1023: Create a new platform for shamela.ws
For https://github.com/openzim/zim-requests/issues/1172, we are going to have 40 recipes, one per category on shamela.ws to have pratical recipe duration and practical ZIM sizes.

However, there is only one upstream server, so we need to ensure that we do not run more than one task (out of these 40 recipes) per worker, or even probably only one task on the whole platform to be fair with their servers.

Shall we create a new `shamela.ws` platform we would assign manually to the corresponding recipes?

benoit74:
## Rationale

Fix #1023

### DIFF
--- a/dispatcher/backend/src/common/enum.py
+++ b/dispatcher/backend/src/common/enum.py
@@ -274,6 +274,7 @@ class Platform:
     ifixit = "ifixit"
     ted = "ted"
     devdocs = "devdocs"
+    shamela = "shamela"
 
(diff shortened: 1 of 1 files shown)


rgaudin:
Should we start adding comments in code about how we expect configuration? Or why we add the platform?

Issue #723, opened 2022-11-09
Mark tasks as failed if no ZIM file was created


rgaudin:
Sort of a complement of #684 ; if the scraper completes with `0` yet we have not seen any ZIM files, something must be wrong and the task should be marked as failed.
In addition, a log message should be produced with the (recursive ? @kelson42 ?) content of `/output` so one can check whether something was created (temp ZIM not finished, incorrect extension).

benoit74:
Removed from zimit2 project unfortunately

benoit74:
The fact that we need to mark task as failed when no ZIMs are found has been tackled in https://github.com/openzim/zimfarm/pull/1045

Still need to log the content of output folder.

benoit74:
Solved in `zimfarm-upgrade` branch

PR #1045, opened 2024-11-08
Mark task with no files created as failed

benoit74:
## Rationale

- would allow to have catched https://github.com/openzim/zimit-frontend/issues/90
- make more sense anyway

## Changes

- when a task has no file, mark it as failed

Nota: not tested, hard to reproduce


rgaudin:
Fixing #723 ?

benoit74:
> Fixing https://github.com/openzim/zimfarm/issues/723 ?

Good point ! Yes, but only partially, we do not log the content of output folder yet.

PR #1050, opened 2024-11-18
Log exception when recording IP changes fails

benoit74:
## Rationale

- When recording the IP change fails, we have no clue in the logs about what happened
- I also changed slightly the message so that it is clearer that the problem is not in our DB (which has been updated) but in pushing the changes to Wasabi


benoit74:
@rgaudin I did not asked for formal review and merged right away because it is minor and I needed the code in production, but feedbacks are welcomed

PR #1051, opened 2024-11-19
Add new flags to mindtouch scraper

benoit74:
## Rationale

Mindtouch scraper needs new flags


benoit74:
@rgaudin feedback welcomed, but I will merge right away to save back-and-forth, it is so straigthforward

PR #1062, opened 2024-12-16
Create a tool to automatically create and maintain Zimfarm recipes

benoit74:
## Rationale

For ted, devdocs, and others, we have a bunch of recipes which:
- all needs to share the same settings
- have to be maintained:
  - create new recipes as new content appears online
  - delete obsolete recipes
  - update settings of all recipes at once

This tool is an attempt to ease this task for developers. It is **NOT** intended to be used in an automated manner or by non-developers (yet, at least).

This tool is already in use for TED, and mostly ready to create devdocs recipes (I'm just waiting for confirmation on how many ZIMs we want, with what inside).

This tool is meant to easily:
- add new sets (currently only ted and devdocs, but new can easily be added like stackoverflow, ...)
- customize what needs to be at recipe level, either with yaml configuration or any Python code
- pass secret values from CLI (avoiding to store these in github 🤣)


rgaudin (approved):
Great ; if it takes only a few minutes, could you add in the gist of how to create a new module (or maintain one). What are the functions to implement and what's the expected output for instance.

benoit74:
Good point, will do

benoit74:
Done

PR #1072, opened 2025-02-17
Add support for requested events so that notifications will be sent

benoit74:
## Rationale

Prerequisite to solve https://github.com/openzim/zimit-frontend/issues/74

Currently, "requested" events are passed to the `task_event_handler` just like other events

https://github.com/openzim/zimfarm/blob/360d1b99b6a8bdf341ec60853ab2f75589405564/dispatcher/backend/src/routes/requested_tasks/requested_task.py#L187-L189

Unfortunately, `requested` is not inside the map of events

https://github.com/openzim/zimfarm/blob/360d1b99b6a8bdf341ec60853ab2f75589405564/dispatcher/backend/src/common/utils.py#L27-L43

Consequence is that it is handled by special "fallback code" which does not sends notifications should they be registered

This is confirmed by the presence of these messages in logs:

```
1739788569127	INFO:common.utils:Other event: requested
1739800812060	INFO:common.utils:Other event: requested
```

This PR fix this by adding `requested` to supported events. Event handler is simply a log line (could help debugging) since `requested` event is already stored in DB at `RequestedTask` creation (which is ok/superior to avoid two SQL queries - INSERT then UPDATE - instead of one).

## Changes

- Add support for `requested` events 
  - special handling in notifications:
    - data comes from a distinct table
    - data sent in the notification body is slightly different (it is a requested task and not a task)
- Fix small typing error in `RequestedTaskFullSchema`

### DIFF
--- a/dispatcher/backend/src/common/notifications.py
+++ b/dispatcher/backend/src/common/notifications.py
@@ -25,8 +25,8 @@ from common.constants import (
 from common.emailing import send_email_via_mailgun
 from common.enum import TaskStatus
 from common.schemas.models import EventNotificationSchema, ScheduleNotificationSchema
-from common.schemas.orms import ScheduleAwareTaskFullSchema
-from errors.http import TaskNotFound
+from common.schemas.orms import RequestedTaskFullSchema, ScheduleAwareTaskFullSchema
+from errors.http import RequestedTaskNotFound, TaskNotFound
 
 logger = logging.getLogger(__name__)
 jinja_env = Environment(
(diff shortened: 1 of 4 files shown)


benoit74:
Nota: tested locally with full zimfarm + zimit-frontend stack and it works well, email is built properly (besides a small bug in zimit-frontend, but this is another story), just not sent due to missing Mailgun configuration

rgaudin:
It was not implemented on purpose so that the event payload was stable across all events (ie. TaskDetails vs RequestedTaskDetail).

rgaudin (approved):
Albeit the potential/marginal API change that was not mentioned in the description this is OK.

benoit74:
> It was not implemented on purpose so that the event payload was stable across all events (ie. TaskDetails vs RequestedTaskDetail).

OK, but then sending a mail in Zimit when task is submitted never worked ... anyway, not a big deal.

> Albeit the potential/marginal API change that was not mentioned in the description this is OK.

Fixed, thank you.

PR #1076, opened 2025-02-20
Add monitoring endpoint for oldest task in a given status

benoit74:
## Rationale

Add an endpoint which can be called from UptimeRobot to e.g. monitor oldest running task

## Changes

- add a generic `/status/{monitor_name}` endpoint
- implement a first monitor `oldest_task_older_than` which takes two parameters (as query string):
  - `status`: a task status which has an associated timestamp, e.g. `scraper_started`, `scraper_completed`, `cancel_requested`
  - `threshold_secs`: a threshold which cause the monitor to fail if oldest task is older than this threshold, in seconds
  - returns a textual answer with HTTP code 200 and details if monitor succeeded to run:
    - e.g. `oldest_task for scraper_started and 500s: OK` or `oldest_task for scraper_started and 5s: KO`
  - returns HTTP code 400 on bad requests
- update (manually 🤢) OpenAPI spec
- add automated tests
  - slightly modify test fixtures so that tasks have timestamp a bit in the past (we didn't cared about their values so far)

The `: KO` or `: OK` strings can be searched for in UptimeRobot `KW` probe. Always returning `200` even in case of "problem" is deemed more appropriate to respect HTTP paradigms and allows to more easily differentiate a buggy API from a down monitor.

### DIFF
--- a/dispatcher/backend/src/routes/status/__init__.py
+++ b/dispatcher/backend/src/routes/status/__init__.py
@@ -1,0 +1,10 @@
+from routes import API_PATH
+from routes.base import BaseBlueprint
+from routes.status.status import StatusMonitorRoute  # , StatusFooRoute
+
+
+class Blueprint(BaseBlueprint):
+    def __init__(self):
+        super().__init__("status", __name__, url_prefix=f"{API_PATH}/status")
+
+        self.register_route(StatusMonitorRoute())
--- a/dispatcher/backend/src/routes/status/status.py
+++ b/dispatcher/backend/src/routes/status/status.py
@@ -1,0 +1,66 @@
+import datetime
+import logging
+
+import sqlalchemy as sa
+import sqlalchemy.orm as so
+from flask import request
+
+import db.models as dbm
+from db import dbsession
+from routes.base import BaseRoute
+from routes.errors import BadRequest
+
+logger = logging.getLogger(__name__)
+
+
+class StatusMonitorRoute(BaseRoute):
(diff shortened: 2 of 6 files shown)


dispatcher/backend/src/routes/status/__init__.py
```diff
@@ -0,0 +1,11 @@
...
+from routes.base import BaseBlueprint
+from routes.status.status import StatusMonitorRoute  # , StatusFooRoute
+
+
+class Blueprint(BaseBlueprint):
+    def __init__(self):
+        super().__init__("status", __name__, url_prefix=f"{API_PATH}/status")
+
+        self.register_route(StatusMonitorRoute())
+        # self.register_route(StatusFooRoute())
```
rgaudin, line 11: devel leftover?

dispatcher/backend/src/routes/status/status.py
```diff
@@ -0,0 +1,78 @@
...
+        if monitor_name not in handlers:
+            raise BadRequest(
+                f"Monitor '{monitor_name}' is not supported. Supported "
+                f"monitors: {','.join(handlers.keys())}"
+            )
+
+        return handlers[monitor_name](session)
+
+
+# class StatusFooRoute(BaseRoute):
```
rgaudin, line 69: leftover again

rgaudin (approved):
Good ; cleanup the commented code first though

benoit74:
🤦🏻‍♂️thank you

PR #1077, opened 2025-02-21
Add requested task rank

benoit74:
## Rationale

This is a prerequisite for https://github.com/openzim/zimit-frontend/issues/32

## Changes

Add a "rank" property to the requested task endpoint `GET /requested_tasks/{task_id}`. Computation of the rank is deemed to be "cheap enough" since we only need the list of all requested task IDs (not the whole objects) and we do not expect to have thousands of requested tasks in the pipe.

Rank starts at 0 (meaning next task to start).

Also fix a mistake about bad exception being used.

### DIFF
--- a/dispatcher/backend/src/routes/requested_tasks/requested_task.py
+++ b/dispatcher/backend/src/routes/requested_tasks/requested_task.py
@@ -300,6 +300,18 @@ class RequestedTaskRoute(BaseRoute):
         )
         resp = RequestedTaskFullSchema().dump(requested_task)
 
+        # also fetch all requested tasks IDs to compute estimated task rank ; this is
+        # only an indicator for zimit.kiwix.org where duration is unknown because
+        # schedule is created on-demand and all tasks have access to same worker(s) ;
+        # sorting by priority and updated_at won't give a good indicator in other cases
+        stmt = sa.select(
+            dbm.RequestedTask.id,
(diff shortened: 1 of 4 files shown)


dispatcher/backend/src/routes/requested_tasks/requested_task.py
```diff
@@ -290,9 +295,23 @@ def get(
...
-        requested_task = dbm.RequestedTask.get(session, requested_task_id, TaskNotFound)
+        requested_task = dbm.RequestedTask.get(
+            session, requested_task_id, RequestedTaskNotFound
+        )
         resp = RequestedTaskFullSchema().dump(requested_task)
 
+        # also fetch all requested tasks IDs to compute estimated task rank ; this is
+        # only an estimation since logic to start tasks is more complex than simple
+        # sorting but it gives a good indicator of when the requested task might
+        # be expected to start (soon or in a long time) especially for zimit.kiwix.org
```
rgaudin, line 306: This is misleading. It can only be a fair estimation on zimit ; not _especially for zimit_.
What makes it usable there is:
- no existing `duration` because schedule is created on-demand
- identical resource provisioning for all tasks
I think source-code comment should be precise about this

rgaudin (approved):
Good but please reword the comment

PR #1086, opened 2025-03-28
convert everywhere from KB, KiB -> KiB, MiB with 3-digit precision

### LINKED ISSUE #989: Ensure we report MiB/GiB/... values everywhere
See https://github.com/kiwix/libkiwix/issues/1095

We need to:
- display a MiB/GiB value with only 3 significant digits, rounded to the nearest
- ensure the unit is written as MiB/GiB

E.g. if a file size is `12341234` bytes, it must be reported as `11.8 MiB` (12341234 / 1024 / 1024 = 11,7695178986)

Mightymanh:
## Rationale

Ensure we report MiB/GiB/... values everywhere instead of MB/GB/... in frontend for consistency

Resolves #989 

## Changes

- upgrade npm package filesize from 6.0.1 -> 6.4.0 to get filesize() with more functionality such as precision (we need 3-digits precision)
- change filesize2 in Constant.js to accept more parameter that display in desired format (MiB/GiB/..., 3-digits precision)
- in all files that import filesize() from npm package filesize, change them to import filesize() from Constant.js for consistency

### DIFF
--- a/dispatcher/frontend-ui/package.json
+++ b/dispatcher/frontend-ui/package.json
@@ -16,7 +16,7 @@
     "bootstrap-vue": "^2.1.0",
     "core-js": "^3.3.2",
     "deep-diff": "^1.0.2",
-    "filesize": "^6.0.1",
+    "filesize": "^10.1.6",
     "follow-redirects": "1.14.9",
     "glob-parent": "^6.0.1",
     "jsonwebtoken": "^8.5.1",
--- a/dispatcher/frontend-ui/src/components/ResourceBadge.vue
+++ b/dispatcher/frontend-ui/src/components/ResourceBadge.vue
@@ -5,7 +5,7 @@
 </template>
 
 <script type="text/javascript">
-  import filesize from 'filesize';
+  import Constants from '../constants.js'
 
   export default {
     name: 'ResourceBadge',
@@ -18,7 +18,7 @@
       displayed_value() {
         if (this.human_value)
           return this.human_value;
-        return (this.kind == 'cpu') ? this.value : filesize(this.value); },
+        return (this.kind == 'cpu') ? this.value : Constants.formattedBytesSize(this.value); },
       icon() { return {cpu: 'microchip', memory: 'memory', disk: 'hdd', shm: 'compact-disc'}[this.kind]; },
       tooltip_text() {
         if (!this.tooltip)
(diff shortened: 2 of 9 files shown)


dispatcher/frontend-ui/package.json
```diff
@@ -16,7 +16,7 @@
     "bootstrap-vue": "^2.1.0",
     "core-js": "^3.3.2",
     "deep-diff": "^1.0.2",
-    "filesize": "^6.0.1",
+    "filesize": "^6.4.0",
```
benoit74, line 19: Why not upgraded to latest version ? There are major changes but the API we are using seems pretty straightforward

dispatcher/frontend-ui/src/components/ResourceBadge.vue
```diff
@@ -15,10 +15,10 @@
       human_value: String,  // human repr of value (instead of raw one)
     },
     computed: {
-      displayed_value() {
+      displayed_value() {
```
benoit74, line 18: Why whitespace?

benoit74 (changes requested):
See inline comments + points below:
- you are using `Constants.filesize` everywhere and do not use `filesize2` so I'm pretty confused ; plus `filesize2` name seems pretty ugly (I now it was here before you came, but it could be a good opportunity to rename it)
- does this fix what is the most important thing to fix: file sizes?
- please share screenshots of resulting display (I'm concerned it does not get broken)

Mightymanh:
Also do you have good command for updating a package? Last time I use `npm update` and it updates to 6.4. In order to update to latest version I `npm uninstall` the package and reinstall it again to fetch latest version.

A lot of our dependencies are outdated:
image

benoit74 (changes requested):
- please remove package-lock.json, we use `yarn` on this project
- `yarn upgrade-interactive --latest` is the de-facto way to upgrade dependencies on this project
- please rename both methods and export to `formattedBytesSize`, for consistency + clarity (we use it on memory as well, it is odd to name it filesize ... even if under the hood we use the filesize package, but this can be hidden inside Constants, and changed at any future point in time)
- regarding the fact that everything is old in this project ... yes, we are aware of this, I will eventually work on this in few weeks/months

Mightymanh:
That was fun as I learn about squash and force push. Git has more utilities I don't know that are worth to learn

benoit74:
> That was fun as I learn about squash and force push. Git has more utilities I don't know that are worth to learn

fixup, squash, amend, reflog, force push, there are many very useful stuff in git. Good reference to learn most of it is probably https://learngitbranching.js.org/

PR #1093, opened 2025-04-10
Improve Navbar Style For Smaller Devices

### LINKED ISSUE #1091: UI: Improve Navbar Style For Mobile devices
Currently, in the website's mobile navbar, all items are left-aligned, and the "Support Us" and "Sign In" buttons have inconsistent widths. We can improve this by adjusting the styles. 

Question is this good idea? -> on larger screens, the "Support Us" button currently appears alongside the other links on the left side. We can improve the layout by moving it next to the "Sign In" button on the right side, and making sure both buttons have the same width.

Here's a screenshot of what it currently looks like, and an example of the expected layout.

<table>
<tr>
<td>
currently
</td>
<td>
expected
</td>
</tr>
<tr>
<td>

</td>
<td>

</td>
</tr>
</table>

Jainex17:
## Rationale
Currently, in the website's mobile navbar, all items are left-aligned, and the "Support Us" and "Sign In" buttons have inconsistent widths. We can improve this by adjusting the styles.

Fixes #1091 

## Changes
In the mobile view, all links and buttons are centered. The buttons are full width on smaller screens. On larger screens, the "Support Us" button is moved to the right side along with the "Sign In" button.

### Old
<table>
<tr>
<td>

</td>
<td>

</td>
</tr>
</table>

### New
With SignIn
<table>
<tr>
<td>

</td>
<td>

</td>
</tr>
</table>

Without signIn
<table>
<tr>
<td>

</td>
<td>

</td>
</tr>
</table>

### DIFF
--- a/dispatcher/frontend-ui/src/components/NavBar.vue
+++ b/dispatcher/frontend-ui/src/components/NavBar.vue
@@ -1,7 +1,7 @@
 
 <template>
     <header>
-        <nav class="navbar navbar-expand-md navbar-dark ">
+        <nav class="navbar navbar-expand-md navbar-dark">
             <a class="navbar-brand branding">
                 <div class="icon"><img :src="publicPath + 'assets/logo.svg'" />
                 <Loading/></div>
@@ -10,7 +10,7 @@
 
             <b-navbar-toggle target="nav-collapse"></b-navbar-toggle>
              <b-collapse id="nav-collapse" is-nav>
-                <ul class="navbar-nav mr-auto">
+                <ul class="navbar-nav mr-auto text-center mb-md-0 mb-2">
                     <li class="nav-item">
                         <router-link class="nav-link" :to="{ name: 'pipeline' }">Pipeline</router-link>
                     </li>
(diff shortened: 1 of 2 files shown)


dispatcher/frontend-ui/src/components/NavBar.vue
```diff
@@ -49,3 +50,11 @@
     components: {Loading, UserButton},
   }
 </script>
+
+<style type="text/css" scoped>
+   @media (min-width: 768px) {
+    .w-md-auto {
+      width: auto !important;
```
benoit74, line 57: Do we really need this? Can't this be solved by using `w-md-auto w-100` rather than `w-100 w-md-auto`? Or with conditional classes? I'm unsure but I feel it odd to have to set an `!important` and specifiy again the breakpoint width (pretty fragile should we decide to move breakpoints at some point).

Jainex17: w-100 has !important by default, so if we remove !important from our w-md-auto, it won't overwrite the w-100 whatever position we put it at. And without w-md-auto, the button looks something like this. If we don't add the breakpoint, then only one width applies.

I also tried adjusting the class order and removing the breakpoint or the !important flag, but couldn’t get it to behave correctly. So, for now, this seems like the most reliable solution unless there's a cleaner workaround I’ve missed. Open to suggestions if you have a better idea though.

image

benoit74: I will need to have a look then

benoit74: Got it, LGTM

Jainex17: Thanks for review

benoit74 (commented):
In fact, I find that `width: 100%` is kinda ugly at intermediate size like below:

image

Shouldn't we remove the `w-100` and let the size adjust to content like below?

image

Jainex17:
> In fact, I find that `width: 100%` is kinda ugly at intermediate size like below:
> 
> image
> 
> Shouldn't we remove the `w-100` and let the size adjust to content like below?
> 
> image

Yeah, it looks kinda ugly on medium screen sizes (425px to 760px). Should I remove w-100 so it adjusts the width automatically?

benoit74:
Yes, please do

benoit74 (changes requested):
LTGM, please squash commits and I will merge

Jainex17:
Done

Jainex17:
@benoit74  Is there anything I need to do on my end? Thanks

benoit74 (approved):
LGTM

PR #1106, opened 2025-05-22
use hatch for dependency management and mv disptacher subdirs to root

elfkuzco:
## Rationale
This PR lays the foundation for the plan to update the zimfarm backend (formerly known as `dispatcher`).

## Changes
- move subdirectories in `dispatcher` to root directory
- Use hatch for dependency management
- upgrade project python version to 3.13
- update QA to use newer actions and new tooling
- add type annotations to functions in `common` and `db` folders
- replace marshmallow with pydantic (only as a dependency yet)
- replace flask with fastapi (only as a dependency yet)


benoit74 (changes requested):
Thank you, few small changes suggested

backend/src/zimfarm_backend/common/external.py
```diff
@@ -159,18 +161,6 @@ def get_statement() -> dict[str, Any]:
     )
 
 
-class ExternalIpUpdater:
```
benoit74, line 152: I'm so sorry, I just realize now why we might need this: for tests. I used this to inject a different behavior during tests. Don't know if this is feasible without this class, but I'm not sure this will work with a static method either. Anyway this way kinda messy. Let's merge this without this class, you will anyway be able to easily restore it should it prove to be mandatory for testing.

benoit74 (approved):
LGTM, shall I merge or do you prefer to squash commits first? (git history is anyway going to be hard to read for this bunch of huge changes ...)

PR #1109, opened 2025-05-26
convert marshmallow models to pydantic models

elfkuzco:
## Rationale
Convert marshmallow models to pydantic models.

## Changes
- use type annotations to compose reusable types
- use enums as field types in stead of calling `.all()`
- use equivalent pydantic functionalities to mirror marshmallow code.


elfkuzco:
Regarding the failing CI for black, most of the files it flags are not from changes in the commit. will work on a fix for it. probably because it's a different black version than the old code.

backend/src/zimfarm_backend/common/schemas/offliners/gutenberg.py
```diff
@@ -1,135 +1,108 @@
...
-        metadata={
-            "label": "Optimization Cache URL",
-            "description": "S3 Storage URL including credentials and bucket",
-            "secret": True,
-        },
-        data_key="optimization-cache",
+    optimization_cache: S3OptimizationCache = Field(
+        title="Optimization Cache URL",
+        description="S3 Storage URL including credentials and bucket",
+        alias="optimization-cache",
```
benoit74, line 95: Wouldn't it be simpler to override base class with a DashField like we did at https://github.com/openzim/mindtouch/blob/4772c2c2664a58b375e5490e4ff44c128e29259f/scraper/src/mindtouch2zim/ui.py#L5-L8 (but with a CamelField). Not sure there is a dash alias generator, but could be implemented in worst case scenario.

backend/src/zimfarm_backend/common/schemas/offliners/nautilus.py
```diff
@@ -1,192 +1,146 @@
...
+        title="Header Logo",
+        description=("Custom logo. Will be resized to 300x65px. Nautilus otherwise."),
+        alias="main-logo",
+        default="Nautilus",
+    )
+    secondary_logo: AnyUrl | None = Field(
+        title="Footer logo",
+        description=("Custom footer logo. Will be resized to 300x65px. None otherwise"),
+        alias="secondary-logo",
+        default=None,
```
benoit74, line 113: Wouldn't it be possible to encapsulate all these `default=None` fields in an `OptionalField` class which automatically sets this default unless override?

elfkuzco:
> I'm also concerned by all the fields which moved to `NotEmptyString` type. Does it means they are required? In marshmallow, default behavior is `required: False`, so it was permitted to not pass most of the parameters.

Okay. Thing is Pydantic makes all fields required by default. I want to experiment with creating an `OptionalField` and then write it as:
```py
field : NotEmptyString | None = OptionalField(....)
```
Not sure how I can avoid  `NotEmptyString | None` on the left hand side of the `=` but having `OptionalField` set the `default=None` is doable, I believe. It would also allow me to reuse `NotEmptyString` unless across, the codebase, all fields are optional.

backend/src/zimfarm_backend/common/schemas/offliners/nautilus.py
elfkuzco: Yeah, this can be done, I think.

backend/src/zimfarm_backend/common/schemas/offliners/mindtouch.py
```diff
@@ -1,215 +1,164 @@
...
+        alias="bad-assets-threshold",
+        default=10,
+    )
+
+    stats_filename: NotEmptyString = Field(
+        title="Stats filename",
+        description="Scraping progress file. Leave it as `/output/task_progress.json`",
+        alias="stats-filename",
+        default="/output/task_progress.json",
+        validate_default=True,
```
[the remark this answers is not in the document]
elfkuzco: `validate_default` merely ensures that any default we set matches the regex. Just an unnecessary safety hatch in the event we set a default that doesn't conform to our pattern.

backend/src/zimfarm_backend/common/schemas/offliners/devdocs.py
```diff
@@ -1,151 +1,110 @@
...
-        truthy=[True],
-        falsy=[False],
-        metadata={
-            "label": "All",
-            "description": "Fetch all Devdocs resources, and produce one ZIM "
-            "per resource. Mutually exclusive with `Slug` setting, set only "
-            "one option. Either this setting or `Slug` must be configured.",
-        },
-        data_key="all",
+    all_flag: bool | None = OptionalField(
```
elfkuzco, line 22: If I add the generator which converts snake_case to kebab-case, how would it be used for this class?

backend/src/zimfarm_backend/common/schemas/fields.py
```diff
@@ -1,77 +1,132 @@
...
+
+ZIMLangCode = Annotated[str, Field(min_length=2, max_length=3)]
+
+type OptionalZIMLangCode = ZIMLangCode | None
+
+ZIMOutputFolder = Annotated[str, Field(pattern=r"^/output$")]
+
+type OptionalZIMOutputFolder = ZIMOutputFolder | None
+
+ZIMCPU = Annotated[int, Field(ge=0)]
```
benoit74: This should be `gt`, not `ge` (same for memory and disk)

backend/src/zimfarm_backend/common/schemas/fields.py
```diff
@@ -1,77 +1,132 @@
...
+
+ZIMDisk = Annotated[int, Field(ge=0)]
+
+type OptionalZIMDisk = ZIMDisk | None
+
+SkipField = Annotated[int, Field(default=0, ge=0)]
+
+type OptionalSkipField = SkipField | None
+
+LimitFieldMax500 = Annotated[int, Field(default=20, ge=0, le=500)]
```
benoit74: Both `Limitxxx` fields should be `gt` rather than `ge` (fetching 0 items make no sense)

backend/src/zimfarm_backend/common/schemas/models.py
```diff
@@ -41,46 +51,45 @@
...
-class DockerImageSchema(Schema):
-    name = String(required=True, validate=validate.OneOf(DockerImageName.all()))
-    tag = String(required=True)
+class DockerImageSchema(BaseModel):
+    name: DockerImageName
+    tag: NotEmptyString
 
-    @pre_load
-    def strip_prefix(self, in_data, **kwargs):
+    @field_validator("name", mode="before")
```
benoit74, line 69: I don't think this needs to be done in a before validator. An after validator seems to be enough. Or do I miss something? 

Plus this should probably be moved to the DockerImageName definition.

backend/src/zimfarm_backend/common/schemas/models.py
```diff
@@ -105,45 +114,44 @@ def get_offliner_schema(offliner):
             Offliner.freecodecamp: FreeCodeCampFlagsSchema,
             Offliner.devdocs: DevDocsFlagsSchema,
             Offliner.mindtouch: MindtouchFlagsSchema,
-        }.get(offliner, Schema)
+        }.get(offliner, BaseModel)
 
-    @validates_schema
-    def validate(self, data, **kwargs):
+    @model_validator(mode="before")
```
benoit74: I feel like we are keeping a hack here.

Shouldn't we use discriminated unions here: https://docs.pydantic.dev/latest/concepts/unions/#discriminated-unions-with-str-discriminators ?

If I understand this correctly, we should probably have a Union on `MWOfflinerScheduleConfigSchema`, `YoutubeScheduleConfigSchema`, ... (to be created) with the discriminator being the `task_name`. 

If we do so, I would love to be able to not repeat the common settings (warehouse_path, image, resources, ...) in each of these `MWOfflinerScheduleConfigSchema`, `YoutubeScheduleConfigSchema`, ... schemas so that we avoid repeating ourselves. Probably possible with multiple inheritance (BaseModel + SharedScheduledConfigSchema).

I feel like this would use pydantic at its best, but I might miss something.

backend/src/zimfarm_backend/common/schemas/models.py
```diff
@@ -105,45 +114,44 @@ def get_offliner_schema(offliner):
...
+    config: ScheduleConfigSchema
+    notification: ScheduleNotificationSchema | None = None
 
 
-PlatformsLimitSchema = Schema.from_dict(
-    {
-        platform: fields.Integer(required=False, validate=validate_platform_value)
-        for platform in Platform.all()
-    }
+PlatformsLimitSchema = (  # pyright: ignore[reportUnknownVariableType, reportCallIssue]
```
benoit74: This feel too hackish, I think we can rework this in a more type-safe manner without big consequences with such a schema (not tested):

```
class PlaftormLimitSchema:
   platform: Platform
   limit: ZimPlatformValue

class PlatformsLimitSchema(BaseModel):
   limits: list[PlaftormLimitSchema] = Field(default_factory=list)
```

We might need to force to have at least one limit in the list, but I'm not even sure this is a requirement.

backend/src/zimfarm_backend/common/schemas/offliners/freecodecamp.py
```diff
@@ -1,136 +1,97 @@
...
     )
 
-    title = String(
-        metadata={
-            "label": "Title",
-            "description": "ZIM title",
-        },
-        required=True,
-        validate=validate_zim_title,
+    title: ZIMTitle = OptionalField(
```
benoit74: this is a required property, so `Field` is probably better if I get it correctly

backend/src/zimfarm_backend/common/schemas/offliners/gutenberg.py
```diff
@@ -1,135 +1,95 @@
...
     # in command_for() (offliners.py)
-    one_language_one_zim = fields.Boolean(
-        truthy=[True, "/output"],
-        falsy=[False],
-        metadata={
-            "label": "Multiple ZIMs",
-            "description": "Create one ZIM per language",
-        },
-        data_key="one-language-one-zim",
+    one_language_one_zim: bool | OptionalZIMOutputFolder = OptionalField(
```
benoit74: why `bool | OptionalZIMOutputFolder`? `bool | None`?

backend/src/zimfarm_backend/common/schemas/offliners/youtube.py
```diff
@@ -1,260 +1,168 @@
...
-            "label": "Temp folder",
-            "placeholder": "/output",
-            "description": "Technical flag: Where to create temporay build folder. "
-            "Leave it as `/output`",
-        },
-        load_default="/output",
-        dump_default="/output",
-        validate=validate_output,
-        data_key="tmp-dir",
+    tmp_dir: OptionalZIMOutputFolder = OptionalField(
```
benoit74, line 165: Why specifying the pattern again? Plus this can be something else than `/output`, this is just the default value, where `ZIMOutputFolder` forces it to be `/output`, isn't it?

Same comment apply to (mostly?) all tmp_dir fields.

backend/src/zimfarm_backend/common/schemas/offliners/mindtouch.py
```diff
@@ -1,215 +1,140 @@
...
-        validate=validate_output,
+class MindtouchFlagsSchema(DashModel):
+    library_url: NotEmptyString = Field(
+        title="Library URL",
+        description="URL of the Mindtouch / Nice CXone Expert instance (must NOT"
+        " contain trailing slash), e.g. for LibreTexts Geosciences it is "
+        "https://geo.libretexts.org",
+    )
+
+    creator: OptionalNotEmptyString = OptionalField(
```
benoit74: creator is required in this scraper

backend/src/zimfarm_backend/common/schemas/offliners/mwoffliner.py
```diff
@@ -1,306 +1,205 @@
...
+        title="Publisher",
+        description="ZIM publisher metadata. `openZIM` otherwise.",
+    )
+    filenamePrefix: OptionalNotEmptyString = OptionalField(
+        title="Filename prefix",
+        description="Custome filename up to the formats and date parts.",
+    )
+    formats: (
+        list[
+            Literal[
```
benoit74: Please move this to a class just like all other enums (class will live in mwoffliner.py, but still way clearer this way IMHO)

backend/src/zimfarm_backend/common/schemas/offliners/mwoffliner.py
```diff
@@ -1,306 +1,205 @@
...
-            [
-                "VisualEditor",
-                "WikimediaDesktop",
-                "WikimediaMobile",
-                "RestApi",
-                "ActionParse",
-            ]
-        ),
+    forceRender: (
+        Literal[
```
benoit74: Move this to a class enum as well

backend/src/zimfarm_backend/common/schemas/offliners/zimit.py
```diff
@@ -1,778 +1,600 @@
...
-            "Pixel 4 landscape",
-            "Pixel 4a (5G)",
-            "Pixel 4a (5G) landscape",
-            "Pixel 5",
-            "Pixel 5 landscape",
-            "Moto G4",
-            "Moto G4 landscape",
-        ]
-    )
+DEVICES = sorted(
```
benoit74: Why not a class enum as "usual"?

benoit74 (changes requested):
I'm very happy of the direction this PR is moving the Zimfarm. 

I still have comments that needs to be addressed, but we are definitely close to something significantly better than this was

backend/src/zimfarm_backend/common/schemas/offliners/zimit.py
benoit74, line 1: Maybe we could split ZimitFlagsSchema into `ZimitDashFlagsSchema(DashModel)` and `ZimitCamelFlagsSchema(CamelModel)` so that we do not have to specify alias for all fields in camel case.

Then `ZimitFlagsSchema` would simply be the concatenation of both classes:

```
class ZimitFlagsSchema(ZimitCamelFlagsSchema, ZimitDashFlagsSchema):
  ...
```

WDYT?

backend/src/zimfarm_backend/common/schemas/models.py
elfkuzco: I used a `before` so we can strip out any `ghcr.io/` in the name. I didn't move it to the `DockerImageName` because that's just an enum. So, we get validation that any name we supply is supported. Having the validator run `before` allows me to do:
```py
d = DockerImageSchema(name="ghcr.io/openzim/mwoffliner", tag="v1.1")
```
as the field validator will strip away the prefix `ghcr.io/` leaving me with `openzim/mwoffliner` which is a valid enum value.

If it were `after`, the above example would always fail because there is no enum with a value of `ghcr.io/openzim/mwoffliner`

backend/src/zimfarm_backend/common/schemas/offliners/zimit.py
elfkuzco: sounds good.

backend/src/zimfarm_backend/common/schemas/models.py
elfkuzco: Discriminators assume there's a common field which will be used to infer the concrete model. Unfortunately, our offliners do not have a common field to be used for this. Using callable either won't help either. However, here's a method that I think would be simpler:

```py
from typing import Annotated, Union
from pydantic import Discriminator

# Create a base flags schema that all flag schemas will inherit from
class BaseFlagsSchema(BaseModel):
    pass

# Create a discriminated union of all flag schemas
FlagsSchemaUnion = Annotated[
    Union[
        MWOfflinerFlagsSchema,
        YoutubeFlagsSchema,
        GutenbergFlagsSchema,
        # rest of the schemas...
    ],
    Discriminator("task_name")
]

class ScheduleConfigSchema(BaseModel):
    task_name: Offliner
    warehouse_path: WarehousePath
    image: DockerImageSchema
    resources: ResourcesSchema
    flags: FlagsSchemaUnion
    platform: Platform | None = None
    artifacts_globs: list[NotEmptyString] = Field(default_factory=list)
    monitor: bool

    @model_validator(mode="before")
    def validate_flags(cls, data: dict[str, Any]) -> dict[str, Any]:  # noqa: N805
        if "task_name" in data and "flags" in data:
            # The discriminator will automatically validate the flags based on task_name
            data["flags"]["task_name"] = data["task_name"]
        return data
```
Now, when this model is called, the `model_validator` injects the `task_name` in the `flags` dictionary. This would prevent calls to pass the full dictionary with an additional `task_name` key as in
 ```py
      flags": {
        "task_name": "mwoffliner",
        # ...other fields
    }
 ```
 However, the cost of this is that we won't be able to do `if...else` inside the union to determine which type of schema like we do for relaxed variants of the schema.
 WDYT?

elfkuzco:
Pending https://github.com/openzim/zimfarm/pull/1109#discussion_r2125184964, the new commit addresses the rest of the reviews.

backend/src/zimfarm_backend/common/schemas/models.py
benoit74: Got it, thanks. Please add a small a comment explaining this in the source code so that it is more obvious.

backend/src/zimfarm_backend/common/schemas/models.py
benoit74: Then we should do a data migration to move `task_name`. It not that a big deal, and it's anyway something I felt like we should do at some point.

I propose to rename `flags` to `offliner`, and move `task_name` to `name` field inside `offliner` field.

Does it makes sense to you?

Then regarding conditional for relaxed variants, it looks like it should work:

```python
from typing import Literal, Union
from pydantic import BaseModel, Field


class ZimitRelaxSchema(BaseModel):
    name: Literal["zimit"]
    url: str


class ZimitFullSchema(ZimitRelaxSchema):
    description: str


class MwofflinerSchema(BaseModel):
    name: Literal["mwoffliner"]
    renderer: str


print(ZimitRelaxSchema(name="zimit", url="https://www.acme.com"))
# print(ZimitFullSchema(name="zimit", url="https://www.acme.com"))

print(ZimitFullSchema(name="zimit", url="https://www.acme.com", description="A desc"))

use_zimit_relax = False
OfflinerSchema = Union[
    ZimitRelaxSchema if use_zimit_relax else ZimitFullSchema, MwofflinerSchema
]


class ScheduleConfigSchema(BaseModel):
    warehouse_path: str
    offliner: OfflinerSchema = Field(discriminator="name")


print(
    ScheduleConfigSchema(
        offliner={"name": "mwoffliner", "renderer": "actionparse"},
        warehouse_path="/other",
    )
)
# print(ScheduleConfigSchema(offliner = {'name': "zimit", 'renderer': "actionparse"}, warehouse_path="/other"))
# print(ScheduleConfigSchema(offliner = {'name': "zimit", 'url': "https://www.acme.com"}, warehouse_path="/other"))
print(
    ScheduleConfigSchema(
        offliner={
            "name": "zimit",
            "url": "https://www.acme.com",
            "description": "A website",
        },
        warehouse_path="/other",
    )
)
```

elfkuzco: Okay. This works, except that we might want to use a different name for the discriminator. I can see some `*FlagscSchema` already have `name` attribute

https://github.com/openzim/zimfarm/blob/a34ef396e4d0b518930cb05fd15f58fbcfb4864b/dispatcher/backend/src/common/schemas/offliners/freecodecamp.py#L52-L59

Should we choose a different name and exclude it from serialization?

benoit74: Let's choose a different one then, indeed. Maybe something mostly unique like `offliner_id`?

And it should be serialized to JSON for DB + HTTP responses, otherwise we do not really know what object we are dealing with. Only logic to generate CLI command needs to drop it for obvious reasons.

backend/src/zimfarm_backend/common/schemas/parameters.py
```diff
@@ -1,167 +1,163 @@
...
+    ZIMDisk,
+    ZIMMemory,
 )
-from common.schemas.models import (
+from zimfarm_backend.common.schemas.models import (
     DockerImageSchema,
+    EventNotificationSchema,
     LanguageSchema,
-    PlatformsLimitSchema,
+    PlatformsLimitSchema,  # pyright: ignore[reportUnknownVariableType]
```
benoit74: You probably don't need this type hint anymore

backend/src/zimfarm_backend/common/schemas/models.py
```diff
@@ -41,109 +47,97 @@
...
-    notification = fields.Nested(
-        ScheduleNotificationSchema(), required=False, dump_default={}, load_default={}
+)
+
+
+class ScheduleConfigSchema(BaseModel):
+    warehouse_path: WarehousePath
+    image: DockerImageSchema
+    resources: ResourcesSchema
+    flags: OfflinerSchema = Field(  # pyright: ignore[reportInvalidTypeForm]
```
benoit74: please rename this from `flags` to `offliner` so that it is even more obvious that we forget to update something in routes / DB (no need to update the whole codebase now, to be done route by route)

backend/src/zimfarm_backend/common/schemas/models.py
```diff
@@ -41,109 +47,97 @@
...
+
+    @field_validator("name", mode="before")
+    def validate_name(cls, v: str) -> str:  # noqa: N805
+        # docker images can have a prefix, e.g. ghcr.io/
+        # set mode="before" and strip the image prefix so that the remaining name is
+        # validated against the DockerImageName enum.
+        return re.sub(r"^ghcr.io/", "", v)
+
+
+OfflinerSchema = (
```
benoit74, line 77: I don't think this is correct, given indentation and parentheses: if `constants.NAUTILUS_USE_RELAXED_SCHEMA`, then you do not have access to TedFlagsSchema and others. You probably miss at least some  parenthesis

elfkuzco: mistake. resolve now

benoit74: I don't think so. the Union stops at `NautilusFlagsSchemaRelaxed` when `constants.NAUTILUS_USE_RELAXED_SCHEMA` is `True` ; `TedFlagsSchema` and others are not included.

benoit74: This seems to be the correct code:

```python
OfflinerSchema = (
    MWOfflinerFlagsSchema
    | YoutubeFlagsSchema
    | GutenbergFlagsSchema
    | PhetFlagsSchema
    | SotokiFlagsSchema
    | (
        NautilusFlagsSchemaRelaxed
        if constants.NAUTILUS_USE_RELAXED_SCHEMA
        else NautilusFlagsSchema
    )
    | TedFlagsSchema
    | OpenedxFlagsSchema
    | (
        ZimitFlagsSchemaRelaxed
        if constants.ZIMIT_USE_RELAXED_SCHEMA
        else ZimitFlagsSchema
    )
    | KolibriFlagsSchema
    | WikihowFlagsSchema
    | IFixitFlagsSchema
    | FreeCodeCampFlagsSchema
    | DevDocsFlagsSchema
    | MindtouchFlagsSchema
)
```

benoit74 (approved):
LGTM

PR #1111, opened 2025-05-30
add route for getting languages

elfkuzco:
## Changes
- add route for getting list of languages
- split db code from routes
- write tests for db and routes


benoit74 (approved):
LGTM

We will merge it once `auth-routes` branch is merged.

PR #1112, opened 2025-06-03
add routes for requesting tasks

elfkuzco:
## Rationale

Add routes to request tasks

## Changes
- move db functions away from utils and routes
- refactor routes from flask to fastapi


elfkuzco:
@benoit74 , I tried writing tests for the function that retrieves a task from the db but it errors with the message "unable to cast datetime to integer". I suppose the error is from here while creating a task:
https://github.com/openzim/zimfarm/blob/6e0df354487a7d22765dd0653ca72b5db90b09d7/dispatcher/backend/src/utils/scheduling.py#L155

but while fetching from the db, we do this:
https://github.com/openzim/zimfarm/blob/6e0df354487a7d22765dd0653ca72b5db90b09d7/dispatcher/backend/src/routes/requested_tasks/requested_task.py#L99-L102

Does this mean that a `$date` is later added to the `requested` key of the `timestamp`?

benoit74:
> Does this mean that a $date is later added to the requested key of the timestamp?

Not really, this transformation is done by `bson` module, which is hooked into SQLAlchemy with following code:

https://github.com/openzim/zimfarm/blob/a34ef396e4d0b518930cb05fd15f58fbcfb4864b/dispatcher/backend/src/db/__init__.py#L29-L36
 
See `json_serializer` and `json_deserializer` functions. Not that for deserializing, we use a custom function to handle naive vs aware datetimes, but this is not linked to the `$date` additional thing.

Using `bson` module (coming from MongoDB) for serializing JSON was deemed a good compromise to avoid having our own logic / handle complex stuff like dates on our own. I don't know if we could have used marshmallow for that. I don't know if pydantic could do it. I'm definitely open to an enhancement on this, but this is maybe a distinct issue.

Your problem is maybe just that you do not hook `bson` when you create the SQLAlchemy session for your tests? But this is weird since you did not seem to have changed stuff around that.

Is that clearer?

elfkuzco:
Okay. works now. except for it being unable to cast the date to an integer
```
sqlalchemy.exc.DataError: (psycopg.errors.InvalidTextRepresentation) invalid input syntax for type bigint: "2025-06-04T02:04:24.318Z"
```

elfkuzco:
Ended up casting the timestamp fields to `TIMESTAMP` instead of `BigInteger`. Worried that it worked in the older code.

benoit74:
> Ended up casting the timestamp fields to TIMESTAMP instead of BigInteger. Worried that it worked in the older code.

Indeed, never knew why I had to cast to BigInteger. Did you tried without any cast at all?

backend/pyproject.toml
```diff
@@ -183,7 +183,7 @@ ignore = [
   # Allow non-abstract empty methods in abstract base classes
   "B027",
   # Allow datetime methods with tz info
-  "DTZ001", "DTZ006",
+  "DTZ001", "DTZ006", "DTZ011",
```
benoit74: Why adding `DTZ011` to ignore list? I feel like this rule is correct, and we should fix the code to not use calls to `datetime.today` and prefer to use `getnow` utility function which we already have coded

backend/src/zimfarm_backend/common/schemas/orms.py
```diff
@@ -133,7 +137,7 @@ class RequestedTaskLightSchema(BaseModel):
     Schema for reading a requested task model with some fields
     """
 
-    id: str = Field(alias="_id")
+    id: UUID
```
benoit74, line 105: You need to be consistent in `id` field handling. `RequestedTaskLightSchema` has different definition than `MostRecentTaskSchema` now (and probably others). I don't mind we decide to rename field `_id` to `id`, it will break the UI but it is easy to fix, and it is probably a wise move anyway. But this needs to be done "everywhere" for consistency. And all those ids are UUID.

backend/src/zimfarm_backend/routes/utils.py
```diff
@@ -23,24 +29,33 @@ def has_dict_sub_key(data: Dict[str, Any], keys: List[str]):
...
-    )().to_desc()
-    fields = [field["data_key"] for field in schema if field.get("secret", False)]
+    schema = ScheduleConfigSchema.get_offliner_schema(response["config"]["task_name"])
+    fields = [
+        field.alias
+        for field in schema.model_fields.values()
+        if field.alias
+        and field.annotation
+        in (
+            AnyUrl,
```
benoit74: Why adding `AnyUrl`? In general URLs are not secret. Do you have an example of an offliner field which was secret and a URL? (besides the OptionalSecretStr, SecretStr and S3OptimizationCache)

backend/src/zimfarm_backend/routes/utils.py
```diff
@@ -23,24 +29,33 @@ def has_dict_sub_key(data: Dict[str, Any], keys: List[str]):
...
     # in commands
-    if "config" not in response or "flags" not in response["config"]:
+    if "config" not in response or "offliner" not in response["config"]:
         return
 
-    schema = ScheduleConfigSchema.get_offliner_schema(
-        response["config"]["task_name"]
-    )().to_desc()
-    fields = [field["data_key"] for field in schema if field.get("secret", False)]
+    schema = ScheduleConfigSchema.get_offliner_schema(response["config"]["task_name"])
```
benoit74: Rather than calling this function, I feel like we should do it "Pydantic-way":
- use [Serialization Context](https://docs.pydantic.dev/latest/concepts/serialization/#serialization-context) to let serializers know when we want to hide secrets in model fields
- use custom serializers for `OptionalSecretStr`, `SecretStr` and `S3OptimizationCache` fields which will use this context to hide what needs to be hidden

Does it make any sense?

backend/src/zimfarm_backend/routes/utils.py
```diff
@@ -74,27 +89,27 @@ def remove_secret_flags(response: dict):
...
-    response[root_key_value]["command"][
-        index
-    ] = f'--{field_name}="{SECRET_REPLACEMENT}"'
+    response[root_key_value]["command"][index] = (
+        f'--{field_name}="{SECRET_REPLACEMENT}"'
+    )
 
 
-def remove_url_secrets(response: dict):
+def remove_url_secrets(response: dict[str, Any]):
```
benoit74: Could this be replaced by a custom serializer: https://docs.pydantic.dev/latest/concepts/serialization/#custom-serializers?

backend/src/zimfarm_backend/routes/utils.py
```diff
@@ -74,27 +89,27 @@ def remove_secret_flags(response: dict):
...
     """remove secrets from any URL we might find
 
     For now we remove:
     - password if passed in the network location
     - keyId and secretAccessKey query params (probably used for S3)
     """
-    for key in response.keys():
+    for key, _ in response.items():
```
benoit74: Why using `items` rather than `keys` if you finally drop the `value`?

backend/src/zimfarm_backend/routes/utils.py
```diff
@@ -104,13 +119,13 @@ def remove_url_secrets(response: dict):
                 continue
             for url in [word for word in response[key].split() if "://" in word]:
                 try:
-                    urlparts = urlsplit(url)
+                    urlparts = cast(SplitResult, urlsplit(url))
```
benoit74: Shouldn't we handle the case where result is not a `SplitResult` explicitly so that we do not fail with a very weird error because urlparts is missing some fields?

benoit74 (changes requested):
See inline comments

backend/src/zimfarm_backend/routes/utils.py
elfkuzco: This was a complaint from Ruff. Didn't want to end up adding another `#noqa` as it kept complaining. 

Should stop listening to these tools more now that I think about it as they aren't the rule. 😅

benoit74: This is weird, it is not complaining on my machine ... you should listen to these tools indeed, but you have to understand why they complain ... ^^

elfkuzco:
Pydantic has SecretStr which hides secrets whenever we dump the schema. So, I used a context `show_secrets` to show the values. This is helpful when dumping the config to the db and for cases where `hide_secrets` is `False` in the request

benoit74 (approved):
Thank you very much ! I love this. Kinda a very big step forward.

This is such a big PR with significant code move that I'm sure I could have missed many subtle bugs but the fact that tests are extensive is providing lot of trust in this change.

And I'm very pleased by all type safety and automated serialization in all directions.

Let's merge!

PR #1129, opened 2025-06-20
add routes for schedules

elfkuzco:
## Changes

- add routes for schedules
- convert notifications dict to pydantic model
- forbid extra attributes from `CamelModel` and `DashModel` used by Offliner schemas. Helps prevent passing wrong flags in update schema even though pydantic normally omits them.
- convert schedule duration dict to pydantic model


backend/src/zimfarm_backend/db/requested_task.py
```diff
@@ -287,11 +291,11 @@ def compute_task_eta(session: OrmSession, task: Task) -> dict[str, Any]:
 
 
 class RunningTask(BaseModel):
-    config: dict[str, Any]
+    config: ExpandedScheduleConfigSchema
     schedule_name: str
     timestamp: dict[str, Any]
```
benoit74, line 296: We can leave it for another PR, but I think this typing is wrong (was here before you came): this is a list of events, supposed to be ordered by order of arrival, and under extreme case (I don't recall exactly, but it is linked to the cancellation of a task while it was just starting on the worker) we can have multiple time the same event (key in the dictionary) and it is a bit sad we loose this info. I let you decide if you want to fix this now or open a distinct issue to fix it a bit later.

backend/src/zimfarm_backend/routes/schedules/logic.py
```diff
@@ -0,0 +1,372 @@
...
+
+    return JSONResponse(content=content)
+
+
+@router.get("/{schedule_name}")
+def get_schedule(
+    schedule_name: Annotated[ScheduleNameField, Path()],
+    session: OrmSession = Depends(gen_dbsession),
+    current_user: User | None = Depends(get_current_user_or_none),
+    hide_secrets: Annotated[bool | None, Query()] = True,  # noqa: FBT002
```
benoit74: Do we really need this `FBT002` hint? I think that parameters are always passed by name by FastAPI, so we could add the `*, hide_secrets: ...` that `FBT002` want us to add for boolean args (to force passing parameter by name)

benoit74 (changes requested):
Thank you! Looks pretty good, only two comments to address (or even postpone for the first one).

backend/src/zimfarm_backend/db/requested_task.py
elfkuzco: I will create a new issue to model it properly. I was under the the impression it came from the Task config here https://github.com/openzim/zimfarm/blob/e1cbd9211113575b12425bdfe81ce4dc70024f25/dispatcher/backend/src/utils/scheduling.py#L367-L381

Will craft a better model once I explore the db model

elfkuzco: I have created #1130 to address this.

PR #1135, opened 2025-06-26
set up dockerfile and remove old files

elfkuzco:
## Changes
- set up docker file for updated API
- remove old test files
- move newer test files to top level directory `tests`
- modify dev `docker-compose.yaml`
- uncomment backend QA CI file


dev/docker-compose.yml
```diff
@@ -3,7 +3,7 @@ services:
     image: postgres:17.3-bookworm
     container_name: zf_postgresdb
     ports:
-      - 127.0.0.1:5432:5432
+      - 2345:5432
```
benoit74: - why prefering `2345` than `5432`? I can understand that on your machine you do prefer this port because 5432 is already used, but I don't get why we would want to push this as a default for everyone
- why removing `127.0.0.1` prefix? is there any reason to expose this on the network (by default at least, again)?

dev/docker-compose.yml
```diff
@@ -12,20 +12,24 @@ services:
...
       - POSTGRES_PASSWORD=zimpass
+    healthcheck:
+      test: ["CMD", "pg_isready", "-q", "-d", "dbname=zimfarm user=zimfarm"]
+      interval: 10s
+      timeout: 5s
+      retries: 3
   backend:
-    build: ../dispatcher/backend
+    build:
+      context: ../backend
```
benoit74, line 22: This context must be changed for `backend-tools` and `backend-tests` as well, and also for `frontend-ui` maybe

backend/Dockerfile
```diff
@@ -43,12 +20,30 @@ ENV MAILGUN_FROM Zimfarm <info@farm.openzim.org>
 # ENV suffixed with event: `GLOBAL_NOTIFICATION_started` or `GLOBAL_NOTIFICATION_ended`
 # ENV GLOBAL_NOTIFICATION_ended slack,#zimfarm-events,@rgaudin|mailgun,reg@kiwix.org
 
+COPY src /app/src
```
benoit74: We should take this opportunity of "big breaking changes" to optimize the Docker image as well.

Here, you add the `src` content very early in the docker file. This is often what has changed (compared to what is below which changes way more rarely). This that means that whenever you rebuild the image because `src` has changed, you have many things to rebuild, including reinstalling all Python dependencies. This is not the recommended approach. This has negative consequence both in terms of build time but also in terms of storage consumption ; this is of course mostly true when you rebuild many times on the same machine ... i.e. only on our dev machines, not much on CI which rebuild from scratch everytime

We should prefer to have mostly "static" things first, and fast-changing stuff later. You can see this in action at https://github.com/openzim/mindtouch/blob/main/Dockerfile for instance. 

Here I thing that we should have:
- copy and "install" pre-start.sh, periodic-schedule, periodic-tasks, supervisor-listener, periodic.conf
- copy "pyproject.toml" and required dependencies and install it (will only install Python deps in fact)
- copy "src" and install it (if needed, but I think so unless I'm mistaken)

We can discuss this live if it is not really clear for you.

benoit74 (changes requested):
Thank you, few remarks

benoit74:
Oups, sent comment too fast, just edited

elfkuzco:
> This is better, but this is more what I meant:
> 
> ```
> COPY maint-scripts /app/maint-scripts
> 
> # install Python dependencies first (since they change more rarely than src code)
> COPY pyproject.toml README.md /app/
> COPY src/zimfarm_backend/__about__.py /app/src/zimfarm_backend/__about__.py
> RUN pip install --no-cache-dir /app
> 
> # install src code itself
> COPY src /app/src
> RUN pip install --no-cache-dir /app  \
>     && rm -rf /app/src \
>     && rm /app/pyproject.toml \
>     && rm /app/README.md
> 
> EXPOSE 80
> ```

Done. Will squash and merge once everything looks good.

dev/docker-compose.yml
```diff
@@ -3,7 +3,7 @@ services:
     image: postgres:17.3-bookworm
     container_name: zf_postgresdb
     ports:
-      - 127.0.0.1:5432:5432
+      - 127.0.0.1:5433:5432
```
benoit74: `5432`? 🤣

benoit74 (approved):
LGTM, please fix the `5432` issue, squash and merge

PR #1137, opened 2025-06-26
Set up nav bar and api services

elfkuzco:
## Changes
- set up stub services to query backend API
- set up navbar using material design and vuetify
- set up tooling for typescript and vue 3


benoit74 (changes requested):
Very good for step!

Some remarks:
- using `import.meta.env.VITE_` env variables is not appropriate for our usage ; these env variables are replaced by vite at build time, while we need to customize them at runtime (at least `VITE_ZIMFARM_BACKEND_URL`) since Docker image will be already built but location of backend will change from one deployment to the next one ; workaround to this is to place such config in a config.json which is loaded at startup ; see https://github.com/openzim/zimit-frontend/tree/main/ui running live at https://zimit.kiwix.org/ for a live example 
- @Popolechien will be extremely happy to see that background is now pure white (he always complained about this "grayish" color which was used so far but never make it into white)
- we should probably specify node version with `engines` key in `package.json` (even if not mandatory once in prod, it helps developers to know which Node version they should use
```
  "engines": {
    "node": ">=24 <25"
  },
```
- I'm very happy to see you've already switched to Vite 7 ... a bit concerned about potential side-effects since it's really brand new, but let's see, rollback will be easy
- small glitch on small screen: "Zimfarm" title is cut, looks like there is still some space reserved for the "Support Us" button
image
- Other than that, I'm not expert enough to spot potential other mistakes, but I did not spotted something else which strikes me ; plus many things are not yet wired, so hard to confirm they work as intended

In general, very pleased by the "design" in the sense that you've customized the strict minimum to have a "zimfarm" look why avoiding to reproduce all bootstrap specificity and prefer to stick to standard Vuetify look.

PR #1138, opened 2025-06-30
unnest name only fields

elfkuzco:
## Changes
- some fields are nested and contain only a `name`. move them to top level where they are defined
- include user scope details in API response as it is used by the frontend


benoit74 (approved):
LGTM, I let you merge

PR #1140, opened 2025-06-30
use legacy json options for dumping json objects to db

elfkuzco:
## Changes
- use legacy json options while serializing python objects to database. the default options converts datetime objects to isoformat instead of milliseconds like the `3.4` version of pymongo used in initial development. Rather than switching back to an old version of the library, we use the options and avoid the cost of migrating the database too.

- change fields to have `_name` suffix especially when they contain only strings. This helps reduce ambiguity on whether the field is a model or a string.


backend/src/zimfarm_backend/db/__init__.py
```diff
@@ -12,7 +12,7 @@
 # custom overload of bson deserializer to make naive datetime
 # this is needed to have objects from the DB with naive datetime properties
 # (otherwise the deserialization produces aware datetimes based on local TZ)
-def my_loads(s: str, *args: Any, **kwargs: Any) -> Any:
+def zimfarm_loads(s: str, *args: Any, **kwargs: Any) -> Any:
```
benoit74, line 15: Shouldn't we use LEGACY_JSON_OPTIONS here as well for consistency?

elfkuzco:
I figured the default was okay with deserializing the payload. Not sure how the library does it but I would imagine the default can work with legacy.

Should I switch it to legacy for consistency?

benoit74:
If legacy works as well, yes please do switch for consistency, it is just too puzzling otherwise, I'm pretty sure we will often have doubts about this inconsistency otherwise.

PR #1141, opened 2025-07-01
set up home pipeline table for home page

elfkuzco:
## Changes

- set up pipeline view for home page
- set up stores to fetch tasks and requested tasks from backend
- use cookies to keep track of selection

### DIFF
--- a/backend/pyproject.toml
+++ b/backend/pyproject.toml
@@ -10,7 +10,6 @@ authors = [
 keywords = ["zimfarm"]
 requires-python = ">=3.13,<3.14"
 description = "zimfarm Backend API"
-readme = "README.md"
 license = {text = "GPL-3.0-or-later"}
 classifiers = [
     "Programming Language :: Python :: 3",
--- a/backend/src/zimfarm_backend/main.py
+++ b/backend/src/zimfarm_backend/main.py
@@ -1,4 +1,7 @@
+import os
+
 from fastapi import APIRouter, FastAPI
+from fastapi.middleware.cors import CORSMiddleware
 
 from zimfarm_backend.routes import (
     auth,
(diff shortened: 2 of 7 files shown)


benoit74 (changes requested):
Quite a lot of moving parts ! Good job ! Still some major points to tackle before merging.

## Handling of auth

The concern I have is regarding the handling of auth tokens. Your code works fine in a context where you reload the whole app regularly. But most users won't do it. And here comes the struggles.

First, you store the headers to use in the tasks store. It means that if token expires while user is using the app, even if token is updated in auth store, task store will continue to use same (old) headers with same (old) token. And this will happen, especially since some people will keep a tab open with zimfarm UI, so app / stores / ... will never get reloaded.

Second, you do not have anything detecting that the token might have expired and automatically renewing it. To me this check must be done at every request (or at least every request which needs auth) but we must take care that multiple requests could be done at the same time, all detecting that auth token is expired ... and we want to block all requests, issue one single renew request and then let requests proceed normally.

Both points above tend to show that we need a central service / interceptor / whatever handling both aspects at once: inject proper (latest) auth token in every request and automatically renew it whenever needed, transparently. From store perspective, the store will just ask for a request to be made, not knowing anything about auth.

## Branching

The fact that we have two branches `frontend-upgrade` and `backend-upgrade` is finally not a good thing. It just makes it too hard for me to test PRs, because you need to have proper backend with proper frontend. Both are evolving together in fact. And you probably have the same kind of issues. I propose that we have one single target branch `big-upgrade` (probably not best name, feel free to propose something else) and continue with "feature branches" on top of this, with PRs, with changes for both frontend and/or backend (and probably worker as well, but this is the next thing). I don't mind to have changes for both frontend and backend in same PR, they are evolving together anyway.

WDYT?

## Docker / dev stack

To easily test your changes, I need to have a working setup in `dev/docker-compose.yml`. This is just way more convenient than having to start frontend and backend manually, not even being sure of which command I should issue. We need to update this to have proper `config.json` and other stuff for frontend to work as well. I can definitely help if needed.

elfkuzco:
Regarding the auth, I am not sure it works yet. I mean, there's no login in yet and it hasn't been tested. I wanted to set up a rough implementation for it as it is called on app mount. The more I wrote the pipeline view, I realized it was branching off towards every section of the application. I decided to stop and add TODO comments for things so I don't get distracted and redesign the whole app in one commit. I will take your considerations in while implementing the auth  store properly.

Regarding the backend and frontend being different and causing pains, you are correct and I will correct that shortly and set up the proper compose.

benoit74 (approved):
I've just added stuff for Docker / dev stack.

Let's merge this, this is mostly OK besides what you already know that has to be fixed.

@elfkuzco  I let you merge at your convenience.

PR #1143, opened 2025-07-02
Remove timestamp from running task

elfkuzco:
## Changes
- remove `timestamp` field from `RunningTask`. Changed the type from `dict[str, Any]` to `list[tuple[str, datetime.datetime]]` but Pyright kept complaining because instances of the model are defined from `Task` and in the db, `Task` timestamp is `dict[str, Any]`. Rather than adding a pyright warning to discard the issue, I removed the field as it wasn't even used.


benoit74 (changes requested):
My point was that both types (including DB type) are wrong. We should fix this since we are making a major API release, this is a good opportunity.

elfkuzco:
Applied a patch to convert timestamp for both `RequestedTask` and `Task` from `dict[str, Any]` to `list[tuple[str, datetime.datetime]]`

As a consequence, the function to get tasks/requested tasks from db can no longer sort by `requested` or `reserved` timestamps as there's no way to know their index in the list. So, I defaulted to sorting by `updated_at` column.

The SQL necessary for migrating existing data is:

```sql
UPDATE task
SET timestamp = (
  SELECT jsonb_agg(jsonb_build_array(key, value))
  FROM jsonb_each(timestamp)
)
WHERE jsonb_typeof(timestamp) = 'object';
```

The above query only applies it to the `Task` table. It should be modified to apply to the `RequestedTask` table too.

elfkuzco:
Also, here's the equivalent SQL for updating the config

```sql
UPDATE task
SET config = (
  config::jsonb
  - 'task_name'
  - 'flags'
  || jsonb_build_object(
       'offliner',
       jsonb_build_object(
         'offliner_id', config::jsonb->>'task_name'
       ) || (config::jsonb->'flags')
     )
)
WHERE config::jsonb ? 'task_name' AND config::jsonb ? 'flags';
```

backend/src/zimfarm_backend/db/models.py
```diff
@@ -161,6 +161,9 @@ class Task(Base):
     files: Mapped[dict[str, Any]]
     upload: Mapped[dict[str, Any]]
     original_schedule_name: Mapped[str]
+    timestamp: Mapped[list[tuple[str, Any]]] = mapped_column(
+        MutableList.as_mutable(JSONB), default_factory=list
```
benoit74: Do you really need to repeat it is a `MutableList.as_mutable(JSONB)` this since you've defined it in the base class?

backend/src/zimfarm_backend/db/models.py
```diff
@@ -257,6 +259,10 @@ class RequestedTask(Base):
     notification: Mapped[dict[str, Any]]
     original_schedule_name: Mapped[str]
 
+    timestamp: Mapped[list[tuple[str, Any]]] = mapped_column(
+        MutableList.as_mutable(JSONB), default_factory=list
```
benoit74: Same remark

benoit74 (approved):
LGTM, thank you!

PR #1145, opened 2025-07-02
set up sign in view and login

elfkuzco:
## Changes
- implement sign in view
- other stores will use the auth token from the store instead of from the cookie directly. 
- define helper function to translate api errors to  a list of strings (to be used by components for displaying errors)


frontend-ui/src/router/index.ts
```diff
@@ -1,26 +1,87 @@
...
+    },
+    {
+      path: '/sign-in',
+      name: 'sign-in',
+      component: SignInView,
+    },
+    {
+      path: '/change-password',
+      name: 'change-password',
+      component: TaskDetailView,
```
benoit74, line 31: I'm pretty sure you already now this, but this and all users paths will have to be fixed 🤣

benoit74 (approved):
See my (captain obvious) comment, other than that, just to be sure we are on the same page, I see that it still misses:
- reauthentication with refresh token (when access token / cookie has expired)
- something simpler to avoid repeating everywhere all this stuff related to authentication (this can clearly come in a second step)

LGTM, I let you merge

frontend-ui/src/router/index.ts
elfkuzco: yeah, router was complaining about missing paths. 😅

PR #1148, opened 2025-07-03
load last schedules for pipeline view page

elfkuzco:
# Changes
- implement remaining logic for pipeline view page by loading last schedule runs

### DIFF
--- a/frontend-ui/src/constants.ts
+++ b/frontend-ui/src/constants.ts
@@ -6,5 +6,9 @@ export default {
   TOKEN_COOKIE_NAME: 'auth',
   TOKEN_COOKIE_EXPIRY: '6m', // 6 months
   COOKIE_LIFETIME_EXPIRY: '10y', // 10 years
-  TASKS_LOAD_SCHEDULES_CHUNK_SIZE: 5,
+  TASKS_LOAD_SCHEDULES_DELAY: 100,
+  // Notification constants
+  NOTIFICATION_DEFAULT_DURATION: 5000, // 5 seconds
+  NOTIFICATION_ERROR_DURATION: 8000, // 8 seconds for errors
+  NOTIFICATION_SUCCESS_DURATION: 3000, // 3 seconds for success
 }
--- a/frontend-ui/src/views/PipelineView.vue
+++ b/frontend-ui/src/views/PipelineView.vue
@@ -2,7 +2,7 @@
 
 <template>
   <TasksListTab
-    :filter="filter"
+    :filter="currentFilter"
     :filterOptions="filterOptions"
     @filter-changed="handleFilterChange"
   />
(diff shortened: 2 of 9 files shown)


frontend-ui/src/views/PipelineView.vue
```diff
@@ -167,7 +171,32 @@ async function loadData(limit: number, skip: number, filter?: string) {
 }
 
 async function loadLastRuns() {
-  // TODO: Load last runs
+  lastRunsLoaded.value = false
+  const schedule_names = Array.from(new Set(tasks.value.map((task) => task.schedule_name)))
+  loadingStore.startLoading('Loading last runs...')
+
+  const chunkSize = constants.TASKS_LOAD_SCHEDULES_CHUNK_SIZE
```
benoit74: I'm sorry but I don't get this logic purpose:
- why making batches of 5 requests? If the goal is to avoid overwhelming the server, then I would rather pause between every requests, maybe only 100ms or something like that
- why awaiting getDelay every time? To me it looks like we always wait even if we have only one chunck

Or do I miss something obvious?

elfkuzco: i was just trying to recreate the logic from here https://github.com/openzim/zimfarm/blob/0feac1c43a48c12079a7123f3fceb793f233d06f/dispatcher/frontend-ui/src/components/PipelineTable.vue#L189-L198


I didn't fully understand it but now that you mention, I could rewrite it this way instead.

benoit74: Indeed, looks like logic is already crap. Please fix this: make a small pause between each request for a schedule so that we do not overwhelm the backend. And if feasible (not sure how this is), update the UI as the results arrive (instead of updating them all at once when we have received all results).

frontend-ui/src/utils/httpRequest.ts
```diff
@@ -38,3 +38,8 @@ export default function httpRequest(
 
   return http
 }
+
+export function getDelay(milliseconds: number) {
```
benoit74: This is not used anymore, do we want to keep it?

frontend-ui/src/views/PipelineView.vue
```diff
@@ -167,8 +170,29 @@ async function loadData(limit: number, skip: number, filter?: string) {
 }
 
 async function loadLastRuns() {
-  // TODO: Load last runs
+  const schedule_names = Array.from(new Set(tasks.value.map((task) => task.schedule_name)))
+  loadingStore.startLoading('Loading last runs...')
```
benoit74: I just realize that this means we cannot "use" the UI until all last runs are loading ... this is not optimal (especially if there are 200 of them). We can postpone this to a later enhancement (meaning we can merge the PR without it), but I would appreciate if we could make the UI immediately available, in many cases we do not mind about last runs.

frontend-ui/src/views/PipelineView.vue
```diff
@@ -167,8 +170,29 @@ async function loadData(limit: number, skip: number, filter?: string) {
...
+    if (schedule_name) {
+      try {
+        const result = await scheduleStore.fetchSchedule(schedule_name)
+        if (result && result.most_recent_task) {
+          schedulesLastRuns.value[schedule_name] = result.most_recent_task
+        }
+        // Add a 100ms pause between requests to avoid overwhelming the backend
+        await new Promise(resolve => setTimeout(resolve, 100))
+      } catch (error) {
+        console.error(`Failed to fetch schedule ${schedule_name}:`, error)
```
benoit74: Would be nice to display an error label (without details) in place of the last run information in the UI, so that it is clearer that UI attempted to load this but failed. Could be done later again.

frontend-ui/src/constants.ts
```diff
@@ -7,4 +7,5 @@ export default {
   TOKEN_COOKIE_EXPIRY: '6m', // 6 months
   COOKIE_LIFETIME_EXPIRY: '10y', // 10 years
   TASKS_LOAD_SCHEDULES_CHUNK_SIZE: 5,
```
benoit74: `TASKS_LOAD_SCHEDULES_CHUNK_SIZE` is not used anymore unless I'm mistaken

frontend-ui/src/constants.ts
```diff
@@ -7,4 +7,5 @@ export default {
...
+  TASKS_LOAD_SCHEDULES_DELAY: 100,
```
benoit74, line 9: This should be used instead of the hard-coded value

benoit74 (changes requested):
Few remarks / suggestions

frontend-ui/src/views/PipelineView.vue
elfkuzco: Okay. Adding this in the next PR.

elfkuzco: Added in current PR instead  so it would be reusable

frontend-ui/src/views/PipelineView.vue
```diff
@@ -167,8 +172,33 @@ async function loadData(limit: number, skip: number, filter?: string) {
...
+        }
+      }
+      // Add a 100ms pause between requests to avoid overwhelming the backend
+      await new Promise(resolve => setTimeout(resolve, constants.TASKS_LOAD_SCHEDULES_DELAY))
+
+    }
+  }
+
+
+  loadingStore.stopLoading()
```
benoit74: loadingStore was already not loading anymore unless I'm mistaken, I think we do not need this.

benoit74 (approved):
One small inline comment + would it be possible to stop background loading of last schedules once we move to another view? It looks like it keeps loading last runs even if we've moved somewhere else (e.g. a schedule or task details) ... which is kinda useless.

I let you decide if you wanna merge this and fix this later, and fix it now.

elfkuzco:
> One small inline comment + would it be possible to stop background loading of last schedules once we move to another view? It looks like it keeps loading last runs even if we've moved somewhere else (e.g. a schedule or task details) ... which is kinda useless.
> 
> I let you decide if you wanna merge this and fix this later, and fix it now.

This is weird. I tried running it on my end and couldn't reproduce. I delete the interval anytime the component is unmounted.

benoit74:
> This is weird. I tried running it on my end and couldn't reproduce. I delete the interval anytime the component is unmounted.

I don't reproduce anymore as well. Either I had something in cache somewhere, or there is a subtle edge case. Let's keep it as-is for the time being and merge.

I let you squash and merge

PR #1149, opened 2025-07-03
Update offliner models

elfkuzco:
# Rationale
Tested the frontend with the schedules store using the database dump sent to me and it became clear there were some mistakes and omissions during the initial model draft. Tested 1k schedules so far and based on that I have made these modificiations.

## Changes
- use the same alias for the discriminator field `offliner_id` as pydantic disallows different aliases for a discriminator. the `CamelModel` and `DashModel` were generating different aliases
- update fields in models to use the proper aliases
- manually type out the aliases for the zimit schema as composing a class from the two alias generators didn't prove successful. in the end, only one alias generator was being used causing validation errors on existing data


backend/src/zimfarm_backend/common/schemas/offliners/zimit.py
```diff
@@ -156,6 +156,65 @@ class ZimitDevice(StrEnum):
 
 
 class ZimitFlagsCamelSchema(CamelModel):
```
benoit74: Rename to `ZimitFlagsFullSchema` ?

elfkuzco:
> Shouldn't we prefer `offlinerId` (camelCase) since this is the "default" for JSON? Any reason not to do so? (I might miss it)

I tried this out as follows

```py
from typing import Literal

from pydantic import BaseModel, Field, ValidationError


class Cat(BaseModel):
    pet_type: Literal['cat'] = Field(alias="petType")
    meows: int


class Dog(BaseModel):
    pet_type: Literal['dog'] = Field(alias="petType")
    barks: float


class Lizard(BaseModel):
    pet_type: Literal['reptile', 'lizard'] = Field(alias="petType")
    scales: bool


class Model(BaseModel):
    pet: Cat | Dog | Lizard = Field(discriminator='pet_type')
    n: int
```

NOTE: Writing `discriminator='petType` will fail with a validationError. I don't know so much about discriminators. But it seems the field you set as the discriminator is the variable name, not the serialization alias. So, this is the only way this code will work. Now, when I do
```py
v = Model(pet={'pet_type': 'dog', 'barks': 3.14}, n=1)
print(v.model_dump(mode="json"))
# the output is "{'pet': {'pet_type': 'dog', 'barks': 3.14}, 'n': 1}"
```
the `alias` is not respected. I tried `serialization_alias` and it isn't respected either. The only way we can get it to be `offlinerId` is if we set the variable itself to `offlinerId`. But then again, if this conforms to "JSON", it doesn't conform to the Python way. Which convention do we stick to?

benoit74:
To be, this is because you need the `populate_by_name` property to be set in model config. 

```python
from typing import Literal

from pydantic import BaseModel, ConfigDict, Field
from pydantic.alias_generators import to_camel

class CamelModel(BaseModel):
    model_config = ConfigDict(
        alias_generator=to_camel,
        populate_by_name=True
    )

class Cat(CamelModel):
    pet_type: Literal['cat']
    meows: int


class Dog(CamelModel):
    pet_type: Literal['dog']
    barks: float


class Lizard(CamelModel):
    pet_type: Literal['reptile', 'lizard']
    scales: bool


class Model(CamelModel):
    pet: Cat | Dog | Lizard = Field(discriminator='pet_type')
    n: int

v = Model(pet={'pet_type': 'dog', 'barks': 3.14}, n=1)
print(v.model_dump(mode="json", by_alias=True))
# {'pet': {'petType': 'dog', 'barks': 3.14}, 'n': 1}

v2 = Model(pet=Dog(pet_type='dog', barks= 2), n=1)
print(v2.model_dump(mode="json", by_alias=True))
# {'pet': {'petType': 'dog', 'barks': 2.0}, 'n': 1}

v3 = Model.model_validate_json('{"pet": {"petType": "dog", "barks": 2.0}, "n": 1}')
print(v3)
# pet=Dog(pet_type='dog', barks=2.0) n=1

# it's even permissive:
v4 = Model.model_validate_json('{"pet": {"pet_type": "dog", "barks": 2.0}, "n": 1}')
print(v4)
# pet=Dog(pet_type='dog', barks=2.0) n=1
```

Not sure about the exact impact it could have on the rest of the stack ...

Btw, why did you used `from pydantic.v1.utils import to_lower_camel` instead of `from pydantic.alias_generators import to_camel` (besides because you didn't found it, happens to me all the time, I even resorted to add pyhump dep somewhere ^^)

elfkuzco:
This mostly works because you are inheriting from the `BaseModel` or a uniform model. We use two models. Without this, we get names like `offliner-id`, `offlinerId`, I tried your example with one class inheriting from `DashModel` and I get this

```py
pydantic.errors.PydanticUserError: Aliases for discriminator 'pet_type' must be the same (got pet-type, petType)

For further information visit https://errors.pydantic.dev/2.11/u/discriminator-alias
```

elfkuzco:
Think I read somewhere about to_camel not being what we expect but it seems it has changed. I guess they have updated it.

elfkuzco:
For our case, we have two alias generators. I have to explicitly add alias to every offliner or I get the error message

```
pydantic.errors.PydanticUserError: Aliases for discriminator 'pet_type' must be the same (got pet-type, petType)

For further information visit https://errors.pydantic.dev/2.11/u/discriminator-alias
```


Even if after adding the alias to the fields (because there are two alias generators), when I call `.model_dump(mode="json")`, it still dumps as `pet_type`, not `petType`. So, the purpose is kinda defeated.

elfkuzco:
My understanding is that is the variable name for the fields that is the `discriminator`

benoit74 (approved):
LGTM. Thank for all these efforts ^^

PR #1150, opened 2025-07-04
generate parseable definition for offliner schema

elfkuzco:
## Changes

- generate definiton for offliner schema

### DIFF
--- a/backend/src/zimfarm_backend/common/schemas/fields.py
+++ b/backend/src/zimfarm_backend/common/schemas/fields.py
@@ -21,7 +21,7 @@ def OptionalField(**kwargs: Any) -> Any:  # noqa N802
 
 NoNullCharString = Annotated[str, AfterValidator(no_null_char)]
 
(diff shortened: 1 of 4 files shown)


benoit74 (changes requested):
Great work! I like it, looks like I expected it to be way more awful than it finally is.

Please add tests (or enrich existing ones) so that we have close to 100% coverage of this whole part of the codebase. This looks like complex code and maybe a bit fragile. Most offliners have pretty stable definitions so you can simply generate all their `schema_to_flags` and assert some properties (one per "base type" + secret + ... typically)

elfkuzco:
> Great work! I like it, looks like I expected it to be way more awful than it finally is.
> 
> Please add tests (or enrich existing ones) so that we have close to 100% coverage of this whole part of the codebase. This looks like complex code and maybe a bit fragile. Most offliners have pretty stable definitions so you can simply generate all their `schema_to_flags` and assert some properties (one per "base type" + secret + ... typically)

Seems like I omitted to add the test file.

PR #1152, opened 2025-07-07
remove routing logic from navbar

elfkuzco:
## Changes

- remove route handling logic from navbar and delegate to vuetify


benoit74 (approved):
LGTM, I let you merge

PR #1155, opened 2025-07-10
set up users view

elfkuzco:
## Changes
- implement view to list users


elfkuzco:
Failing tests have been addressed in #1156

PR #1156, opened 2025-07-10
perfrom crud on user obtained from username

elfkuzco:
## Changes
- fix bug where routes that took a `username` path param were acting on the current user instead of the user identified by the username
- add support to load keys in openSSH format. existing implementation only worked for keys in PEM format
- use common `getnow()` for datetime objects

### DIFF
--- a/backend/src/zimfarm_backend/utils/token.py
+++ b/backend/src/zimfarm_backend/utils/token.py
@@ -8,7 +8,9 @@ from cryptography.exceptions import InvalidSignature, UnsupportedAlgorithm
 from cryptography.hazmat.primitives import hashes, serialization
 from cryptography.hazmat.primitives.asymmetric import padding
 from cryptography.hazmat.primitives.asymmetric.rsa import RSAPrivateKey, RSAPublicKey
+from cryptography.hazmat.primitives.asymmetric.types import PublicKeyTypes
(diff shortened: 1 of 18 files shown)


backend/src/zimfarm_backend/utils/token.py
```diff
@@ -99,7 +103,17 @@ def get_public_key_fingerprint(public_key: RSAPublicKey) -> str:
 
 
 def load_rsa_public_key(key: str) -> RSAPublicKey:
```
benoit74, line 107: Looks like we do not check that key passed is RSA, while we do expect this. Should we add an explicit check?

benoit74 (changes requested):

backend/src/zimfarm_backend/utils/token.py
elfkuzco: So, throughout the development, I had been thinking that the keys were passed in the PEM format, not knowing that it was the openSSH format. It was during the design of the frontend that I realized the keys were `.pub` files which were often in openSSH format. By adding it as the first format to try, the system can allow users to upload both formats.

elfkuzco: yeah. added an explicit `isinstance` check as the type returned is a union type

PR #1157, opened 2025-07-10
set up user detail view

elfkuzco:
## Changes

- implement view to show, edit and delete user
- add confirmation dialog for buttons to delete items


benoit74 (approved):
Please fix padding around the textbox below:

image

And then merge, LGTM

PR #1158, opened 2025-07-10
add support view

elfkuzco:
## Changes
- redesign support page


benoit74 (changes requested):
I feel like this page design is odd.

image

While original design was too minimalist and improving it is definitely a plus, I'm puzzled by:
- the horizontal whitespaces around every links are weird
- the paragraph on donating money should be given more importance than the "Donate a worker" section; probably just invese both styles
- the fact that "Donate a worker" details is spread on two paragraphs with a big space in between is odd
- the "Donate a worker" (or its equivalent once you've moved to money) text must be vertically centered on its icon

elfkuzco:
removed the excess whitespaces. turns out switching to a new line for the closing anchor tag was adding the extra space. reverted back styling for `Donate a Worker` as in the first comment. Didn't like the green after much. Review and let me know if it's okay.

PR #1160, opened 2025-07-13
Standardize schedule validation errors

elfkuzco:
## Changes
- raise fastapi RequestValidationErrors from pydantic ValidationErrors. This is then handled by the app exception handler which formats it in a way that can be used by the frontend.
- make kolibri fields optional (oversight from the initial PR that created the model schema)


benoit74 (approved):
LGTM

PR #1161, opened 2025-07-13
set up schedule detail view

elfkuzco:
## Changes
- set up schedule detail view
- set up components like `ScheduleEditor`, `ScheduleActionButton`, `FlagsList`, `CloneSchedule` to inherit props from view
- remove previous logic to hide secrets from command. This is hidden  by the backend which uses `show_secrets` to hide sensitive keys from commands
- load offliner definitions from backend
- add remaining methods to stores to perform crud on backend for requestedtasks, schedules and tasks.


frontend-ui/src/components/FlagsList.vue
```diff
@@ -1,35 +1,69 @@
...
-      <tr v-for="(value, name) in flags" :key="name">
-        <td><code>{{ name }}</code></td>
-        <td class="value" v-if="is_protected_key(name)" v-tooltip="'Actual content hidden'">{{ value }}</td>
-        <td class="value" v-else>{{ value }}</td>
+      <tr v-for="(value, name) in filteredOffliner" :key="name">
+        <td><code class="text-pink-accent-2">{{ name }}</code></td>
+        <td>
+          <span
+            v-if="isProtectedKey(name)"
+            v-tooltip="'Actual content hidden'"
```
benoit74: I don't see this tooltip in the UI, do I miss something?

benoit74 (changes requested):
This is huge! Thank you very much!

I still found some rough edges, but still impressive work done. Thank you, and well done!

Note that given PR size I probably missed few things, but we will polish them in the aftermath.

Logic to compute the buttons to request/unrequest/prioritize is wrong somewhere, I always get the "unrequest/prioritize" buttons

image

Renaming a recipe (change recipe name) is causing a bug in UI (everything seems fine on the backend): a notification appears saying that "old_recipe_name" is not found

Hints below combobox and text box of content settings + task settings are appearing only once the get focus. The hint should always be displayed just like for "Status" and "Monitoring" checkbox.

Logic to detect if one changed the configuration is not working well. As soon as ones edit some flag, the "Update offliner details" button stays "on", even if user goes back to original config. It does work well for content settings and task settings.

List of image tags is not populated; changing "Image Name" does populate the list. Looks like you missed to call this on load. Have a bad "Image Name" causes a "Network Error" to be displayed, while we could be more specific (but it is probably a backend error, it throws a 500 on invalid image name while it should be a 404 since ghcr.io responded with a 404 ... it works like this on live version at least)

Space between textboxes of flags on hint text is way too big. I would prefer something like below (from zimit.kiwix.org advanced options):

image

Add bold to flag names so that they are easier to read. Maybe a bit bigger would help as well. And the "star" for mandatory flags is just way too small. 

Validation errors of flags (e.g. a missing mandatory value) is correctly disabling the "Update offliner details" button but there is no other visual clue on what is wrong. We should have a hint next to the offending flag. This is how I did it on zimit.kiwix.org (once validation fails):

image

When logged-in with appropriate credentials, secret flags are still not shown and it causes them to be removed/hidden when editing the recipe. I think we should always pass `hideSecrets: False` from this UI, and let backend decide if they have to be hidden or not. This flag is more here so that by default flags are hidden for consumers unless consumer know he wants the secret flags. For instance, zimit.kiwix.org knows it doesn't want secret flags to appear under any circumstance.

elfkuzco:
Turns  out the secrets were being hidden by the `expanede_config` function. Added a flag to show it by default as it almost always used with default of `True`. Then, based on request params or role of caller, it can be set to `False`. This would close #1165

elfkuzco:
Okay.

PR #1163, opened 2025-07-14
add view for task details

elfkuzco:
## Changes

- add view for routes
- remove unused `stats` route


elfkuzco:
After #1143 has been implemented, I will rewrite this view to use the new schema definition. Doing them both concurrently was taking quite some time to get right. Will create an issue for it.

frontend-ui/src/constants.ts
```diff
@@ -9,6 +9,7 @@ export default {
   COOKIE_LIFETIME_EXPIRY: '10y', // 10 years
   TASKS_LOAD_SCHEDULES_DELAY: 100,
   MAX_SCHEDULES_IN_SELECTION_REQUEST: 200,
+  KIWIX_DOWNLOAD_URL: 'https://download.kiwix.org/zim',
```
benoit74: Rename to `ZIM_DOWNLOAD_URL` + move to `config.json` so that we can customize it per deployment (value is different between farm.openzim.org and farm.zimit.kiwix.org typically)

frontend-ui/src/stores/tasks.ts
```diff
@@ -62,11 +63,29 @@ export const useTasksStore = defineStore('tasks', () => {
     }
   }
 
+  const fetchTask = async (id: string, hideSecrets: boolean = true) => {
+    const service = await authStore.getApiService('tasks')
+    try {
+      const response = await service.get<null, Task>(`/${id}`, { params: { hide_secrets: hideSecrets } })
```
benoit74, line 69: Passing `hide_secrets: False` and letting backend decide is probably easier / more accurate, it will save us from some nasty bugs where frontend and backend do not agree about wether to hide secrets or not.

benoit74 (changes requested):
Why `command` is not the same as in schedule view? (with the purple / pink color + without the bug hiding horizontal lines).

image

+ see inline comments

benoit74 (approved):
LGTM

PR #1168, opened 2025-07-17
update timestamp schema in frontend

elfkuzco:
## Changes
- update task schema timestamp from `Record<string, string>` to `[string, string][]` to reflect changes in db


elfkuzco:
This depends on #1143

benoit74:
Looks like there is many commits which are not really related to this PR. Something strange is happening, I let you clean the situation and ask for review again once this is sorted out.

elfkuzco:
> Looks like there is many commits which are not really related to this PR. Something strange is happening, I let you clean the situation and ask for review again once this is sorted out.

Rebased.

PR #1175, opened 2025-07-18
show confirm dialog before removing requested task

elfkuzco:
## Changes
- show confirmation dialog before deleting requested task
- remove unnecessary files


frontend-ui/src/components/RemoveRequestedTaskButton.vue
```diff
@@ -1,26 +1,44 @@
...
+      @click="showConfirmDialog"
+    >
+      <v-icon size="small" class="mr-1">mdi-delete</v-icon>
+      {{ isRemoving ? 'Removing...' : 'Remove' }}
+    </v-btn>
+
+    <ConfirmDialog
+      v-model="showDialog"
+      title="Remove Requested Task"
+      message="Are you sure you want to remove this requested task? This action cannot be undone."
```
benoit74: `Are you sure you want to remove this requested task?` is enough for the message, `This action cannot be undone.` is not really true, user just has to request the task again.

PR #1181, opened 2025-07-22
refactor common worker logic

elfkuzco:
## Changes

- refactor common logic for worker


worker/pyproject.toml
```diff
@@ -0,0 +1,227 @@
...
+[project.urls]
+Homepage = "https://github.com/openzim/zimfarm/worker"
+Donate = "https://www.kiwix.org/en/support-us/"
+
+[tool.hatch.build]
+exclude = [
+  "/.github",
+]
+
+[tool.hatch.build.targets.wheel]
```
benoit74, line 66: Do we really need this `wheel` target since we are not publishing to Pypi anyway?

worker/pyproject.toml
```diff
@@ -0,0 +1,227 @@
...
+packages = ["src/manager", "src/common"]
+
+[tool.hatch.metadata]
+allow-direct-references = true
+
+[tool.hatch.envs.default]
+features = ["dev"]
+
+[tool.hatch.envs.manager]
+dependences = [
```
benoit74: Typo

worker/pyproject.toml
```diff
@@ -0,0 +1,227 @@
...
+    "requests == 2.32.3",
+]
+
+[tool.hatch.envs.task]
+dependences = [
```
benoit74: Typo

benoit74: + why do we need `SQLAlchemy` and don't need `requests`? Kinda weird for the task manager (but maybe this is for next PR ^^)

worker/pyproject.toml
```diff
@@ -0,0 +1,227 @@
...
+    "SQLAlchemy == 2.0.41",
+]
+
+[tool.hatch.envs.test]
+features = ["scripts", "test"]
+
+[tool.hatch.envs.test.scripts]
```
benoit74, line 94: We need to add `tasks.py` for this to work.

worker/src/common/cryptography.py
```diff
@@ -0,0 +1,67 @@
...
+
+@dataclass
+class AuthMessage:
+    """An authentication message for a worker"""
+
+    body: str
+    signature: str
+
+
+def load_private_key_from_path(private_key_fpath: Path) -> RSAPrivateKey:
```
benoit74, line 24: Is it too much work to also support EDCSA? I thought it was ready on backend side, but I might have misunderstood you. Not a big game changer, we can postpone this to later.

benoit74 (changes requested):
Thank you, lots of work again, much more than anticipated as always 🐰

worker/pyproject.toml
elfkuzco: If I don't set it and do `hatch shell`, it fails because with this:
```
ValueError: Unable to determine which files to ship inside the wheel using the following heuristics: https://hatch.pypa.io/latest/plugins/builder/wheel/#default-file-selection

      The most likely cause of this is that there is no directory that matches the name of your project (zimfarm_worker).

      At least one file selection option must be defined in the tool.hatch.build.targets.wheel table, see: https://hatch.pypa.io/latest/config/build/

      As an example, if you intend to ship a directory named foo that resides within a src directory located at the root of your project, you can define the following:

      [tool.hatch.build.targets.wheel]
     ```

benoit74: OK, let's keep it then

worker/src/common/cryptography.py
elfkuzco: It's not too much work. I can do it in a new PR.

elfkuzco: Created #1186 to track this.

PR #1184, opened 2025-07-22
generate v2 docs from server

### LINKED ISSUE #1177: Zimfarm Upgrade: Host API v1 and v2 swagger UI side-by-side
Would it be feasible to create a small static website with Swagger UI of both v1 and v2 side-by-side? No need to build this in an automated manner, this is mostly only a dev tool we could host somewhere during the migration.

I think it would help a lot to:
- easily have v2 doc available before the migration (to prepare)
- easily have v1 doc after the migration (to check if some bug occurs)

For v1 it is probably straightforward, just need to take the `docs` folder and have a website serve it

For v2 ... I don't know 🤣 

Could be a quick and dirty solution I publish manually somewhere, it is not meant to stay there for long.

elfkuzco:
## Changes
- add description to api
- change version prefix from `/api/v2` to `/v2`
- set up script to generate openapi docs from fastapi.

To generate the script:
- start your fastapi application
- run `python generate_v2_docs.py`
- the docs would be saved in `docs_v2` directory.
- `cd docs_v2/swagger-ui/dist`
- start a python server with `python -m http.server 5000` and view the docs at `localhost:5000`

This closes #1177


benoit74 (approved):
Please move `generate_v2_docs.py` into `maint_scripts` folder so that it is clear this is not production code.

After that, good to merge.

kelson42:
@benoit74 Once merged, we should probably trigger WP1 to merge.

benoit74:
@kelson42 what's the relation with WP1? at some point we will have to update WP1 and zimit-frontend to use v2 API, but I've not yet opened the issue because I was waiting for proper documentation to help with the migration

Issue #1131, opened 2025-06-23
Searching for `c++` in recipes causes a 500 error


benoit74:
See https://farm.openzim.org/recipes?name=c%2B%2B

elfkuzco:
Fixed as part of the zimfarm upgrade

kelson42:
@elfkuzco No PR in particular to link to this issue?

elfkuzco:
> [@elfkuzco](https://github.com/elfkuzco) No PR in particular to link to this issue?

Um, there's no particular PR that addresses. Just realized it came as a consequence of the updates made so far on the `zimfarm-upgrade` branch.

elfkuzco:
@benoit74 , can you take a look at this? Seems to have been fixed as a consequence of the `zimfarm-upgrade`

benoit74:
I confirm, let's close this issue then. Too hard to find precisely which PR solved the issue, never mind.

benoit74:
Nota: this is not yet deployed in production, only fixed in `zimfarm-upgrade` branch.

PR #1188, opened 2025-07-22
Sotoki recipe name must start with domain name, not only contain it

benoit74:
When SE dumps is `ai.stackexchange.com`, the watcher is requesting https://farm.openzim.org/recipes/ai.stackexchange.com_en (correct) and https://farm.openzim.org/recipes/genai.stackexchange.com_en (mistake)

This is fixing this situation.

### DIFF
--- a/watcher/watcher.py
+++ b/watcher/watcher.py
@@ -331,8 +331,11 @@ class WatcherRunner:
             logger.error(f"Can't get `{domain}` schedules from zimfarm")
             return True
 
+        # filter by recipe name exactly starting with domain name, since API call
+        # search for the domain "anywhere" in the recipe name
         return [
-            recipe["name"] for recipe in payload.get("items", []) if recipe["enabled"]
+            recipe["name"] for recipe in payload.get("items", [])
+            if recipe["enabled"] and recipe["name"].startswith(f"{domain}_")
         ]
 
     def blocked_by_zimfarm(self, domain):


benoit74:
@elfkuzco happy to have a small feedback, just confirming I'm not doing something silly with this API request (I do get you probably barely understand yet the rest of the logic)

watcher/watcher.py
```diff
@@ -325,7 +325,7 @@ def get_recipes_for(self, domain):
         success, status, payload = self.query_api(
             "GET",
             "/schedules/",
-            params={"category": ["stack_exchange"], "name": f"{domain}_.+"},
+            params={"category": ["stack_exchange"], "name": f"^{domain}_.+"},
```
elfkuzco: I don't think the PostgreSQL `ilike` comparsion function used in the DB query works with regexes.

elfkuzco: Perhaps just use the `domain_` as it is without the anchors. The string is sandwiched between `%` in an `ilike` expression. Which is probably what you want. Ran the current query and nothing returns. But without the `^` and `.+`, you get results back.

elfkuzco (changes requested):

watcher/watcher.py
elfkuzco: That explains  the issue with `genai.stackexchange.com_en`. The query should probably be revised at the dB level

elfkuzco: Turns out the old code did use `regexp_match` to filter while I use `ilike`. The use of `regexp_match` explains #1131 where special characters in searches fail. Should I revert it?

benoit74: I feel like it would be simpler to filter the results we get with `f"{domain}_.+"` to get only what we want ; I feel like using `ilike` is still the right choice to avoid many too subtle issues involved with `regexp_match`. I will fix this then. It is just that when I saw `f"^{domain}_.+"` worked I got very lazy and satisfied 🤣

elfkuzco (approved):
LGTM

PR #1192, opened 2025-07-23
define __init__ for workers

elfkuzco:
## Changes
- remove paramiko as dependency and compute fingerprint directly
- define `__init__` for workers so it's clear how instance attributes are set
- change annotation for webapi uri's from urllib ParseResult to str as there are lots of calls to `get_uri()` and there are only a couple of places where the attributes of the webapi (like netloc) are used.
- make dataclasses to be `kw_only` so it's obvious how they are constructed.
- move `can_stream_logs` from `BaseWorker` to `TaskWorker` as it relies on attributes set by the `TaskWorker` and not really the `WorkerManager`

### DIFF
--- a/worker/pyproject.toml
+++ b/worker/pyproject.toml
@@ -18,7 +18,6 @@ classifiers = [
     "License :: OSI Approved :: GNU General Public License v3 or later (GPLv3+)",
 ]
 dependencies = [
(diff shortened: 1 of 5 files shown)


benoit74 (approved):
LGTM

PR #1196, opened 2025-07-24
Schedule editor update btn

### LINKED ISSUE #1171: UI update: impossible to submit / reset form when field is emptied
https://github.com/user-attachments/assets/debc0270-a77b-4759-9fd4-5ebbaf795bda

elfkuzco:
## Changes
- allow submit/reset button to be active when field is emptied. This closes #1171

### DIFF
--- a/frontend-ui/src/components/ScheduleEditor.vue
+++ b/frontend-ui/src/components/ScheduleEditor.vue
@@ -6,7 +6,8 @@
           :disabled="!hasChanges"
           type="submit"
           :color="hasChanges ? 'primary' : 'secondary'"
-          variant="elevated">
+          variant="elevated"
+        >
           Update Offliner details
         </v-btn>
         <v-btn
@@ -15,7 +16,8 @@
           :color="hasChanges ? 'dark' : 'secondary'"
           variant="outlined"
           class="ml-2"
-          @click="handleReset">
+          @click="handleReset"
+        >
           Reset
         </v-btn>
       </v-col>
@@ -32,7 +34,8 @@
           href="https://github.com/openzim/zimfarm/wiki/Recipe-configuration-%E2%80%90-Content-settings"
           target="_blank"
           color="primary"
-          variant="outlined">
+          variant="outlined"
+        >
           Help
         </v-btn>
       </v-col>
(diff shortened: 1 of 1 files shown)


benoit74 (changes requested):
This indeed fixes the bug in the UI, but then DB is not updated correctly by the backend. Looks like logic to merge flags in PATCH is wrong somewhere.

Can we fix this is same issue, so that from user perspective removing a flag works as expected? Thank you

frontend-ui/src/components/ScheduleEditor.vue
```diff
@@ -858,19 +893,27 @@ const cleanFlagsPayload = (flags: Record<string, any>) => {
     for (const key in obj) {
       if (typeof obj[key] === 'object' && obj[key] !== null) {
         recursivelyCleanup(obj[key])
-      } else if (!Array.isArray(obj) && (obj[key] === '' || obj[key] === undefined || obj[key] === null)) {
-        delete obj[key]
+      } else if (
```
elfkuzco, line 896: applied a fix that sets empty strings to null if they were explicitly set by the caller. this is because for optional fields on the backend, setting it to empty would fail because of the `NotEmptyString` underlying type. So, given the user wants to mark it as empty, we set it to null before forwarding the payload. Think my editor ended up reformatting the file and messed up the actuall diff but there it is

frontend-ui/src/components/ScheduleEditor.vue
```diff
@@ -858,19 +893,27 @@ const cleanFlagsPayload = (flags: Record<string, any>) => {
...
+        (obj[key] === '' || obj[key] === undefined || obj[key] === null)
+      ) {
+        // if an object is set to empty string, we update it to null as it should
+        // be on the backend since fields are either set or not set and fields that
+        // are set shouldn't be empty
+        if (obj[key] == '') {
+          obj[key] = null
+        } else {
+          delete obj[key]
+        }
```
elfkuzco, line 907: applied a fix that sets empty strings to null if they were explicitly set by the caller. this is because for optional fields on the backend, setting it to empty would fail because of the `NotEmptyString` underlying type. So, given the user wants to mark it as empty, we set it to null before forwarding the payload. Think my editor ended up reformatting the file and messed up the actuall diff but there it is

PR #1199, opened 2025-07-25
Worker fixes

elfkuzco:
## Changes
- add healthcheck endpoint for backend and curl to Dockerfile. 
- rewrite query for update_schedule_duration to use the updated timestamp field for Task (now `list[tuple[str, Any]]`
- bind task container to same network as manager if `ENVIRONMENT` is set to `development`


backend/Dockerfile
```diff
@@ -20,6 +20,10 @@ ENV MAILGUN_FROM Zimfarm <info@farm.openzim.org>
 # ENV suffixed with event: `GLOBAL_NOTIFICATION_started` or `GLOBAL_NOTIFICATION_ended`
 # ENV GLOBAL_NOTIFICATION_ended slack,#zimfarm-events,@rgaudin|mailgun,reg@kiwix.org
 
+# curl needed for healthcheck
```
benoit74, line 23: Since we are adding `curl` in the container, please add `HEALTHCHECK` directive in the container as well.

This is handy to have default healthcheck defined in the image rather than having to define it in production. It can always be disabled / overriden if needed.

dev/docker-compose.yml
```diff
@@ -42,6 +56,11 @@ services:
     depends_on:
       postgresdb:
         condition: service_healthy
+    healthcheck:
```
benoit74: `healthcheck` moved in `Dockefile`, see other comment.

benoit74 (changes requested):
See inline comments.

I've also opened https://github.com/openzim/zimfarm/issues/1200 and https://github.com/openzim/zimfarm/issues/1201 which are probably unrelated to this PR.

It looks like there is also an issue with the `remove_container` Docker operation. The task manager often uses this operation "just in case". So it is expected to fail. However, currently the `retry` causes it to retry 5 times, which makes the whole system very slow.  I don't remember how it worked before. Maybe `remove_container` was the only Docker operation without retry logic since it is expected to fail in many occasions?

dev/docker-compose.yml
```diff
@@ -3,7 +3,7 @@ services:
     image: postgres:17.3-bookworm
     container_name: zf_postgresdb
     ports:
-      - 127.0.0.1:5432:5432
+      - 127.0.0.1:2345:5432
```
benoit74, line 6: Please revert this change

dev/docker-compose.yml
```diff
@@ -26,12 +26,26 @@ services:
...
+      - --reload
+      - --reload-dir
+      - /usr/local/lib/python3.13/site-packages/zimfarm_backend
     ports:
       - 127.0.0.1:8000:80
     environment:
       DEBUG: true
       JWT_SECRET: DH8kSxcflUVfNRdkEiJJCn2dOOKI3qfw
-      POSTGRES_URI: postgresql+psycopg://zimfarm:zimpass@postgresdb:5432/zimfarm
+      POSTGRES_URI: postgresql+psycopg://zimfarm:zimpass@postgresdb:5432/zimfarm1
```
benoit74, line 48: Please revert this change

dev/docker-compose.yml
```diff
@@ -115,35 +130,41 @@ services:
...
-  #     context: ../workers
-  #     dockerfile: manager-Dockerfile
-  #   container_name: zf_worker_mgr
-  #   depends_on:
-  #     - backend
-  #   command: worker-manager --webapi-uri 'http://backend:8000/v1' --username test_worker --name test_worker
-  #   volumes:
-  #     - /var/run/docker.sock:/var/run/docker.sock
-  #     - ./test_worker-identity/id_rsa:/etc/ssh/keys/zimfarm
+  # uncomment this only if you want to run a worker manager
```
benoit74, line 133: Please comment out again worker_mgr and task_worker

benoit74 (approved):
LGTM, please take into account my comments, rewrite commits as discussed via Slack and you can merge, this is fine for me.

elfkuzco:
Applied the various changes introduced across #1204 , #1205 , #1206 , #1207  and #1208

PR #1211, opened 2025-08-01
use wrapvalidator to skip enum validation

elfkuzco:
# Rationale
As part of the zimfarm upgrade, validation should be skipped when reading data from db. However, some offliner values are defined as enums at the application level while the values in the db are not valid enum values. The existing `AfterValidator` which wraps validation logic for custom fields runs after pydantic internal validation and this would always fail because they are not valid enum values. Rather than annotate to `str` to relax this, I used a `WrapValidator` as that allows me to completely bypass pydantic internal validation based on context.

## Changes
- write validator for enum values that skips validation if `skip_validation` is True
- add tests for the validator


rgaudin (approved):
LGTM ; now I understand the other PR better 😉

PR #1212, opened 2025-08-01
Relax language schema validation while reading from db

elfkuzco:
## Changes
- skip language schema validation while reading from db
- sort by language name in english so frontend dropdown language list is ordered. can not sort in the db as the `distinct on` requires the first column to be the same as that used in `order_by`
- change frontend limit to 400 (same as original limit used by old UI)

### DIFF
--- a/backend/src/zimfarm_backend/db/language.py
+++ b/backend/src/zimfarm_backend/db/language.py
@@ -32,16 +32,22 @@ def get_languages(
 
     return LanguageListResult(
         nb_languages=count_from_stmt(session, query),
-        languages=[
-            LanguageSchema(
-                code=language_code,
-                name_en=language_name_en,
-                name_native=language_name_native,
-            )
-            for (
-                language_code,
-                language_name_en,
-                language_name_native,
-            ) in session.execute(query.offset(skip).limit(limit)).all()
-        ],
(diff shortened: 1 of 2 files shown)


backend/src/zimfarm_backend/db/language.py
```diff
@@ -32,16 +32,22 @@ def get_languages(
...
-        ],
+        languages=sorted(
+            [
+                LanguageSchema.model_validate(
+                    {
+                        "code": language_code,
+                        "name_en": language_name_en,
+                        "name_native": language_name_native,
+                    },
+                    context={"skip_validation": True},
```
rgaudin, line 43: What's the rationale for validating but skipping the validation? Is it that it's assumed the DB data is valid? What's difference with previous version?

elfkuzco: The DB data can sometimes be "invalid" owing to constraints changing. So, while reading from DB, we skip validating. But while updating or adding, we don't. Some of the data in the db fails the validation constraints and it won't be feasible to start modifying them. For example, the language code should be between length 2 and 8 but it's been observed some existing data is of length 12 and pydantic fails. So, we use context to skip validation on reads.

PR #1213, opened 2025-08-04
serialize datetime with Z suffix for naive datetimes

### LINKED ISSUE #1201: Zimfarm upgrade: Datetime presented in the UI are wrong
While the DB stores only naive datetime (assuming they are all UTC), in the past the backend API always returned TZ-aware datetime for proper UI rendering. It simply consisted in adding back the UTC TZ information.

Now, the API is returning naive datetime, and the UI is hence wrongly considering these are local times. 

One typical symptom is that a task which has just started is reported as having started 2 hours ago (because I'm UTC+2 currently).

This should be fixed in some way. I feel like restoring previous behavior where the API returned TZ-aware datetime is the good solution, because we cannot expect all API consumers to "guess" that all naive datetime are in fact UTC. The fact that we manipulate naive datetime at DB-leve and inside the backend is an internal optimization which should not leak to the API inputs / outputs.

elfkuzco:
## Changes
 - add `Z` suffix to datetime naive objects during serialization. This mimicks the same functionality provided by the `ZimfarmJSONEncoder` at  https://github.com/openzim/zimfarm/blob/200daaec43207519f9ebbf9b6bc6a14d3f5294df/dispatcher/backend/src/main.py#L36
 
 - relax language validation constraints  while reading schedules from db
 
 This closes #1201 as part of the zimfarm upgrade

### DIFF
--- a/backend/src/zimfarm_backend/common/schemas/__init__.py
+++ b/backend/src/zimfarm_backend/common/schemas/__init__.py
@@ -1,11 +1,23 @@
+import datetime
+
 import pydantic
 from pydantic import ConfigDict
 from pydantic.alias_generators import to_camel
 
 
+def serialize_datetime(value: datetime.datetime) -> str:
+    """Serialize datetime with 'Z' suffix for naive datetimes"""
+    if value.tzinfo is None:
+        return value.isoformat() + "Z"
+    return value.isoformat()
+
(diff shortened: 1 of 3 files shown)


rgaudin (approved):
LGTM

PR #1214, opened 2025-08-06
use vuetify server side tables to render data from API

elfkuzco:
## Changes

- use [Vuetify Server Side Tables](https://vuetifyjs.com/en/components/data-tables/server-side-tables/#api) to render data from the API
- delegate loading of data to the server side tables and remove from `onMounted` lifecycle to avoid duplicate API calls as it is called by the server side table
- remove calls to fetch data when limit changed on table as calls were duplicated
- use loading store to communicate loading state and text
- add slots for empty data from api and loading status in server tables
- update page title on route change
-

### DIFF
--- a/frontend-ui/src/App.vue
+++ b/frontend-ui/src/App.vue
@@ -28,6 +28,8 @@ onBeforeMount(async () => {
 })
 
 onMounted(async () => {
+  loadingStore.startLoading('Loading application data...')
+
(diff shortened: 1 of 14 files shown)


rgaudin (approved):
Wow, I can't wait to see this in action.
LGTM ; didn't test anything, just looked at the code.

PR #1215, opened 2025-08-07
rename s3optimizationcache to secreturl

### LINKED ISSUE #795: Set nautilus collection param as secret in offliner
To support nautilus-webui, we need to be able to hide the `collection` CLI flag of nautilus runs to unauthenticated users as this URL would be public and include URLs to files uploaded to S3.

We don't want those URLs to be known or people might abuse the tool to transfer files.

cc @FledgeXu

elfkuzco:
## Rationale
By changing `S3OptimizationCache` and `OptionalS3OptimizationCache` custom types to `SecretUrl` and `OptionalSecretUrl` respectively, the name becomes more informative when using it in other places that aren't actually S3 urls. Besides, it doesn't do any check to see if it's an S3 URL, just that it's a URL.

This was done while introducing a fix for #795 and it became clear there were other fields that could benefit from this generalization of the type name.

## Changes
- Set nautilus collection field to be a secret field
- Rename S3OptimizationCache types to SecretUrl to better capture it's intent.

This closes #795

### DIFF
--- a/backend/src/zimfarm_backend/common/schemas/fields.py
+++ b/backend/src/zimfarm_backend/common/schemas/fields.py
@@ -144,18 +144,16 @@ Percentage = Annotated[int, AfterValidator(between(low=1, high=100))]
 OptionalPercentage = Percentage | None
 
 
-def validate_optimization_cache(v: ZIMSecretStr | str) -> ZIMSecretStr:
+def validate_secret_url(v: ZIMSecretStr | str) -> ZIMSecretStr:
     url = v.get_secret_value() if isinstance(v, SecretStr) else v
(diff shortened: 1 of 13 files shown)


rgaudin (approved):
Makes sense ; LGTM

kelson42:
I don't understand the explanation.

"SecretURL" is meaning less in comparison to current name. It is also not clear to me which recipes are impacted! Why it is no clearly stated that this impacts only Nautilus?

In general, ZIMfarm recipe fields/options should be named the same as the one of the scrapers.

rgaudin:
The option is still named the same (`optimization_cache`), it's the field type that changes

Issue #1203, opened 2025-07-27
ZIM Metadata values (title, description, longdescription) are not escaped properly


kelson42:
Which means that if it contains a quote or single quote it will fail. Here and example:
https://farm.openzim.org/pipeline/8d87e5b5-1467-4325-ac13-220c014d58f2/debug

elfkuzco:
@benoit74 , can you take a look at this? #1216 escapes the flag arguments given they are processed by the shell. When passed to the scraper, they should be reverted to strings. If issue still persists, it's probably an issue of the scraper

benoit74:
Yep, to be deployed in production with `zimfarm-upgrade` branch

PR #1216, opened 2025-08-07
use shlex to escape config commands for offliners

### LINKED ISSUE #668: Escape quotes in config
See https://farm.openzim.org/pipeline/a7e335828162de57f03eaf16/debug. It seems the description argument is not quoted properly! Or maybe this is a ticket for Zimfarm?

elfkuzco:
## Changes
- use [shlex.quote](https://docs.python.org/3/library/shlex.html#shlex.quote) to escape offliner config flags while building commands. This fixes #668 and openzim/zim-requests#1058

This fixes #668

### DIFF
--- a/backend/src/zimfarm_backend/utils/offliners.py
+++ b/backend/src/zimfarm_backend/utils/offliners.py
@@ -1,4 +1,5 @@
 import pathlib
+import shlex
 from typing import Any, NamedTuple
 
 from pydantic import AnyUrl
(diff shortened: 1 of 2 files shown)


elfkuzco:
@rgaudin As per the [quote](https://docs.python.org/3/library/shlex.html#shlex.quote)  documentation, the output is designed to work only for UNIX shells. Given the commands generated are used in Docker shells, I think this is sufficient. Or are they used by other environments?

elfkuzco:
@kelson42 , is this sufficient to resolve #1203? or do you mean, it should be applied to the flags directly and not just the `command`. I reckon since the flags are never used directly (except in building commands), this should suffice. what do you think?

elfkuzco:
linked issues should not be closed till `zimfarm-upgrade` is merged to main.

kelson42:
I don't understand why we don't have an extensive test set of these security relevant PRs, features!?

@elfkuzco Thank you for working on this very old and nasty problem. I'm not sure to understand the difference in your question between the flags (you mean the API) and the command (I guess you mean the command line).

I can not be prescritive on how to fix this, because I don't know enough about the architecture of the solution... but this kind of things should be done rigourously, that means:
* there is a concept about what, where, how we check
* there is an extensive test set with everything which is forbidden

I see none of this and I'm therefore quite skeptical.

@rgaudin @benoit74 We need a written security concept about our API to guide the devs.

PR #1218, opened 2025-08-12
change token expires_in to isoformat string

elfkuzco:
## Rationale

While working on the changes required to make WP1 conform to the spec of the new API, it became clear that the `expires_in` of the token response needs to be changed to datetime string. This keeps the response similar to the previous API response and is also more informative than the constant I had been returning in the sense that clients can know when it would expire.

## Changes
- strip out microseconds from the datetime isoformat string to keep response compatible with older API
- use the same `issue_time` to create the access token and setting the `expires_in`. This reduces minor offsets in `exp` of access token and `expires_in` top-level field.

### DIFF
--- a/backend/src/zimfarm_backend/routes/auth/models.py
+++ b/backend/src/zimfarm_backend/routes/auth/models.py
@@ -1,3 +1,4 @@
+import datetime
 from typing import Literal
 from uuid import UUID
 
@@ -26,5 +27,5 @@ class Token(BaseModel):
 
     access_token: str
     token_type: str = "Bearer"
-    expires_in: float
+    expires_time: datetime.datetime
     refresh_token: str
--- a/backend/V1_TO_V2_MIGRATION.md
+++ b/backend/V1_TO_V2_MIGRATION.md
@@ -11,11 +11,13 @@ Zimfarm Backend v2 represents a significant modernization of the codebase, migra
 ### 1. Web Framework Migration
 
 **v1 (Flask):**
+
 - Uses Flask
 - Traditional WSGI application
 - Manual error handling with custom exception classes
 
 **v2 (FastAPI):**
+
 - Uses FastAPI
 - ASGI application with uvicorn server
 - Automatic request validation and error handling with Pydantic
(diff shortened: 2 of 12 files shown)


rgaudin (approved):
LGTM

What's the status regarding datetime and the API? Are we only sending datetimes formatted up to seconds? Is it clear in the generated doc?
What about input? Is there a place where we accept datetime?

elfkuzco:
> LGTM
> 
> What's the status regarding datetime and the API? Are we only sending datetimes formatted up to seconds? Is it clear in the generated doc? What about input? Is there a place where we accept datetime?

Regarding the datetime, I know there's a custom json encoder that adds `Z` suffix to datetime objects because they are not timezone aware in the db. https://github.com/openzim/zimfarm/blob/200daaec43207519f9ebbf9b6bc6a14d3f5294df/dispatcher/backend/src/utils/json.py#L8-L16

As far as I can see, there's only one place for datetime input which is for SSH authentication. That works fine because it's sent and processed as a string.

My main reason for adding the timespec to seconds was that during updating of wp1, i observed it uses a datetime string format to convert the data to datetime 
https://github.com/openzim/wp1/blob/c472eaa384f679f7a3737cf38108aa89e531eb37/wp1/zimfarm.py#L111-L113

Given datetime objects when called with `isoformat` include the milliseconds, I am having a bit of trouble in figuring out how it got stripped off from the `expires_in` which wp1 reads with this format.

I am going to leave the PR unmerged or convert to a draft till I get more context on this.

rgaudin:
OK, @benoit74 is coming back soon so maybe you'll discuss it together. I believe being explicit is better (thus using `timespec`) and I also think we should be consistent.

backend/src/zimfarm_backend/routes/auth/models.py
```diff
@@ -26,5 +27,5 @@ class Token(BaseModel):
 
     access_token: str
     token_type: str = "Bearer"
-    expires_in: float
+    expires_in: datetime.datetime
```
benoit74: Shouldn't we rename this to `expires_time` for clarity? If I get it correctly, it will be another breaking change to document in `V1_TO_V2_MIGRATION.md`, but it is the good time to do it.

benoit74 (approved):
I like the idea of stripping microseconds everywhere, I don't see where they would be meaningful.

Also take into account (minor) inline comment and we are good to merge.

backend/src/zimfarm_backend/routes/auth/models.py
elfkuzco: Okay

PR #1219, opened 2025-08-12
show uploaded pub key filename in ui

elfkuzco:
## Changes
- remove `ref` on v-file-input for uploaded pub key. this was preventing the filename from showing in the ui after being uploaded


rgaudin (approved):
LGTM

PR #1221, opened 2025-08-17
redesign views for small screens

elfkuzco:
## Changes

- make navbar icon clickable (redirects to home)
- show notification when access token copied
- use screen breakpoints to redesign grids on small screens


benoit74 (approved):
LGTM ; I did not tested it thoroughly on all screens, but I do not see any downside to merging this. At worst, we will find more changes needed on few places, which is not a problem.

PR #1223, opened 2025-08-18
use show_secrets context while fetching requested task

elfkuzco:
## Changes
- determine whether to show secrets while fetching requested task based on query param and permissions
- consider `hide_secrets` flag only if user has required permission

### DIFF
--- a/backend/src/zimfarm_backend/routes/requested_tasks/logic.py
+++ b/backend/src/zimfarm_backend/routes/requested_tasks/logic.py
@@ -280,15 +280,27 @@ def get_requested_task(
 
     # exclude notification to not expose private information (privacy)
     # on anonymous requests and requests for users without schedules_update
-    if not current_user or not check_user_permission(
-        current_user, namespace="schedules", name="update"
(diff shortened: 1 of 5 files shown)


backend/src/zimfarm_backend/routes/requested_tasks/logic.py
```diff
@@ -280,15 +280,27 @@ def get_requested_task(
...
     ):
         requested_task.notification = None
 
-    if hide_secrets:
-        return JSONResponse(content=requested_task.model_dump(mode="json"))
+    # if the user doesn't have permission, then their flag to hide secret
+    # does not matter
+    if not (
+        current_user
+        and check_user_permission(current_user, namespace="schedules", name="create")
```
benoit74: Why using `create` permission instead of `update`? I feel like having `update` permission is sufficient to have secrets shown (assuming `update` permission is "lower level" than `create`).

benoit74 (approved):
See comment and then we are good to merge.

Issue #783, opened 2023-05-17
Review all input validations


rgaudin:
In order to prevent incorrect data to be entered in the Zimfarm, we want to review all recipes inputs and update their constraints.

All constraints are in https://github.com/openzim/zimfarm/blob/main/dispatcher/backend/src/common/schemas/fields.py and https://github.com/openzim/zimfarm/tree/main/dispatcher/backend/src/common/schemas/offliners

Beside Offliner Flags, recipe inputs are:

- Recipe Name
- Language
- Tags
- Category
- Warehouse Path
- Status
- Periodicity
- Offliner
- Platform
- Image Name
- Image Tag
- Monitoring
- CPU
- Memory
- Disk
- RAM fs

**Please list below the list of changes you'd like** to constraints on those fields or Offliner flags. If a change is to be applied to all offliners (albeit using their own scaper-specific names), please say so.

My proposal:

- Recipe Name should enforce the [Naming Convention](https://github.com/openzim/overview/wiki/Naming-Convention)
- All fields setting `Name` metadata should enforce the Naming Convention
- All fields setting `Title` metadata should be restricted to 30 chars max (as per our [recommendations](https://wiki.openzim.org/wiki/Metadata))
- All fields setting the `Description` metadata should be restricted to 80 chars max
- All fields setting the `LongDescription` metadata should be restricted to 400chars max
- All fields setting the `Language` metadata should be checked for ISO-639-3 validity (as a comma-separated list).

rgaudin:
* All fields setting ZIM filename shoud enforce `{Name}_{period}.zim`

kelson42:
devdocs.ios failing scenario because of this weakness https://farm.openzim.org/pipeline/cb6a074a-e33a-4552-9c45-932486a1dde9/debug

rgaudin:
I believe it's a bit different: allowing arbitrary text in choice field or (more likely) a previously value choice that is not anymore and we did not update the recipe

benoit74:
I think that #910 proves that we also need to regularly check that flags are still valid.

We should check:
- does the flag still exists for given offliner?
- is the flag value still valid?

I've opened a distinct issue #911 since this is a bit different than just ensuring that constraints of all offliners are appropriate.

benoit74:
https://github.com/kiwix/k8s/issues/192 just shown how important it is to avoid unwanted chars in ZIM name / filename.

benoit74:
Removed from zimit2 project unfortunately

[...]

PR #1224, opened 2025-08-18
Enforce metadata conventions

elfkuzco:
## Rationale
This PR serves as foundation for enforcing zim metadata conventions and inputs for fields according to #783 

## Changes
- Change regex for `ZIMFilename` to conform to [Naming Convention](https://wiki.openzim.org/wiki/Content_team/ZIM_Naming_Convention)
- Change language code to be of 3 characters according to ISO 639-3 length

### DIFF
--- a/backend/src/zimfarm_backend/common/schemas/fields.py
+++ b/backend/src/zimfarm_backend/common/schemas/fields.py
@@ -170,7 +170,9 @@ OptionalZIMDescription = ZIMDescription | None
 ZIMFileName = Annotated[
     str,
     AfterValidator(
-        pattern(r"^(.+?_)([a-z\-]{2,3}?_)(.+_|)([\d]{4}-[\d]{2}|\{period\}).zim$")
+        pattern(
+            r"^([a-z0-9\-\.]+_)([a-z\-]{2,3}?_)([a-z0-9\-\.]+_|)([\d]{4}-[\d]{2}|\{period\}).zim$"
+        )
     ),
 ]
 
@@ -184,7 +186,7 @@ ZIMPlatformValue = Annotated[int, AfterValidator(between(low=1))]
 
 OptionalZIMPlatformValue = ZIMPlatformValue | None
 
-ZIMLangCode = Annotated[str, AfterValidator(length_between(low=2, high=8))]
+ZIMLangCode = Annotated[str, AfterValidator(length_between(low=3, high=3))]
 
 OptionalZIMLangCode = ZIMLangCode | None
 
--- a/backend/src/zimfarm_backend/common/schemas/fields.py
+++ b/backend/src/zimfarm_backend/common/schemas/fields.py
@@ -171,12 +171,19 @@ ZIMFileName = Annotated[
     str,
     AfterValidator(
         pattern(
-            r"^([a-z0-9\-\.]+_)([a-z\-]{2,3}?_)([a-z0-9\-\.]+_|)([\d]{4}-[\d]{2}|\{period\}).zim$"
+            r"^([a-z0-9\-\.]+_)([a-z\-]+_)([a-z0-9\-\.]+_)([a-z0-9\-\.]+_|)([\d]{4}-[\d]{2}|\{period\}).zim$"
         )
     ),
 ]
 
 OptionalZIMFileName = ZIMFileName | None
+
+ZIMName = Annotated[
+    str,
+    AfterValidator(pattern(r"^([a-z0-9\-\.]+_)([a-z\-]+_)([a-z0-9\-\.]+)$")),
+]
+
+OptionalZIMName = ZIMName | None
 
 SlackTarget = Annotated[str, AfterValidator(pattern(r"^[#|@].+$"))]
 
--- a/backend/tests/test_validators.py
+++ b/backend/tests/test_validators.py
@@ -4,7 +4,7 @@ import pytest
 from _pytest.python_api import RaisesContext
 from pydantic import ValidationError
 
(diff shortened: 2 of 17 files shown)


elfkuzco:
@rgaudin , in #783, you mentioned `LongDescription` should be 400 chars max but the [Naming Convention Wiki](https://wiki.openzim.org/wiki/Content_team/ZIM_Naming_Convention) says 4000. Is this a typo on your end? Or is the wiki not updated?

elfkuzco:
@benoit74 , a quick scan of the db reveals there are certain languages that might be difficult to migrate to iso639-3. For example: there's `Tarantino` which is an Italian dialect that has no iso 639-3 code.

rgaudin:
yes it's 4,000

backend/src/zimfarm_backend/common/schemas/fields.py
```diff
@@ -170,7 +170,9 @@ def validate_secret_url(v: ZIMSecretStr | str) -> ZIMSecretStr:
 ZIMFileName = Annotated[
     str,
     AfterValidator(
-        pattern(r"^(.+?_)([a-z\-]{2,3}?_)(.+_|)([\d]{4}-[\d]{2}|\{period\}).zim$")
+        pattern(
+            r"^([a-z0-9\-\.]+_)([a-z\-]{2,3}?_)([a-z0-9\-\.]+_|)([\d]{4}-[\d]{2}|\{period\}).zim$"
```
benoit74: The language portion must be allowed to be longer than 3 chars. Values like `nds-nl` are now allowed. `([a-z\-]{2,3}?_)` should be replaced with something like `([a-z\-]+?_)` (usually 6 is sufficient, but we do have exceptions like `en-simple`).

Please add such a test cases as well.

benoit74: And this misses the fact that selection is mandatory, and we have an optional part for flavor.

Correct expression is more `^([a-z0-9\-\.]+_)([a-z\-]+_)([a-z0-9\-\.]+_)([a-z0-9\-\.]+_|)([\d]{4}-[\d]{2}|\{period\}).zim$`.

See test cases to fix.

backend/tests/test_validators.py
```diff
@@ -28,3 +34,58 @@ def test_enum_validator_skips_validation_when_context_set():
         TestModel.model_validate(
             {"value": "invalid"}, context={"skip_validation": True}
         )
+
+
+@pytest.mark.parametrize(
+    "filename,expected",
+    [
+        ("wikipedia_en_2024-01.zim", does_not_raise()),
```
benoit74: `wikipedia_en_2024-01.zim`, `ted-talks_eng_2024-03.zim` and `ted-talks_eng_{period}.zim` are all invalid (they miss the "selection" part).

`wikipedia_en_all_2024-01.zim`, `ted-talks_eng_football_2024-03.zim` and `ted-talks_eng_football_{period}.zim` are all valid (selection)

`wikipedia_en_all_nopic_2024-01.zim` is valid as well (selection + flavor)

Nota: you should also fix other invalid tests below to have only one validation issue, e.g. replace `wikipedia_eng_2024-01` (which misses the selection and the .zim extension) with `wikipedia_eng_all_2024-01` (which misses only the .zim extension).

benoit74 (changes requested):
Thank you very much, totally forgot this regexp needed to be adapted to reflect recent changes in the convention and was somehow very significantly wrong.

This makes me realize that we miss a type constraint for `ZimName` (and optional `OptionalZimName`. It is present in most offliners and should match our convention as well. Correct regexp is `^([a-z0-9\-\.]+_)([a-z\-]+_)([a-z0-9\-\.]+)$`. Shall we add this in this same PR? (we can do it in a distinct one if you prefer)

Regarding the fact that we do have languages which do not have an exact ISO match, this is a known limitation and we do have a decision on this: we must use nearest ISO match. E.g. for Tarantino, we should have said it is `nap`. This seems to be the only way to ensure this is discoverable when using filters, and avoid have tons of "non-standard" languages. See https://wiki.openzim.org/wiki/Content_team/ZIM_Metadata_Convention and linked github issue for details on the discussion.

elfkuzco:
@benoit74 , should I rename the `ZIM*` types to `Zim*`? What do you think? Just a matter of style

elfkuzco:
> Sorry if I seem a bit picky, but could you please create same kind tests with `ZIMName` than with `ZIMFileName`?
> 
> I'm fine with `ZIMName` instead of `ZimName`, it is just me not yet remembering the convention we chose.

While developing test cases, I discovered `wikipedia_en_all.zim` conforms to the regex. Is this desirable?

benoit74:
> While developing test cases, I discovered `wikipedia_en_all.zim` conforms to the regex. Is this desirable?

Indeed this is not correct. Flavor is optional but period is mandatory. Please fix regex / tests. Thank you for notifying.

elfkuzco:
> > While developing test cases, I discovered `wikipedia_en_all.zim` conforms to the regex. Is this desirable?
> 
> Indeed this is not correct. Flavor is optional but period is mandatory. Please fix regex / tests. Thank you for notifying.

I meant for `ZIMName`, not `ZIMFileName`

benoit74:
Technically speaking, it matches the convention for now. Let's live with it for now, we've never faced the problem so far.

elfkuzco:
> Technically speaking, it matches the convention for now. Let's live with it for now, we've never faced the problem so far.

Good to merge then?

PR #1227, opened 2025-08-19
update offliner details on artifacts globs change

### LINKED ISSUE #994: Recipe edit form consider there is always a pending modification
Location: UI

See e.g. https://farm.openzim.org/recipes/iranwire.com_persian/edit

"Update Offliner details" button is always enabled while it should be only if there are pending modifications.

Looks like this is a zimit specific bug, or at least youtube, mwoffliner and sotoki are not impacted

Pending modification seems to be linked to `artifacts_globs`.

elfkuzco:
## Rationale
The schedule editor is set to track changes on the `artifacts_globs_str` but it is an array on the backend. So, while checking for changes, we convert the input to an array and compare with the original from the API.

By using the `stringArrayEqual`, order does not matter in comparison. This resolves #994

### DIFF
--- a/frontend-ui/src/components/ScheduleEditor.vue
+++ b/frontend-ui/src/components/ScheduleEditor.vue
@@ -565,13 +565,9 @@ const hasChanges = computed(() => {
     return true
 
   // Check artifacts globs
-  if (
-    !stringArrayEqual(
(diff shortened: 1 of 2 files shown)


benoit74 (approved):
Thank you, LGTM

PR #1228, opened 2025-08-20
Record ip change before updating on Wasabi

### LINKED ISSUE #1226: `zimfarm-upgrade`: restore current logic of commiting changes to DB asap
In https://github.com/openzim/zimfarm/pull/1205, logic has been changed (and simplified) to commit changes of worker IPs only after IP has been updated in Wasabi.

This simplification is however probably not possible because it had bad side-effects in the past.

With simplified logic, the IP change is recorded in DB only after Wasabi IP update is completed. 

The problem it caused in the past is that when Wasabi IP update is taking too long (e.g. because Wasabi is unavailable and it takes time to timeout), then we can have multiple calls to update the same worker new IP which are proceeding in parallel since first call did not yet updated the DB when second call arrives. This means we have multiple attempts in parallel to update Wasabi configuration, and this is a problem from Wasabi perspective, and only causing confusion in the logs.

I think we should go back to previous logic. To be discussed.

elfkuzco:
## Changes
- save ip changes to db before updating on wasabi as wasabi could be unavailable

This resolves #1226


benoit74 (approved):
LGTM, thank you! Please rebase and merge

PR #1229, opened 2025-08-20
use expires_time while constructing worker Token

elfkuzco:
## Changes

- use `expires_time` while constructing worker `Token` object


benoit74 (approved):
LGTM, please rebase and merge

PR #1230, opened 2025-08-20
make schedule name nullable

elfkuzco:
## Rationale
`schedule_id` is nullable for both `Task` and `RequestedTask`, so `schedule_name` should be nullable on pydantic model letting frontend/clients determine what to default to instead of `"none"`

## Changes
- make schedule name nullable and avoid returning `"none"` when it doesn't exist
- show `original_name`, `"none"`, or `"N/A"` on UI when schedule name is missing.
- avoid showing links for schedules that have been unset when viewing the task or worker.


frontend-ui/src/views/TaskDetailView.vue
```diff
@@ -71,7 +71,10 @@
                   <tr>
                     <th class="text-left w-20">Recipe</th>
                     <td>
-                      <router-link :to="{ name: 'schedule-detail', params: { scheduleName: task.schedule_name } }">
+                      <span v-if="task.schedule_name === null || task.schedule_name === 'none'">
```
benoit74, line 74: Why can we have `none` here? Looks like backend is not sending this value (anymore). I would rather get rid of this "oddity" unless we are sure we need it. To be removed in `WorkersView.vue` as well obviously.

Issue #896, opened 2024-01-15
Slash in schedule/recipe name is causing issues


benoit74:
- Location: API

## Problem

When we put a slash `/` in a recipe name, this seems to be saved to the database properly but afterwards we cannot open the recipe anymore in the UI:

![image](https://github.com/openzim/zimfarm/assets/7102089/608cce5f-d7ec-4de6-913b-42b9018fd27f)

What is fun (or not) is that even the preflight `OPTIONS` request end-up with a `404`, while when we just input a wrong recipe name the `OPTIONS` end-up with a `200` and the GET with a `404`

I think that we should probably forbid unusual characters in recipe names at creation / edition. alphanums + underscore is probably enough ; I would even forbid dash to avoid confusion between underscores and dashes.

The other solution (allow mostly any chars) is probably way more complex / not very useful.

rgaudin:
Keep in mind that we start the recipe name with domain name for which dash `-` are allowed

benoit74:
Good point, dots as well I imagine 🤣 (and probably other chars)

benoit74:
There is no other characters allowed in domain names, so I would change my mind:
- allow lower-case letters (only), no upper-case letters (to avoid duplicates changing only casing, just like in domain names)
- allow `.`, `-` (both are present in domain names) and `_` (used a lot in our naming conventions)
- allow digits

Is it better?

rgaudin:
Yes it's probably enough. Domain names can be almost anything now and that will be reflected in `Name` metadata but zimfarm could use an approximation for the 1% that actually use extended characters.
This should probably be discussed in a content meeting. What's the current way of doing with websites that are in non-ASCII charset like Arabic or Chinese?

benoit74:
Solved in `zimfarm-upgrade` branch

PR #1231, opened 2025-08-20
validate schedule names in ScheduleNameField

### LINKED ISSUE #634: Creating recipe with URL as title returns error
So I made the mistake of creating [https://www.skybrary.aero/](https://farm.openzim.org/recipes/https://www.skybrary.aero/) as a recipe name (cloning from another, unrelated recipe). 

It looks like it has been created, but I cannot access nor modify it and only get the following (and cryptic) error message: `Cross-Origin Request Blocked: preflight request failed`.

elfkuzco:
## Changes
- prevent slash `(/)` in schedule names in validator function for custom type `ScheduleNameField`. 
- disallow uppercase characters
This type is reused across routes that get, create, clone, update, delete, etc that deal with schedules

This PR resolves #634


elfkuzco:
@benoit74 , i think this PR could also tackle #896 . What do you think?

benoit74:
And yes, this fixes #896 as well

PR #1233, opened 2025-08-21
enhance assignment of tasks to workers with contexts

### LINKED ISSUE #909: Enhance assignment of tasks to workers
As of today, workers can select which scrapers they want to run.

But at a central level, we have no control over which workers are allowed to run which scrapers, and this could be an issue for various reasons:
- we want the task to be executed close to the server / on "approved" machines (e.g. mwoffliner tasks on mwoffliner machines)
- the task cannot succeed if not executed on given machines (typically due to IP whitelisting)

So, in addition to the platforms limitation, we probably also need to:
- control on a central side which workers can run which scrapers (e.g. mwoffliner should run only on mwoffliner workers) ; maybe at the recipe level ; or with specific tags
- control on a central side the affectation of some recipes to some worker(s) (e.g. for Zimit recipes of a Cloudflare protected website, we want the recipe to run on a single worker whose IP has been whitelisted)

### LINKED ISSUE #1139: Wikimedia recipes should run only on Wikimedia VPS workera
The reason is that these workers/servers benefit from a proviliged access (no request quota) to the Wikimedia API.

Today the assignement of taks to workers is based on worker based criterias:
* list of accepted scrapers
* memory/cpu limitations

Now we need a solution where part of the MWoffliner recipes have to be executed on part of the worker accepting MWoffliner recipes.

elfkuzco:
## Rationale
Requested tasks are assigned to workers based on their specified scrapers (platforms) and resources they specified. Thus, there is no control on the server side to limit which workers can run which recipes. By introducing tags for the recipes and providing similar for workers, an admin can create/edit a recipe with certain tags and edit workers which have no quota for said recipes. This allows community workers to retain control on the scrapers they can run while enriching admin with the decision to exclude them from running certain tasks. 

This resolves #909 
This resolves #1139 

## Changes
- allow workers to have multiple tags for restricting them to certain recipes
- allow recipes to specify tags which a worker must have to be able to request


elfkuzco:
NOTE: This PR is not complete yet as it would require changes on the UI. This is more of a PoC to get opinions on the implementation

benoit74:
I like that schedule / requested tasks have one single `worker_tag` and not a list, this greatly simplifies the matter. Together with the fact that worker can have multiple tags, it seems sufficient to cover #1139 use cases.

I like that at schedule level, the `worker_tag` is distinct from the regular tags we use to classify recipes.

I like that schedule `worker_tag` can be overwritten at requested task level. 

I wonder why we do not copy this `worker_tag` at task level as well (so that they are not lost when the requested task is transformed into a task, seems to be mandatory for traceability / debugging purposes).

I propose we choose another name than `tag` to describe this concept. `tag` is already used at schedule level and seems too broad / confusing. I propose to use `context` since this is not already used and maybe a bit clearer, but I'm open to suggestions. A schedule can be association with 0 to 1 context. A worker can be whitelisted to execute some `context`.

I'm not sure this PR changes totally covers #909 ; I realize it was not really expressed very clearly in the issue but one problem we also had in mind was that sometimes we do not want some scraper to run at all on some worker. Not some recipes of the scraper, but the whole scraper. The only example I have in mind is youtube which should only run on workers which IP has been whitelisted in Youtube API console. But the reality is that this we are worried about the fact that this might happen. But we never experienced such an issue, so it is maybe better to move this topic to a distinct issue to discuss first.

I'm pretty sure at some point we will need to make a distinction at worker level between `allow_contexts` and `deny_contexts`, in order to be able to allow tasks to be executed on any worker except few ones, but this is only a supposition without a real use case behind it, so no need to implement it for now.

@kelson42 @rgaudin feedback is welcomed

rgaudin:
I think I like the change as well but I lack some information.
- Can we have have listed here a few scenarios highlighting the difference between before and after.
- Not clear if this complements the `ZIMFARM_OFFLINERS` or replaces it.
- How are the worker_tag populated? Manually by ZF user? By worker config? Both?

Examples would help.

I also find the name _tag_ not appropriate.

elfkuzco:
>     * Can we have have listed here a few scenarios highlighting the difference between before and after.
The major difference between before and after is that there would be a way to restrict some workers from doing certain tasks. A good scenario would be in #1139  where even though a worker declares to process mwoffliner tasks, they might not have the privileges to process Wikimedia recipes with unrestricted access. So, we can mark such recipes with a tag/context to exclude them from processing such recipes while still allowing them to do other tasks associated with mwoffliner.

>     * Not clear if this complements the `ZIMFARM_OFFLINERS` or replaces it.
This complements the `ZIMFARM_OFFLINERS` which is a way for the worker to specify that the tasks they want to run are those associated with the offliners specified. This feature allows the admins to have some control on how tasks are assigned. More or less, it says "Even though you say you want to do this task, it's not available for you"

>     * How are the worker_tag populated? Manually by ZF user? By worker config? Both?
Could be populated via API call or the UI by ZF user. Or having a script do it directly at the DB.

elfkuzco:
> I wonder why we do not copy this `worker_tag` at task level as well (so that they are not lost when the requested task is transformed into a task, seems to be mandatory for traceability / debugging purposes).

Yeah, I could add that.

> I propose we choose another name than `tag` to describe this concept. `tag` is already used at schedule level and seems too broad / confusing. I propose to use `context` since this is not already used and maybe a bit clearer, but I'm open to suggestions. A schedule can be association with 0 to 1 context. A worker can be whitelisted to execute some `context`.


> I'm pretty sure at some point we will need to make a distinction at worker level between `allow_contexts` and `deny_contexts`, in order to be able to allow tasks to be executed on any worker except few ones, but this is only a supposition without a real use case behind it, so no need to implement it for now.
> 

Could you elaborate more on the context. Do you mean the implemented logic should be renamed from `tag` to `context` while keeping things the same? Or do you mean, it's shape could be changed to be a dictionary and then further enhanced to have allow/deny rules?

benoit74:
> Could you elaborate more on the context. Do you mean the implemented logic should be renamed from tag to context while keeping things the same? Or do you mean, it's shape could be changed to be a dictionary and then further enhanced to have allow/deny rules?

For now, only rename from `tag` to `context` (or something even better if someone finds it).

>  How are the worker_tag populated? Manually by ZF user? By worker config? Both?

They are populated centrally by ZF user. And mostly manually (pretty sure I will set few through DB query or API calls when there are too many recipes to set). Worker config cannot change this (for now at least).


>  Can we have have listed here a few scenarios highlighting the difference between before and after.

### Wikimedia recipes

All wikimedia recipes will have `wikimedia` context, and all mwofflinerx workers will have the `wikimedia` context. This will ensure wikimedia recipes are ran only on mwofflinerx workers, where today they can run on any worker

### vikidia

vikidia has whitelisted some of our IPS. All vikidia recipes will have the `vikidia` context, and all workers whose IPs have been whitelisted will have the `vikidia` context. This will ensure vikidia recipes are ran only on workers whose IPs have been whitelisted.

Same will be done for other IP whitelistings. We will use on context per "hosting". E.g. we will avoid to reuse `vikidia` context for `iranwire.com` just because the workers are the same today. This might (and will) inevitably get out of sync at some point, so better to have two (many) distinct context. On the other hand only one `vikidia` context for all vikidia recipes is OK because their is only one "hosting/infra" behind all these vikidia recipes.

benoit74:
@kelson42 @rgaudin any more thought before we move on with the implementation of UI?

rgaudin:
LGTM ; satified with both your explanations.

kelson42:
LGTM

benoit74:
@elfkuzco do you wanna proceed with UI implementation in same PR or in another PR? 

Probably good to share how you intend to modify things before doing all the coding, so that we ensure we are all aligned.

We need to:
- select an existing context for a given schedule
- remove any context for a given schedule
- add / remove existing contexts from workers
- create new context (only on workers ? or also on schedules ? if having both is cheap, then I think we should have both)

Deleting a context is not required since we do not really maintain a list of context, it is just the union of all contexts used on tasks and on workers. It make me think we might want to have an API to get the list of existing contexts.

elfkuzco:
I would like to apply it in the same PR.

For the recipe, I was originally going to show it as a text field in the schedule editor as there's  not really a list of options to them to select.

for the worker, same thing would apply in their edit tab.

> create new context (only on workers ? or also on schedules ? if having both is cheap, then I think we should have both)

by this, do you mean having an endpoint for creating contexts that applies to schedules/workers in batch?

benoit74:
If everything is text-based (like not a combobox), then we are good indeed. 

I was thinking that everything would be in a combobox, so I was a bit anticipating. 

The combobox approach seems "cleaner" in that it avoids silly typo which are going to be hard to debug (you will never notice immediately there is a typo, but only once a problem hits you because task started on wrong worker ; could take months or years if you are "lucky").

A combobox on schedule side to choose one existing context + the possibility to input a new one.

A combobox with checkboxes on worker side to select which contexts apply + the possibility to add new ones.

In both cases, the list of context is the union of contexts defined on any schedule or any worker ... hence the potential need for an additional API endpoint.

Do you feel like it is not going to be worth the effort?

elfkuzco:
It's worth it. I agree with you that the text-based approach would easily lead to typos in contexts. Would implement it this way. By combo box, do you mean, a text field which would show dropdown of exisitng contexts as user types?

benoit74:
> By combo box, do you mean, a text field which would show dropdown of exisitng contexts as user types?

Yes

elfkuzco:
will implement the UI in a different PR

benoit74 (commented):
Why using empty context like `""` instead of `None` when a schedule/task has no context? I feel like using `None` is more natural, but I might miss something. Note that DB column must be nullable if we choose to do so. But again, this feel more natural to me.

elfkuzco:
> Why using empty context like "" instead of None when a schedule/task has no context? 
Initially, I had made it a nullable column but it became clear that it could be an empty string if a user made a call with context set to empty string. To account for this, I initially did the query like this. 
```py
        or_(
            RequestedTask.context == "",
            RequestedTask.context.is_(None),
            RequestedTask.context.in_(worker.contexts),
        )
  ```
  Even though it might look like one would never intentionally set it to empty string from the UI, it can be done via the API. 
  Or should we enforce constraints on it like `NotEmptyString` or string with certain minimum length?

elfkuzco:
Another consequence of setting it to be a nullable Field is with the `update_schedule` function. if it's `str | None`, the user would always have to pass it because we can't distinguish if they are explicitly setting None or they omitted it (in which case they don't want it to be set). The `update_schedule` function is used the `patch` method of the endpoint. 

What do you think?

benoit74:
Got it. Makes sense indeed, will complete review.

PR #1238, opened 2025-08-22
detect/update validity of schedule

elfkuzco:
## Rationale

When we update an offliner details, we don't get feedback till the backend processes it. This PR implements an endpoint for validating a schedule and validates the schedule anytime a value is changed from the UI.

## Changes
- add `is_valid` column to `Schedule` table. This allows schedules validity to be stored while updating/reading.
- `is_valid` is re-computed when schedule is retrieved by client as DB might not represent actual validity
- notify user when they set an invalid value for a recipe immediately
-


frontend-ui/src/views/ScheduleDetailView.vue
```diff
@@ -832,6 +860,8 @@ onMounted(async () => {
       flagsDefinition.value = offlinerDefinition.flags
     }
     await fetchScheduleImageTags(schedule.value.config.image.name)
+    // Validate the schedule on mount
```
benoit74: Is this necessary since we already validate on `GET /schedule/xxx` in the backend?

frontend-ui/src/views/ScheduleDetailView.vue
```diff
@@ -742,13 +759,24 @@ const updateSchedule = async (update: ScheduleUpdateSchema) => {
         notificationStore.showError(error)
       }
     }
+    // Validate the schedule after update
```
benoit74: Is this necessary since we already validate schedule at update in the backend?

benoit74 (changes requested):
See comments.

Plus for some reason, this is not working on my machine. Looks like all `GET /schedules/xx` are ending with a 500 error. Can you check is works fine on your end? Did I missed something?

elfkuzco:
> Plus for some reason, this is not working on my machine. Looks like all GET /schedules/xx are ending with a 500 error. Can you check is works fine on your end? Did I missed something?

Did you run the migrations?
Added the `is_valid` flag on the DB model with a server default of `true` (assumption is that models saved in the db are valid). However, it is re-validated when fetching a schedule and the proper validity is shown to the client.

frontend-ui/src/views/ScheduleDetailView.vue
elfkuzco: The rationale was that when a user fetches a schedule, we return the schedule if it exists. However, we only set the `is_valid` to notify them if the schedule meets current constraints. There is no way for them to know which fields are wrong with the schedule that was fetched. And I didn't want to return an error if a schedule was invalid.

So, on the UI, after fetching, we validate so we can see the specific fields that are wrong. They can then go to the Editor and resolve

frontend-ui/src/views/ScheduleDetailView.vue
elfkuzco: Thought the same reason as the previous would apply. Would fix. Thanks.

benoit74:
> > Plus for some reason, this is not working on my machine. Looks like all GET /schedules/xx are ending with a 500 error. Can you check is works fine on your end? Did I missed something?
> 
> Did you run the migrations? Added the `is_valid` flag on the DB model with a server default of `true` (assumption is that models saved in the db are valid). However, it is re-validated when fetching a schedule and the proper validity is shown to the client.

This is a good point indeed, I did not ran migrations. But they are supposed to be ran automatically at server startup, so it looks like we broke something. Can you please have a look?

frontend-ui/src/views/ScheduleDetailView.vue
benoit74: OK, please explain this in comment (something short like `Validate the schedule on mount to get details about validation errors`).

And please rebase once https://github.com/openzim/zimfarm/pull/1241 is merged, my DB is more or less lost in translation ^^

elfkuzco:
converting this to a draft till I implement the PR that returns constraints to frontend as this would rely on it.

frontend-ui/src/views/ScheduleDetailView.vue
elfkuzco: Done.

PR #1243, opened 2025-08-25
remove old tasks periodically

### LINKED ISSUE #1125: Cleanup old tasks and schedules
On zimit.kiwix.org instance, we want to automatically cleanup old tasks, requested tasks and schedules since these objects are not expected to be of any value once the task completes (the schedules are even supposed to be automatically cleaned up by the external system creating it ... but it does not always achieves to do so).

Zimfarm should hence be capable to support this, customizable via env vars just like the rest.

Configuring this on our various instance will be another issue

elfkuzco:
## Rationale
To reduce the size of the DB backups, it is imperative to delete old tasks from the DB since they are not expected to be of any value once the task completes.

This resolves #1125

## Changes
- add function to remove tasks periodically to `periodic-tasks.py`
- add env vars to control removal.
     - `SHOULD_REMOVE_OLD_TASKS` determines whether tasks should be removed
     - `TASKS_OLDER_THAN` controls the number of days in the past to start deleting from


dispatcher/backend/src/periodic-tasks.py
```diff
@@ -148,6 +155,49 @@ def staled_statuses(session: so.Session):
...
 
 
+def tasks_cleanup(session: so.Session):
+    """removes old tasks excluding most recent tasks for schedules"""
+
+    if not SHOULD_REMOVE_OLD_TASKS:
+        logger.info(f":: skipping removal of tasks older than {TASKS_OLDER_THAN} days")
+        return
+
+    logger.info(f":: removing tasks older than {TASKS_OLDER_THAN} days")
```
benoit74: We should compute the "cutover" date and log it in this message, so that it is clear which tasks we've deleted.

elfkuzco:
Modified the `most_recent_task_id` constraint on the schedule to be set to null when a task is deleted and simplified the query for deletion. This would require an alembic migration to be made

PR #1246, opened 2025-08-27
Extract offliner validation constraints from API

elfkuzco:
## Rationale
By returning the same validation constraints (`min_length`, `max_length`, etc) used in the definition of an offliner from the backend, the UI/clients can reuse these validation rules to enforce constraints on their end e.g the ScheduleEditor. 

## Changes
- extract validation information from fields and return to clients
- use in schedule editor for client side validation of user inputs
- default to original Pydantic `Field` for passing validation rules
- use [WrapValidator](https://docs.pydantic.dev/latest/concepts/validators/#field-wrap-validator) to bypass validation instead of `AfterValidator` for cases where the validation information can be passed to `Field`

### DIFF
--- a/backend/src/zimfarm_backend/common/schemas/fields.py
+++ b/backend/src/zimfarm_backend/common/schemas/fields.py
@@ -1,5 +1,4 @@
-import re
-from collections.abc import Callable, Sized
+from collections.abc import Callable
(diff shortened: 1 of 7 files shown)


benoit74:
I let you merge at your convenience

PR #1241, opened 2025-08-25
Standardize schedule language codes

elfkuzco:
## Changes

- standardize language codes for schdules to ISO 639-3 format
- use only language_code in DB and drop `name_en` and name_native` metadata. `name` is generated using `pycountry`
- allow clients to only specify `language` code while operating on schedules endpoints
- provid defaults that skip validation while reading old data from dB

Issue #1185, opened 2025-07-22
Merge/deploy `zimfarm_upgrade` branch


benoit74:
We're preparing a big breaking upgrade of the Zimfarm in `zimfarm_upgrade` branch.

This is not yet ready, especially since we need to prepare all systems for this change.

This issue is a placeholder to track everything which will need to be done on D-day when we will merge the branch.

- [x] Implement https://github.com/openzim/zimfarm/issues/1125, https://github.com/kiwix/operations/issues/436 and https://github.com/kiwix/operations/issues/437 to backup farm.zimit.kiwix.org DB
- [x] When long awaited tasks (see list below) are close to completion:
   - [x] Post a message on Slack #zimfarm channel informing worker owners of the near upgrade, that during the upgrade their worker will not be able to communicate with the API and that after the upgrade they will have to restart it with `zimfarm restart` to grab updated version
   - [x] Disable scheduling on farm.openzim.org (so that no new tasks are requested)
- [x] Check that all code from `main` branch has been backported to `zimfarm-upgrade` branch
- [x] Disable continuous deployment in both `main` and `zimfarm-upgrade` branch (comment Github job / workflow) but keep Docker image building
- [x] Let long awaited tasks complete on farm.openzim.org:
  - [x] [wikipedia_en_all_maxi](https://farm.openzim.org/recipes/wikipedia_en_all_maxi)
  - [x] [stackoverflow.com_en](https://farm.openzim.org/recipes/stackoverflow.com_en)
- [x] Post a message on Slack #zimfarm channel informing worker owners that migration is running
- [x] Put zimit-frontend in "maintenance" mode (`stop_new_requests_on` in frontend config JSON file)
- [x] Put WP1 in "maintenance" mode
- [x] Check again that all code from `main` branch has been backported to `zimfarm-upgrade` branch
- [x] Check again that continuous deployment is disabled in both `main` and `zimfarm-upgrade` branch
- [x] Cancel all remaining tasks and wait for them to be really cancelled
- [x] Shutdown API of farm.openzim.org and farm.zimit.kiwix.org
- [x] Perform one last manual DB backup of farm.openzim.org and farm.zimit.kiwix.org
- [x] Tag `main` commit as `zimfarm-v1`
- [x] Force push `main` to `zimfarm-upgrade`
- [x] Check CI has build again all Docker images from new `main` commit
- [x] Run DB migration SQL on farm.zimit.kiwix.org DB
[...]

PR #1247, opened 2025-08-27
Convert DB schedule language codes to ISO 639-3 codes

elfkuzco:
## Rationale

Given the standardization of schedule language codes to ISO 639-3 in #1241, it is imperative to update existing data in the DB to conform to the standard.

This PR is not intended to be merged. Created mostly for discussion regarding codes for which there is no clear replacement to use and would require suggestions. If satisfactory, will be closed (or converted to draft) and will be referenced in #1185

### DIFF
--- a/backend/update_language_codes.py
+++ b/backend/update_language_codes.py
@@ -21,13 +21,15 @@ class Language(BaseModel):
 normalization_map: dict[str, Language] = {
     "bat-smg": Language(name="Samogitian", alpha_3="sgs"),
     # There is no ISO 639-3 code for Belarusian (Taraškievica orthography)
-    # "be-x-old": Language(name="Belarusian (Taraškievica orthography)", alpha_3=...)
+    # So, we use the nearest parent language => bel
+    "be-x-old": Language(name="Belarusian (Taraškievica orthography)", alpha_3="bel"),
     "bh": Language(name="Bhojpuri", alpha_3="bho"),
     "cbk-zam": Language(name="Chavacano", alpha_3="cbk"),
     "nds-nl": Language(name="Low Saxon", alpha_3="nds"),
     # Emiliano-Romagnolo  ISO 639-3 code has been deprecated as the ethnicities have
     # been split into Emilian and Romagnolo. What should we default to? egl for Emilian
-    # or rgn for Romagnol?
+    # or rgn for Romagnol? For now, use mul
+    "eml": Language(name="Emilian-Romagnol", alpha_3="mul"),
     "fiu-vro": Language(name="Võro", alpha_3="vro"),
     "in": Language(name="Indonesian", alpha_3="ind"),
     "iw": Language(name="Hebrew", alpha_3="heb"),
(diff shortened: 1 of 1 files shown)


elfkuzco:
`pycountry` exposes a `add_entry` method that offers the capabilities to register custom codes to fit application needs. One approach to encompass these "unknown" codes would be to register them at module initialization

backend/update_language_codes.py
```diff
@@ -0,0 +1,80 @@
...
+# alpha_3 represents the ISO 639-3 code of the language. So, the actual name of the
+# language might vary from the name in the DB when we rely on pycountry for the name
+normalization_map: dict[str, Language] = {
+    "bat-smg": Language(name="Samogitian", alpha_3="sgs"),
+    # There is no ISO 639-3 code for Belarusian (Taraškievica orthography)
+    # "be-x-old": Language(name="Belarusian (Taraškievica orthography)", alpha_3=...)
+    "bh": Language(name="Bhojpuri", alpha_3="bho"),
+    "cbk-zam": Language(name="Chavacano", alpha_3="cbk"),
+    "nds-nl": Language(name="Low Saxon", alpha_3="nds"),
+    # Emiliano-Romagnolo  ISO 639-3 code has been deprecated as the ethnicities have
```
benoit74, line 29: We don't know, there is probably a mix of both `egl` and `rgn` in the wiki, so we should use `mul`.

backend/update_language_codes.py
```diff
@@ -0,0 +1,80 @@
...
+
+# For languages in the DB that we can't find in pycountry, we can use this map to
+# normalize the language code. The key represents the language code in the DB.
+# For each value, the name represents the name of the language in the DB and the
+# alpha_3 represents the ISO 639-3 code of the language. So, the actual name of the
+# language might vary from the name in the DB when we rely on pycountry for the name
+normalization_map: dict[str, Language] = {
+    "bat-smg": Language(name="Samogitian", alpha_3="sgs"),
+    # There is no ISO 639-3 code for Belarusian (Taraškievica orthography)
+    # "be-x-old": Language(name="Belarusian (Taraškievica orthography)", alpha_3=...)
```
benoit74: We then use nearest parent language => `bel`

backend/update_language_codes.py
```diff
@@ -0,0 +1,80 @@
...
+    "nds-nl": Language(name="Low Saxon", alpha_3="nds"),
+    # Emiliano-Romagnolo  ISO 639-3 code has been deprecated as the ethnicities have
+    # been split into Emilian and Romagnolo. What should we default to? egl for Emilian
+    # or rgn for Romagnol?
+    "fiu-vro": Language(name="Võro", alpha_3="vro"),
+    "in": Language(name="Indonesian", alpha_3="ind"),
+    "iw": Language(name="Hebrew", alpha_3="heb"),
+    "mo": Language(name="Romanian", alpha_3="ron"),
+    # Some languages in the DB are families of languages, so we can't use a single
+    # language code. What should we default to? Such languages include:
```
benoit74, line 38: There are multiple ISO639-5 codes in these recipes => we should use `mul`

backend/update_language_codes.py
```diff
@@ -0,0 +1,80 @@
...
+    #  - Mayan languages
+    #  - Nahuati and Nahuatl langauges
+    #  - North American Indian languages
+    "roa-rup": Language(name="Aromanian", alpha_3="rup"),
+    "roa-tara": Language(name="Tarantino", alpha_3="nap"),
+    # some languages are unknown in the db. Such languages include:
+    # - Unknown language [ef]
```
benoit74: `ef` => https://en.wikipedia.org/wiki/Efik_language => `efi`

backend/update_language_codes.py
```diff
@@ -0,0 +1,80 @@
...
+    # - Unknown language [sp]
```
benoit74: `sp` => Sepedi => https://en.wikipedia.org/wiki/Northern_Sotho => `nso`

backend/update_language_codes.py
```diff
@@ -0,0 +1,80 @@
...
+
+            if schedule.language_code != language.alpha_3:
+                logger.info(
+                    f"Updating schedule language code {schedule.language_code} to "
+                    f"{language.alpha_3}"
+                )
+
+            # Normalize the language code to the ISO 639-3 code
+            schedule.language_code = language.alpha_3
+            session.add(schedule)
```
benoit74: Do we really need to add it to the session, since it has been selected in the session anyway?

benoit74 (changes requested):
Luckily, I had an answer for all open points ^^

For the record, all of these "decisions" comes from what has been discussed and decided in https://github.com/openzim/overview/issues/51

elfkuzco:
Applied changes and closing PR without merging.

benoit74:
LGTM

PR #1248, opened 2025-08-27
Always serialize offliners by alias

elfkuzco:
## Rationale
All offliners either inherit from `DashModel` or `CamelModel`. In order to avoid omission of calling `by_alias=True` while serializing the model to DB, it is necessary to add `serialize_by_alias=True` to such base models since that is the goal of having them in the first place

## Changes

- add `serialize_by_alias=True` to `DashModel` and `CamelModel`
- remove all `by_alias=True` as it is now handled automatically


benoit74 (approved):
So obvious once you realize it ^^

PR #1249, opened 2025-08-28
use combobox for showing list of schedule contexts

elfkuzco:
## Changes
- use combo box to show list of contexts from backend in the UI
- fuzzy search user input against current contexts
- add store for retrieving contexts from backend

### DIFF
--- a/frontend-ui/src/components/ScheduleEditor.vue
+++ b/frontend-ui/src/components/ScheduleEditor.vue
@@ -125,12 +125,12 @@
           persistent-hint
         />
       </v-col>
-      <v-col cols="12" sm="6">
+      <v-col cols="12" sm="4">
         <v-combobox
           v-model="editSchedule.context"
           :items="contexts"
           label="Context"
-          hint="Defines what values a worker must have to be able to execute tasks for this schedule"
+          hint="Execute schedule only on workers associated with this context"
           placeholder="Context"
           :clearable="!!editSchedule.context"
           density="compact"
--- a/frontend-ui/src/components/ScheduleEditor.vue
+++ b/frontend-ui/src/components/ScheduleEditor.vue
@@ -126,14 +126,18 @@
         />
       </v-col>
       <v-col cols="12" sm="6">
-        <v-text-field
+        <v-combobox
           v-model="editSchedule.context"
+          :items="contexts"
           label="Context"
           hint="Defines what values a worker must have to be able to execute tasks for this schedule"
           placeholder="Context"
+          :clearable="!!editSchedule.context"
           density="compact"
           variant="outlined"
           persistent-hint
+          :menu-props="{ maxHeight: '200px' }"
+          :custom-filter="fuzzyFilter"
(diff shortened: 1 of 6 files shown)


benoit74:
Waiting for other PRs to settle so that I can run migration scripts on my local DB and test this properly. But it seems OK at first sight, and we might want to use this fuzzy search text input in many other places in the UI (e.g the `Language` schedule combobox which is a pain to use, even if for this field we do not want to allow new values so it will be a bit different)

frontend-ui/src/components/ScheduleEditor.vue
```diff
@@ -126,14 +126,18 @@
         />
       </v-col>
       <v-col cols="12" sm="6">
```
benoit74: Why `sm="6"` instead of `sm="4"`? I feel like this would be more consistent with the 4.

frontend-ui/src/components/ScheduleEditor.vue
```diff
@@ -126,14 +126,18 @@
...
-        <v-text-field
+        <v-combobox
           v-model="editSchedule.context"
+          :items="contexts"
           label="Context"
           hint="Defines what values a worker must have to be able to execute tasks for this schedule"
```
benoit74: `Defines what values a worker must have to be able to execute tasks for this schedule`

=> `Execute schedule only on workers associated with this context`

frontend-ui/src/components/ScheduleEditor.vue
elfkuzco: yeah, hint i had was too long and i figured `6` would be wide enough to make it sufficient. now that i have reduced the wording, `4` captures it better

elfkuzco:
>  I do not achieve to clear the context once it has been set on a schedule, can you please check?


think i messed something up during the rebase but it's back as it should now. Verify?

benoit74 (changes requested):
1. We should add this context information in the "Info" tab as well:

image

2. when I set the context on a schedule, move to another schedule, the context I just set is not available in the dropdown ; I need to reload the app to see it ; we should probably add it directly to the list of available contexts, or just trigger a refresh ; this is kinda important to avoid typo / make things faster when setting context on a bunch of recipes ; if we choose to trigger a refresh of contexts, ideally this should be done only when context is changed, to not reload contexts at every schedule modification (even if it does not happen that often, so reloading at every schedule modification is acceptable if significantly simpler)

3. when I remove a context from last schedule (and soon worker) using it, it is still available in the dropdown, until I refresh the whole app obviously ; this is less important but should we decide to trigger a refresh on point 2 above, we should do so when unsetting context as well ; otherwise never mind, it is just too complex to check if context is still used somewhere only from the UI

elfkuzco:
applied changes to reload contexts if `context` is set in payload to `PATCH` and contexts are now shown in Info tab

benoit74 (approved):
LGTM, thank you

PR #1250, opened 2025-08-28
Re-design workers view

elfkuzco:
## Changes
- use server side tables to render workers list
- show tasks for worker in collapsible table
- use cards to show statistics
Screenshot_20250828_150600

### DIFF
--- a/frontend-ui/src/components/WorkersTable.vue
+++ b/frontend-ui/src/components/WorkersTable.vue
@@ -1,0 +1,229 @@
+<template>
+  <div>
+    <v-card v-if="!errors.length" :class="{ 'loading': loading }" flat>
+      <v-card-title class="d-flex align-center justify-end">
+        <div class="d-flex align-center">
+          <v-btn
+            size="small"
+            variant="outlined"
+            color="primary"
+            :prepend-icon="expandedAll ? 'mdi-unfold-less-horizontal' : 'mdi-unfold-more-horizontal'"
+            class="text-none"
+            @click="toggleAllRows"
+          >
+            {{ expandedAll ? 'Collapse all' : 'Expand all' }}
+          </v-btn>
+        </div>
+      </v-card-title>
+
+      <v-data-table-server
+        :headers="workerHeaders"
+        :items="workers"
+        :loading="loading"
+        :items-per-page="selectedLimit"
+        :items-length="props.paginator.count"
+        :items-per-page-options="limits"
+        class="elevation-1"
+        hover
+        item-key="name"
+        item-value="name"
+        v-model:expanded="expanded"
+        show-expand
+        @update:options="onUpdateOptions"
+        :hide-default-footer="props.paginator.count === 0"
+        :hide-default-header="props.paginator.count === 0"
+      >
+        <template #loading>
+          <div class="d-flex flex-column align-center justify-center pa-8">
+            <v-progress-circular indeterminate size="64" />
+            <div class="mt-4 text-body-1">{{ loadingText || 'Fetching workers...' }}</div>
+          </div>
+        </template>
+
+        <template #[`item.name`]="{ item }">
+          <code>{{ item.name }}</code>
+        </template>
+
+        <template #[`item.status`]="{ item }">
+          <v-chip
+            :color="item.status === 'online' ? 'success' : 'dark-grey'"
+            size="small"
+            variant="tonal"
+            :prepend-icon="item.status === 'online' ? 'mdi-server' : 'mdi-skull-crossbones'"
+          >
+            {{ item.status }}
+          </v-chip>
+        </template>
+
+        <template #[`item.last_seen`]="{ item }">
+          <v-tooltip location="bottom">
+            <template #activator="{ props }">
+              <span v-bind="props">{{ fromNow(item.last_seen) }}</span>
+            </template>
+            <span>{{ formatDt(item.last_seen) }}</span>
+          </v-tooltip>
+        </template>
+
+        <template #[`item.resources`]="{ item }">
+          <div class="d-flex flex-column flex-md-row py-1">
+            <ResourceBadge kind="cpu" :value="item.resources.cpu" variant="text" />
+            <ResourceBadge kind="memory" :value="item.resources.memory" variant="text" />
+            <ResourceBadge kind="disk" :value="item.resources.disk" variant="text" />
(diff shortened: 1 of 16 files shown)


frontend-ui/src/components/WorkersTable.vue
```diff
@@ -0,0 +1,190 @@
...
+            <div class="text-body-1">No workers found</div>
+          </div>
+        </template>
+
+        <template #[`item.data-table-expand`]="{ internalItem, isExpanded, toggleExpand }">
+          <v-btn
+            :append-icon="isExpanded(internalItem) ? 'mdi-chevron-up' : 'mdi-chevron-down'"
+            :text="isExpanded(internalItem)
+              ? `Collapse (${internalItem.raw?.tasks?.length || 0})`
+              : `Recent tasks (${internalItem.raw?.tasks?.length || 0})`"
```
benoit74: Recent tasks => Running tasks

benoit74 (changes requested):
I feel like we should expand rows by default, allowing users to close only to save space.

At least this is the only screen I've found to answer my regular questions I have about workers capacity, e.g. which mwoffliner has not-so-long tasks running and could start my schedule next? Having to expand all workers one-by-one is going to be painful. Or maybe having a button to expand / collapse all at once?

elfkuzco:
Add a button to Expand/Collapse all. Selection is saved in cookie and re-used on reload

frontend-ui/src/components/WorkersTable.vue
```diff
@@ -0,0 +1,254 @@
...
+  errors: string[]
+  loadingText: string
+}>()
+
+const emit = defineEmits<{
+  'limitChanged': [limit: number]
+  'loadData': [limit: number, skip: number]
+}>()
+
+const limits = [25, 50, 75, 100]
```
benoit74: This is not consistent with the backend which set 20 as default ; and this is not consistent with other tables ; why not using same thresholds as other screens (10, 20, 50, 100)?

frontend-ui/src/stores/workers.ts
```diff
@@ -10,20 +10,28 @@ import { ref } from 'vue'
...
+    limit: 100,
+    count: 0,
+  })
 
   const authStore = useAuthStore()
 
-  const fetchWorkers = async (params: {limit?: number} = {}) => {
-    const { limit = 100 } = params
+  const fetchWorkers = async (params: { limit?: number; skip?: number } = {}) => {
+    const { limit = 100, skip = 0 } = params
```
benoit74: why having 100 as default instead of 20?

frontend-ui/src/stores/workers.ts
```diff
@@ -10,20 +10,28 @@ import { ref } from 'vue'
 export const useWorkersStore = defineStore('workers', () => {
   const workers = ref<Worker[]>([])
   const errors = ref<string[]>([])
+  const paginator = ref<Paginator>({
+    page: 1,
+    page_size: 100,
```
benoit74: why 100 instead of 20 / 25 ?

frontend-ui/src/views/WorkersView.vue
```diff
@@ -148,41 +111,53 @@ const loadingStore = useLoadingStore()
...
+  { title: 'Contexts', value: 'contexts' },
+]
+
+const taskHeaders = [
+  { title: 'Schedule', value: 'schedule' },
+  { title: 'Task', value: 'task' },
+  { title: 'Resources', value: 'resources' },
+]
+const workers = ref<Worker[]>([])
+const paginator = ref<Paginator>({
```
benoit74: I might miss something but I don't get why we define yet another paginator here, while we already have one in the store. I probably miss something, but it looks like there are too many.

frontend-ui/src/views/WorkersView.vue
```diff
@@ -148,41 +111,53 @@ const loadingStore = useLoadingStore()
...
+  page: 1,
+  limit: Number($cookies?.get('workers-table-limit')) || 20,
```
benoit74: 20 is not in the list of options on this screen, see other comment ; but it should be, so that's probably fine.

frontend-ui/src/components/WorkersTable.vue
elfkuzco: Yeah. Forgot this while testing because my only online worker was not showing up and I didn't want to go to 50. Will fix

frontend-ui/src/stores/workers.ts
elfkuzco: Was mimicking the old code: https://github.com/openzim/zimfarm/blob/a45b8c32a35ef55e6aad87cf85478dcd0bbe3e5d/dispatcher/frontend-ui/src/views/WorkersView.vue#L223-L231

elfkuzco: Do you want to revert to 20?

frontend-ui/src/stores/workers.ts
elfkuzco: I tried to make it the same page size as the limit being used by the old code
https://github.com/openzim/zimfarm/blob/a45b8c32a35ef55e6aad87cf85478dcd0bbe3e5d/dispatcher/frontend-ui/src/views/WorkersView.vue#L223-L231

frontend-ui/src/stores/workers.ts
benoit74: Yes, this was a kind of hack because UI was not handling pagination at all. Now that it handles pagination, we can simplify / streamline

frontend-ui/src/stores/workers.ts
benoit74: Again, this was a hack, no need to keep it

elfkuzco:
not entirely sure on what the small screen would look like. do you want some columsn to be hidden on small screens?

elfkuzco:
Regarding the small screen design, there's not much i can do on it for now. Not really proficient with UI and struggling a bit to come up with view that works fine. Think I will do it in a new PR

benoit74:
I was more thinking about having two rows per worker instead of one, but this means it is not a datatable anymore indeed. Let's keep it for another PR.

elfkuzco:
okay. should i squash and merge this?

benoit74 (approved):
LGTM, please squash and merge at your convenience

PR #1255, opened 2025-09-01
Add gutenberg recipes management

elfkuzco:
## Rationale
Port of 5265e6d with some subtle differences like:
- use pycountry to return ISO 639-3 code of language as website returns ISO-639-1 language codes.


elfkuzco:
ported 9ff19dd too so both commits mirro #1118

benoit74 (approved):
LGTM

PR #1256, opened 2025-09-01
add gutenberg platform

elfkuzco:
## Rationale
port of 7091469 and #1120 from main to zimfarm-upgrade

### DIFF
--- a/backend/src/zimfarm_backend/common/enums.py
+++ b/backend/src/zimfarm_backend/common/enums.py
@@ -301,6 +301,7 @@ class Platform(StrEnum):
     shamela = "shamela"
     libretexts = "libretexts"
     phet = "phet"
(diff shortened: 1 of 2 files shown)


backend/src/zimfarm_backend/common/enums.py
```diff
@@ -301,6 +301,7 @@ class Platform(StrEnum):
     shamela = "shamela"
     libretexts = "libretexts"
     phet = "phet"
+    gutenberg = "gutenberg"
 
     @classmethod
     def all(cls) -> list["Platform"]:
```
benoit74, line 307: Shouldn't this also return `cls.gutenberg`?

elfkuzco: Omitted by accident. Fixed now.

PR #1263, opened 2025-09-02
use jsonb query expression to filter timestamp in periodic tasks

elfkuzco:
## Rationale

Owing to the change of the `timestamp` field to `list[tuple[str, Any]]`, the periodic tasks filter expression needed to be
updated to use jsonpath to extract the `$date` of the tasks that match the status and datetime criteria for `status_cancel`


benoit74 (approved):
Nice catch!

PR #1264, opened 2025-09-03
remove zimcheck from receiver

### LINKED ISSUE #655: Zimcheck should be removed from receiver
we do it in the worker, not needed anymore there.

elfkuzco:
## Rationale
Remove `zimcheck` from receiver as it is done in worker. Receiver now moves files only to uploader.

## Changes
- remove zim-tools in Dockerfile since it's the provider of the `zimcheck` CLI
- remove `zimcheck` specific options as they are no longer needed

This fixes #655

### DIFF
--- a/receiver/Dockerfile
+++ b/receiver/Dockerfile
@@ -48,18 +48,11 @@ VOLUME /mnt/quarantine
 VOLUME /mnt/zim
 VOLUME /mnt/check_logs
 
-ENV ZIM_TOOLS_URL https://download.openzim.org/release/zim-tools/zim-tools_linux-x86_64-3.6.0.tar.gz
 ENV ZIM_SRC_DIR /jail/zim
 ENV ZIM_DST_DIR /mnt/zim
 ENV ZIM_QUAR_DIR /mnt/quarantine
-ENV VALIDATION_LOG_DIR /mnt/check_logs
-ENV ZIMCHECK_OPTION -A
 ENV ZIMCHECK_PARALLEL_JOBS 2
 
-RUN wget -nv -O zim-tools.tar.gz -q $ZIM_TOOLS_URL && \
-  tar -xzf zim-tools.tar.gz && \
-  cp zim-tools*/* /usr/local/bin/ && \
-  rm -rf zim-tools*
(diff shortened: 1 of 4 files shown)


benoit74 (changes requested):
Thank you, well done.

Please also (sorry for not having specified this earlier, but reviewing the code makes me realize we probably want to simplify further):
- confirm that `/jail/logs` is absolutely unused and can be removed ; on my end I checked on server and folder is absolutely empty ; if confirm, please clean this situation as well
- remove the `VALIDATION_OPTION` argument, i.e. remove the ability to have the `NO_QUARANTINE` option ; we don't use / need this option ; corresponding code can be removed as well
- fix README: currently it is still mentioning removed folders, zimcheck, ... might be useful to explain somewhere that the `check` step used to run `zimcheck` but is not anymore, and now it just checks that ZIMs are in appropriate subfolder and moves them to final destination with proper permissions

receiver/README.md
```diff
@@ -22,57 +19,36 @@ docker run -p 1622:22 \
...
--R , --redundant       Redundant data check
--U , --url_internal    URL check - Internal URLs
--X , --url_external    URL check - Internal URLs
--E , --mime            MIME checks
--D , --details         Details of error
-```
+- The ZIMs dir to check : `-v <YOUR_ZIM_DIRECTORY_TO_CHECK>:/jail/zim`
+- The valids ZIMs dir : `-v <YOUR_ZIM_DIRECTORY>:/mnt/zim`
+- The quarantine dir : `-v <YOUR_ZIM_QUARANTINE_DIRECTORY>:/mnt/quarantine`
+- The log dir : `-v <YOUR_LOG_DIRECTORY>:/mnt/check_logs`
```
benoit74: We do not have check logs anymore.

receiver/README.md
```diff
@@ -22,57 +19,36 @@ docker run -p 1622:22 \
...
 
 ## Warehouse Keeper
```
benoit74: We do not have log files anymore, so you can remove this whole `Warehouse keepr` section IMHO, or I miss what this is about.

PR #1265, opened 2025-09-03
log contents of task directory

elfkuzco:
## Rationale
Log contents of task directory so we can check whether something was created

## Changes
- log contents of task directory before handling files. 

This complements #1045 to complete #723


benoit74 (approved):
LGTM

I originally understood we wanted to log output folder content only when there was no ZIM produced at all, but in fact I like the idea of doing this unconditionally since we do not expect this to produce much logs on a normal basis and it might help spot other problems.

Issue #886, opened 2023-12-12
Move offliners definitions (or at least flags descriptions) to a data-driven source


benoit74:
- Location: API

## Situation
Currently, the offliners definitions are placed in the [source code](https://github.com/openzim/zimfarm/tree/main/dispatcher/backend/src/common/schemas/offliners), including the description displayed for every offliners flags.

This makes their edition pretty complex (code change, code review, deployment) even if all this is as automated as possible.

It is particularly true for flag description which should be able to be updated "fast".

We (content team) wonder if we should move this to somewhere else (a file somewhere, or a DB table with UI screens and special user role, ...).

Let's discuss about this.

benoit74:
Remark from @kelson42 (live in a random meeting): priority is probably to start with flag description only, the rest of the offliner definition is obviously more complex to transfer to a data driven source

benoit74:
See https://github.com/openzim/zimit-frontend/issues/46, localization need should maybe be taken into account when working on this issue

kelson42:
Raisinf prio and put in project Zimit2.

kelson42:
@benoit74 @rgaudin What about proposing a JSON format where the necessary information should be defined by the scraper. This data would be then indgested by both the scraper and the Zimfarm to move forward?

benoit74:
I don't get where you want to put this JSON file and what it will contain.

kelson42:
> I don't get where you want to put this JSON file and what it will contain.

* "Defined by the scraper", so in the scraper repository
* "necessary information", whatever is needed (description, name, constraints on the value, etc..)

benoit74:
Yes, it could work.

I however wonder if adding all scrapers as Zimfarm dependencies wouldn't make more sense. 

With a plugin approach (i.e. dependencies are not here at build time but discovered live at runtime when the Zimfarm starts). 

This would be a great enabler for https://github.com/openzim/zimfarm/issues/891 where we might want to run some scraper code in Zimfarm backend to check for recipe validity / capability for the scraper to proceed with the task.

kelson42:
> I however wonder if adding all scrapers as Zimfarm dependencies wouldn't make more sense.

I have no strong opinion.

My proposal has the advantage to be pretty light. And Yes, this is intend to fix https://github.com/openzim/zimfarm/issues/891 too. You just need to define the format and the indgester/parser. The very limitation I see is that you can not put logic in it. 

[...]

PR #1266, opened 2025-09-03
build offliner schemas from json files

elfkuzco:
## Rationale
By dynamically building offliner schemas from json files, this PR lays the groundwork to define offliner flags and metadata
in data files, thus, allowing them to be defined in one place (TBD) and reduce redundancy of keeping definitions in sync
between scraper and zimfarm

## Changes
- build offliner schema from json files


elfkuzco:
A few comments regarding the json schema:
- `title` and `description` are required
- `min` and `max` should only be set for numeric types like `int`, `float`, etc
- `min_length` and `max_length` are for `strings`
- `pattern`s typically should be of Perl Compatible Regular Expressions as it could be consumed by different programming languages
- `choices` can either be a list of `{"title": ..., "value": ...}` or simply be  a list of strings. For cases where it is provided as a list of `{"title": ..., "value": ...}`, the `title` should be generally acceptable as an identifier across programming lanauges as it is used in constructing of the name of the enum member. If it is simply a list of strings, each string should satisfy the same constraint.
- The keys of the JSON schema should be generally acceptable as identifiers too as they constitute the parameters when constructing the `Offliner`. Provide an `alias` instead if a different name should be used.

backend/src/zimfarm_backend/common/schemas/offliners/mwoffliner.json
```diff
@@ -0,0 +1,268 @@
...
+    "type": "string",
+    "required": false,
+    "title": "OS Temp Dir",
+    "description": "Override default operating system temporary directory path environnement variable"
+  },
+  "outputDirectory": {
+    "type": "string",
+    "required": false,
+    "title": "Output folder",
+    "description": "Output folder for ZIM file or build folder. Leave it as `/output`",
```
benoit74: This is typically where we realized that description on Zimfarm is different than description on scraper CLI. In the Zimfarm, we need this to be `/output` for proper operation ; on the scraper, it can be whatever the user needs it to be.

elfkuzco: Okay. This means, I would have to remove the `pattern` constraint on it too. Or is it safe to assume that other consumers of the JSON file (like the scrapers) can disregard the pattern?

benoit74:
And final comment regarding custom validation rules at Zimfarm level for some parameters: I feel like we can simply add a list of these custom validations rules in the JSON file, and automatically add them when parsing the JSON and creation the pydantic schema at Zimfarm level. WDYT?

elfkuzco:
> how do we handle the fact that on Zimfarm we need both a title and a description while on scraper CLI we have only one description? do we assume we will concatenate both in scraper CLI with something like $title: $description? Or we will use only the description (which must then be self explanatory, not always the case on Zimfarm description currently).

I think we should use the same `description` on both scrapers and zimfarm. Certain fields like `title`, `pattern`, etc would only be meaningful for consumers like zimfarm. This would solve situations where the `output` description would have a description that does not contain `Leave it as output` but have a `pattern` to enforce it.

Another alternative would be to remove the `pattern` and enforce them at model creation. For example, I could register validators like
```py
validators = {
    'output_validator': field_validator('output')(validate_output)  
    }
    
    create_model(...., __validators__=validators
  ```
  
  But I wouldn't be able to extract the pattern and share to the frontend/API clients.
  
  Essentially, the scraper needs to care about a subset of the fields for each parameter.

**Comment by elfkuzco** [comment]
> Another question: how do we plan to share this JSON file between scraper and Zimfarm? I feel like having a copy in both repo is fine since it is way simpler to handle and allows to decouple release. It comes at the price of forgetting to update Zimfarm file when scraper is released, but it does not seems to be a very big deal.

I thought the major benefit of the PR was to avoid maintaining the two copies in sync. If that is not a benefit, then, I think leaving it as it is without this PR is essentially the same. Except that I have lost type checks on property access.
Perhaps, there would be a central source (maybe an API endpoint) where these files are hosted and retrieved by both scrapers and zimfarm.

**Comment by benoit74** [comment]
> Essentially, the scraper needs to care about a subset of the fields for each parameter.

I like the idea of forcing descriptions to be valid for both scraper and Zimfarm and using validators to refuse values which are not appropriate in Zimfarm context. This somehow solves my concern about `articleLists` on mwoffliner: description will be "too broad" for Zimfarm, but validator will refuse inappropriate values.

> Perhaps, there would be a central source (maybe an API endpoint) where these files are hosted and retrieved by both scrapers and zimfarm.

I was originally kinda concerned by the fact that scrapers needs to reach an online endpoint to start operation, but in fact this is kinda non-relevant, all scrapers needs to reach online endpoints to operate anyway. This is just "one more". 

It however brings two concerns:
- this endpoint must be "highly available" in the sense that it is a bit sad a scraper does not starts only because it misses its parameters definition
- how do we ensure that the online definition is aligned with scraper version (some might still want to use an old version, needing the old version)

Perhaps we can:
- store the definition in the scraper repo / codebase, so that it is easy to maintain in sync and scraper will not require the online endpoint
- at release time, in the CI, publish the file somewhere online for Zimfarm consumption ; keeping all released versions definition + a "dev" one which gets updated at every push to main branch and is used as fallback when image version has no definition (custom branch, ...)

**Comment by benoit74** [comment]
And I feel like keeping the `pattern` parameter is mandatory for better frontend/UI operation ; when I spoke about having a list of custom validators it was more about very custom ones (e.g. the fact that we do not want a local file path in `articleList` of `mwoffliner`)

**Comment by elfkuzco** [comment]
yes, i think keeping it the files at the scraper level and using CI to release versions of the files.
given each scraper would probably have a version, would the zimfarm always be requesting the `latest` or `stable` version or will we have a way for pinning versions? for example, `mwoffliner` might be on `1.15` while `phet` is on `3.1`. or is this out of the question?

**Comment by elfkuzco** [comment]
> And I feel like keeping the pattern parameter is mandatory for better frontend/UI operation ; when I spoke about having a list of custom validators it was more about very custom ones (e.g. the fact that we do not want a local file path in articleList of mwoffliner)

yes, i can inject custom validators at creation time in addition to those that can be extracted like `patterns`. These kinds of validators would be functional and might (not sure) need to know the name of the field i am validating against.

**Comment by benoit74** [comment]
> given each scraper would probably have a version, would the zimfarm always be requesting the latest or stable version or will we have a way for pinning versions? for example, mwoffliner might be on 1.15 while phet is on 3.1. or is this out of the question?

Would it be feasible and reasonable to have one pydantic model per scraper version, and select the appropriate one based on Docker image version of current recipe / task?

**Comment by elfkuzco** [comment]
> Would it be feasible and reasonable to have one pydantic model per scraper version, and select the appropriate one based on Docker image version of current recipe / task?

not really sure if that would work given the models are imported in order to define the `discriminator` field. Even if that were possible, I wonder how we would know which version to build. Unless we are going to set the version of each offliner as env vars and fetch that file version before starting

**Comment by elfkuzco** [comment]
> > Would it be feasible and reasonable to have one pydantic model per scraper version, and select the appropriate one based on Docker image version of current recipe / task?


 not really sure if that would work given the models are imported in order to define the `discriminator` field. Even if that were possible, I wonder how we would know which version to build. Unless we are going to set the version of each offliner as env vars and fetch that file version before starting. But that would only be one pydantic model.

**Comment by benoit74** [comment]
Just discussed with @elfkuzco and he suggested we might use scraper CIs to trigger a commit on Zimfarm repo everything JSON file is modified

@kelson42 I think we should discuss this live soon to check we are on the right path

**Comment by benoit74** [comment]
From discussions with kelson42:
- we are moving in the right direction
- the fact that scrapers are updating something in Zimfarm when CI runs seems to be correct
- we should probably rather store these definitions in DB (in JSON format)
- we need to have a system coping with multiple scraper version

More on this last point: it is now becoming clear that we must store in a recipe which scraper version this recipe is intended to use, and adapt flags (and their definitions) accordingly. We regularly wanna test `dev` version in the Zimfarm and are every now and then blocked by a missing flag / wrong definition (see https://github.com/openzim/zimfarm/issues/886). Not being to be warned of a problem when downgrading to previous worker version is also a problem.

@elfkuzco can you please have a second look at what it would take to be able to have a dynamic type definition **and** a dynamic list of allowed types? We are very open to any change needed to make this work.

I feel like if we create the `OfflinerSchema` type dynamically with `create_model` as well, it might work. Most probably static typing tools like pyright will not be able to validate anything ... but I don't mind much since on this field we precisely do not want static but dynamic typing.

**Comment by elfkuzco** [comment]
I have created a table to store the offliner definitions. The table has the following columns:
- offliner
- definition
- version
- created_at

Also, I have created a foreign key from the schedule, task and requested task to this new table. This ensures that whenever any of the resource is finished, I know which version to retrieve and class schema to build at runtime. 

Right now, the PR should not be merged but can be reviewed. At least till I have the frontend set up. As it is, it fetches all offliner definitions on load. But this can no longer be done as we need to know which schedule to fetch first, retrieve it's definition before getting the definition.

Created migration too to seed the database with the initial version for all resources that are linked.

File: backend/src/zimfarm_backend/common/schemas/offliners/serializer.py [hunk 2]
```diff
@@ -217,7 +212,7 @@ def schema_to_flags(schema_class: type[BaseModel]) -> list[Flag]:
             data_key=field_alias,
             description=description,
             key=field_alias,
-            label=title,
+            title=capitalize_words(title),
```
benoit74: Why moving from `label` to `title` and why do we need the `capitalize_words` function? I feel like this is correct but it does deserve an explanation.

backend/src/zimfarm_backend/common/schemas/fields.py
```diff
@@ -67,8 +58,10 @@ def not_empty(value: str) -> str:
     return value
 
 
-def validate_language_code(value: str, info: ValidationInfo) -> str:
+def validate_language_code(value: str | None, info: ValidationInfo) -> str | None:
```
benoit74, line 61: Why do we suddenly need to cope with `None` values?

backend/src/zimfarm_backend/db/offliner_definition.py
```diff
@@ -0,0 +1,161 @@
...
+        offliner=offliner_definition.offliner,
+        version=offliner_definition.version,
+        created_at=offliner_definition.created_at,
+        definition=OfflinerFlagSchema.model_validate(offliner_definition.definition),
+    )
+
+
+def create_offliner_definition(
+    session: Session,
+    definition: OfflinerFlagSchema,
```
benoit74: This `definition: OfflinerFlagSchema` argument makes me realize we probably have something wrong here in terms of naming.

I feel like we should find another name for both the field and the class.

What about switching from `definition` to `schema` (both in DB and in pydantic Schemas), and from `OfflinerFlagSchema` to `OfflinerSchema` (it can and will contain more than only flags at some point)

The `schema` is this dictionary. The `definition` is the whole object.

Or any other name, but reusing `definition` here makes things a bit confusing in the code.

backend/src/zimfarm_backend/migrations/versions/12096716adfd_add_offliner_definitions.py
```diff
@@ -0,0 +1,167 @@
...
+    op.add_column("task", sa.Column("offliner_definition_id", sa.Uuid(), nullable=True))
+    op.create_foreign_key(
+        op.f("fk_task_offliner_definition_id_offliner_definition"),
+        "task",
+        "offliner_definition",
+        ["offliner_definition_id"],
+        ["id"],
+    )
+
+    data_dir = Path(__file__).parent.parent / "seed"
```
benoit74: Maybe rename folder from `seed` to `initial_offliner_definitions`

backend/tests/test_validators.py
```diff
@@ -745,72 +732,3 @@ def test_skipable_url_model_skip_validation(
         SecretUrlModel.model_validate(
             {"value": value}, context={"skip_validation": True}
         )
-
```
benoit74, line 748: Where do we test this "slug validation" logic now?

benoit74 (commented):
Very well done, we are really on the good path. As you said, it must not be merged yet. I did not yet tested much the API behavior, but from the fact that you mostly left tests intact, it seems to be quite OK. I will test it a bit more once UI will be ready.

I do have some "little" concerns which needs to be addressed to be able to consider that work is really completed. I feel like it might be better to do it in later PRs, so that we can better document the decisions made.

I will only list concerns below for now. If needed, I do have ideas about how to address them (but again, it might be better to discuss this in a dedicated issue). Only first point is kinda important to solve soon, other point are a bit more "far-fetched".

- changes for nautilus and zimit relaxed schemas + specific validators on `xxx_schema_creator` (e.g. `lang_validator` for kolibri) + specific logic in `command_for`: I'm concerned about the fact that should flag name change, we will have to update Zimfarm code ; we need to find a solution to have this in offliner definition as well
- `offliner_id` field: now that we do not need it as a discriminator anymore, I feel like we moving it back to a "very top" level would help a lot ; but this is another breaking API change
- `xxx_schema_creator`s: I think we could move this into a specific new DB table as well, so that this is very extensible

backend/src/zimfarm_backend/common/schemas/offliners/serializer.py
elfkuzco: I used `title` at the application level so it closely mirrors the `FlagSchema` itself. The UI/clients expect `label` and I used a `serialization_alias` for that. As far as they are concerned, nothing changed.

Regarding the function, while editing the JSON files, I realized I forgot to make the `title` uniform on the UI. There are some places where words start with small letter and I figured this function would be nice to transform cases like this so things look nice.

elfkuzco:
> changes for nautilus and zimit relaxed schemas + specific validators on xxx_schema_creator (e.g. lang_validator for kolibri) + specific logic in command_for: I'm concerned about the fact that should flag name change, we will have to update Zimfarm code ; we need to find a solution to have this in offliner definition as well


I am a bit short of ideas on how to do this in the offliner definition. Even the simplest approach I can think of touches the code too. It involved using `if..else` blocks to update validators because a field might exist on a version and be missing in another. Currently, it would throw an error.


Having it in the offliner definition would mean injecting Python code into the definitions and probably using `eval` against it. But I think this is quite hacky and would rely on the author of the JSON files to know the zimfarm code. If this is automatically stored in the DB from CI, I reckon it would be even more hackier to get right.

WDYT? Do we need to have a meeting on it?

backend/src/zimfarm_backend/common/schemas/fields.py
elfkuzco: Before we moved to this approach, the validation was enforced at the type level. For example, I had:
```py
ZIMLangCode = Annotated[
    str,
    Field(min_length=3, max_length=3),
    WrapValidator(skip_validation),
    AfterValidator(validate_language_code),
]


OptionalZIMLangCode = ZIMLangCode | None
```

The validator would never be called when the field is None because `None` has no validator. The validators were defined on the non-None types.

Now that we have moved to this approach, I only have some subset of types: `string`, `integer`, `float`, etc. And some of these fields are not required. So, this meant that the type would most likely be `string | None`. Thankfully, pydantic allows me to add some base constraints like `pattern`, `min_length` etc while creating the field for a particular flag. So, it won't call such inbuilt validations because those are None. Same thing with our concrete types.

Now, the issue is that we can't define functional/model validators in the JSON schema. And those would always be called by Pydantic. It's up to the author (me) to handle the case that these fields could be nullable and proceed accordingly. Essentially, I am having to do the same thing core types and pydantic `Field` do. Just more verbose.

backend/src/zimfarm_backend/db/offliner_definition.py
elfkuzco: Okay

backend/src/zimfarm_backend/migrations/versions/12096716adfd_add_offliner_definitions.py
elfkuzco: Okay

backend/tests/test_validators.py
elfkuzco: The slug type no longer exists and isn't used anymore. It was just a facade behind a regex. I suppose could test the regex instead. It is used only in relaxed versions of the offliners to replace stricter patterns.

elfkuzco:
> offliner_id field: now that we do not need it as a discriminator anymore, I feel like we moving it back to a "very top" level would help a lot ; but this is another breaking API change

I think it's used in a lot of more places too and it wasn't just serving as a discriminator. Maybe we could this in a later PR?

elfkuzco:
> specific logic in command_for

I think i could use `hasattr` checks before proceeding to read an attribute. Clueless on how to do it on the definition. 

Should we have python logic in the JSON files?

elfkuzco:
> xxx_schema_creators: I think we could move this into a specific new DB table as well, so that this is very extensible

Not sure how to do this either. These are function calls.

elfkuzco:
>  I did not yet tested much the API behavior, but from the fact that you mostly left tests intact, it seems to be quite OK. I will test it a bit more once UI will be ready.

UI is now ready. Applied a fix for it. It doesn't have the features of selecting a version, etc. It just updates the API calls and things are pretty much the same. New features could be done in a later PR.

backend/src/zimfarm_backend/common/schemas/offliners/serializer.py
benoit74: Indeed, thank you.

backend/src/zimfarm_backend/common/schemas/fields.py
benoit74: Got it, thank you

backend/src/zimfarm_backend/migrations/versions/12096716adfd_add_offliner_definitions.py
elfkuzco: Done

backend/src/zimfarm_backend/db/offliner_definition.py
elfkuzco: Done.

benoit74:
> changes for nautilus and zimit relaxed schemas 

What about introducting an optional `relaxed_pattern` field like this in the definition:

```json
    "zim_file": {
      "type": "string",
      "required": false,
      "title": "ZIM filename",
      "description": "ZIM file name (based on --name if not provided). Include {period} to insert date period dynamically",
      "alias": "zim-file",
      "pattern": "^([a-z0-9\\-\\.]+_)([a-z\\-]+_)([a-z0-9\\-\\.]+_)([a-z0-9\\-\\.]+_|)([\\d]{4}-[\\d]{2}|\\{period\\}).zim$",
      "relaxed_pattern": ""
    }
```

And then in `generate_field_type`, we check if "xxx_USE_RELAXED_SCHEMA" env variable is set, and we switch accordingly between "pattern" and "relaxed_pattern" value. Probably means we need to automatically build a  `OFFLINERS_WITH_RELAXED_SCHEMAS` set from the "xxx_USE_RELAXED_SCHEMAS" environment variables (assuming all offliner_id are lowercase).

Something like this:

```py
    pydantic_field = pydantic_field_cls(
        title=flag.title,
        description=flag.description,
        ge=flag.min,
        le=flag.max,
        pattern=flag.relaxed_pattern if flag.relaxed_pattern and offliner_id in OFFLINERS_WITH_RELAXED_SCHEMAS else flag.pattern,
        min_length=flag.min_length,
        max_length=flag.max_length,
        alias=flag.alias,
    )
```

And we could even extend to `relaxed_min`, `relaxed_max`, `relaxed_min_length` and `releaxed_max_length` quite easily (maybe worth to do it straight away).

> specific validators on xxx_schema_creator (e.g. lang_validator for kolibri) 

What about extending the schema like this:

```json
    "lang": {
      "type": "string",
      "required": false,
      "title": "Language",
      "description": "ISO-639-3 (3 chars) language code of content. If unspecified, will attempt to detect from main page, or use 'eng'",
      "minLength": 3,
      "maxLength": 3,
      "custom_validator": "language_code"
    },
```

and then in `build_offliner_model` have some logic to compute the validators:

```py

def get_custom_validator(validator_name: str):
    match validator_name:
        case "language_code":
            return validate_language_code
        case "...":
            return ...
        case _:
            raise ValueError("...")
        
def build_offliner_model(
    *,
    model_name: str,
    offliner_id: str,
    schema: OfflinerFlagSchema,
    base_model_cls: type[BaseModel] | type[DashModel] | type[CamelModel],
    validators: dict[str, Callable[[Any, ValidationInfo], Any]] | None = None,
):
    fields: dict[str, Annotated[Any, Any]] = {}
    for flag_name, flag_schema in schema.flags.items():
        fields[flag_name] = generate_field_type(flag_schema, flag_name)
    # Add the offliner_id field to the model with the base model class which
    # typically defines the alias generators and necessary config for the model
    return create_model(
        model_name,
        offliner_id=(Literal[offliner_id], Field(alias="offliner_id")),
        __base__=base_model_cls,
        __validators__= {f"{flag}_validator": get_custom_validator(flag_schema.custom_validator) for flag, flag_schema in schema.flags.items() if flag_schema.custom_validator},
        **fields,
    )
```

This means we will have a limited set of custom validators, and we must ensure they are unique to the whole Zimfarm, but this is probably fine. Should we need new custom validation logic for a specific scraper, we will have to implement it in the Zimfarm. But at least, we can change flag name in scraper without any consequence at Zimfarm level.

> specific logic in command_for

I do not have a solution for gutenberg `one_language_one_zim` flag. Maybe let's keep this aside (i.e. keep it as-is) for now. This is a very specific and dirty edge case anyway.

For `publisher`,  `std_stats` and `std_output`, we can simply add optional properties to mark flag to use:

```json
    "publisher": {
      "type": "string",
      "required": false,
      "title": "Publisher",
      "description": "Custom publisher name (ZIM metadata). openZIM otherwise",
      "is_publisher": true
    },
    "outputDirectory": {
      "type": "string",
      "required": false,
      "title": "Output folder",
      "description": "Output folder for ZIM file or build folder. Leave it as `/output`",
      "pattern": "^/output$",
      "is_output": true
    },
    "stats_filename": {
      "type": "string",
      "required": false,
      "title": "Stats filename",
      "description": "Scraping progress file. Leave it as `/output/task_progress.json`",
      "pattern": "^/output/task_progress\\.json$",
      "is_stats": true
    }
```

For other ones needed default value (if not set by Zimfarm user) or forced value (always set to a given value no matter what user might input) I feel like we could extend the schema with two optional properties: `default_value` and `force_value`. This is probably the part where I'm less confident because these values looks a lot like a hack. But we can probably live with this for now. For sotoki this would give something like this:

```json
    "mirror": {
      "type": "url",
      "required": false,
      "title": "Mirror",
      "description": "URL from which to download compressed XML dumps",
      "secret": true,
      "default_value": "https://s3.us-west-1.wasabisys.com/org-kiwix-stackexchange"
    },
    "redis_url": {
      "type": "string",
      "required": false,
      "title": "Redis URL",
      "description": "Redis URL to use as database. Keep it as unix:///var/run/redis.sock",
      "force_value": "unix:///var/run/redis.sock"
    }
```

> xxx_schema_creators: I think we could move this into a specific new DB table as well, so that this is very extensible

Could we add an `Offliner` DB table with something like this (meaning we drop current `Offliner` class):

```
class Offliner(Base):
    __tablename__ = "offliner"
    offliner: Mapped[str] = mapped_column(
        init=False, primary_key=True)
    )
    base_model: Mapped[str]
    docker_image_name: Mapped[str]
```

And use this to create the offliner:

```py
def get_base_model_cls(offliner: Offliner) -> type[BaseModel]:
    match offliner.base_model:
        case "DashModel":
            return DashModel
        case "CamelModel":
            return CamelModel
        case _:
            raise ValueError("...")
        
def create_offliner(
    *,
    offliner: Offliner,
    offliner_definition: OfflinerDefinitionSchema | OfflinerDefinition,
    data: dict[str, Any],
    skip_validation: bool = True,
) -> BaseModel:
    """Create the offliner instance from the offliner definition and data"""
    if isinstance(offliner_definition, OfflinerDefinition):
        offliner_definition = create_offliner_definition_schema(offliner_definition)
    return build_offliner_model(
        model_name=f"{offliner_definition.offliner}FlagsSchema",
        offliner_id=offliner_definition.offliner,
        base_model_cls = get_base_model_cls(offliner),
        schema=offliner_definition.definition, 
    ).model_validate(data, context={"skip_validation": skip_validation})
```

Since we reuse this `Offliner` class everywhere, I think we should probably load all offliners at application startup (meaning we will have to reload application when we create a new offliner in DB ... this seems acceptable).

> offliner_id field: now that we do not need it as a discriminator anymore, I feel like we moving it back to a "very top" level would help a lot ; but this is another breaking API change
> I think it's used in a lot of more places too and it wasn't just serving as a discriminator. Maybe we could this in a later PR?

I'm fine with this, especially since it is a breaking change. We can probably live with current situation for a while. I will open an issue to track this idea / question.

elfkuzco:
> Since we reuse this Offliner class everywhere, I think we should probably load all offliners at application startup (meaning we will have to reload application when we create a new offliner in DB ... this seems acceptable).

The `Offliner` class is an enum and while some of its helpers could be refactored as functions that take an argument, It is imported across functions too and used in places like https://github.com/openzim/zimfarm/blob/ef91be84720e4a60936a7482261859f1717b6803/backend/src/zimfarm_backend/utils/offliners.py#L29-L45

How do we handle  cases like this?
Also, you mention:
> For publisher, std_stats and std_output, we can simply add optional properties to mark flag to use:

If we allow the schema to have fields like 'is_publisher`, `is_std_stats`, etc.., I don't know how it could be used. What we are using the data for the fields is to build a pydantic `Field` and extra params like this are probably going to be excluded or raise a TypeError. 

One way I can think of is to extend the table to have columns like `std_stats`, `std_output`, `cmd` and use the data from there in the `command_for`

What do you think?

benoit74:
`is_publisher`, `is_output` and `is_stats` must be "officially" defined in the `Flag` model.

`OFFLINER_DEFS` would be removed once we have have these properties directly on the flags. And this seems to be the correct way to do it because having it defined per offliner is wrong (like it was a simplification), this might change over the course of scraper changes. And having to modify Zimfarm when offliner became capable to write a stats file (for instance) is wrong, this information should come automatically with the offliner definition.

I already had a look at other places where we used this `Offliner` enum and I don't see any problematic hard to solve. When we adjust mount point for phet or docker config for zimit ; this would be way better in the offliner definition file (as high level config next to the flags) rather than directly in the code.

elfkuzco:
> is_publisher, is_output and is_stats must be "officially" defined in the Flag model

Okay. When I build up the field using `Field(is_publisher, is_output, is_stats, ....)`, this information is going to be lost. I can't do `offliner.config.<field>.is_publisher` because these values are not set by pydantic on the `Field`. 

Unless, I would have to use the definition while building the `command`.

benoit74:
> Unless, I would have to use the definition while building the command.

What would be the downside of having access to the definition while building the command ? Looking quickly at the code, we build the command only in `expanded_config`, and when we call `expanded_config` it looks like we always already have access to the definition, so it looks like we only have to pass the offliner definition object down together with current config.

elfkuzco:
- Added `stdOut` and `stdStats` to top level offliner definition to mirror the previous `OfflinerDefinition`
- Added `isPublisher` at the flag level. Used to determine which field is the publisher. Provided reasonable defaults
- Add Offliner model and remove corresponding enum
- Use map to prevent hitting the db to retrieve an offliner as it's used across a lot of places and it's likely going to be static data
- Created migration to load the DB with offliners

elfkuzco:
Need to improve coverage...

backend/src/zimfarm_backend/common/enums.py
```diff
@@ -215,68 +215,68 @@ def all(cls) -> set[str]:
...
-            cls.openedx: DockerImageName.openedx,
-            cls.zimit: DockerImageName.zimit,
-            cls.kolibri: DockerImageName.kolibri,
-            cls.wikihow: DockerImageName.wikihow,
-            cls.ifixit: DockerImageName.ifixit,
-            cls.freecodecamp: DockerImageName.freecodecamp,
-            cls.devdocs: DockerImageName.devdocs,
-            cls.mindtouch: DockerImageName.mindtouch,
-        }.get(offliner, "-")
+# class Offliner(StrEnum):
```
benoit74: Looks like you forgot to delete this 😄

benoit74 (changes requested):
Thank you very much, I really like what you've achieve to build, this is definitely going to be a game changer. And to be honest, it really looks like it is working like a charm.

I feel like we could polish this PR a little bit more by:
- never returning `offliner_definition_id` in ORMs object (`TaskFullSchema`, `RequestedTaskFullSchema` and `ScheduleFullSchema`) but rather `offliner` and `version`, keeping the UUID an internal identifier
- this in turn would allow to get rid of the `GET /offliners/definition/{definition_id}` endpoint

WDYT?

What I see as currently broken / left to do before merging this PR (correct me if I miss something):
- for backward compatibility / transition, we need to maintain a `GET /offliners/{offliner_id}` endpoint: this will return the `initial` definition, allowing zimit-frontend to continue working.
- we cannot change offliner for a given schedule from UI, because it misses the combobox allowing to select version to use ; what about just "forcing" it to "initial" from the UI ; this could be sufficient until we've implemented the full UI
- confirm with two scrapers (1 Node.JS + 1 Python) that they can ingest the JSON file as well for their own purpose (CLI) ; probably mwoffliner for Node.JS ; do you wanna try this or is it better if I do this myself? scrapers are not that a dangerous beast after all ^^

What is left to do to close #886 (probably way better in distinct PRs, or even distinct issues and we close #886):
- implement the "version" combobox in "schedule edit" form
- show the "version" in schedule configuration + task configuration
- decide + implement what we wanna do when a scraper is released: endpoint to create new definition, permission to scope it to one offliner and only definition creation, what about upgrade/downgrade, is it defined in the schema as well and fully automated on release? or do we start with something manual as we currently do? could be linked to version changes from the user in the UI (same migrations are needed)
- decide + implement what we wanna show / edit in the UI in terms of offliner + definition (std_stats, full definition, ... nothing this is just technical data, ...)
- cleanup (drop "initial" defaults in some endpoints used for the transition)

elfkuzco:
> we cannot change offliner for a given schedule from UI, because it misses the combobox allowing to select version to use ; what about just "forcing" it to "initial" from the UI ; this could be sufficient until we've implemented the full UI

already implemnted https://github.com/openzim/zimfarm/blob/f932eb3eee84ba163c11d4731a30fe7a8afa2318/frontend-ui/src/components/ScheduleEditor.vue#L1058-L1062

elfkuzco:
> implement the "version" combobox in "schedule edit" form

created #1379 already for this.

benoit74:
> already implemnted 

This is somehow not working then ...

https://github.com/user-attachments/assets/d0c7d392-b848-4624-83b3-6f3af77a64ef

elfkuzco:
> confirm with two scrapers (1 Node.JS + 1 Python) that they can ingest the JSON file as well for their own purpose (CLI) ; probably mwoffliner for Node.JS ; do you wanna try this or is it better if I do this myself? scrapers are not that a dangerous beast after all ^^

for mwoffliner, it's easy to do as it really needs is that the parser returns the paramter list as it needs. shouldn't be hard to implement.

elfkuzco:
if there's going to be similar parsing for the python parser, it's going to be a bit tricky as it would require some way to tell what the respective actions of `argparse` mean. for example, there's `action=store_true`

elfkuzco:
> confirm with two scrapers (1 Node.JS + 1 Python) that they can ingest the JSON file as well for their own purpose (CLI) ; probably mwoffliner for Node.JS ; do you wanna try this or is it better if I do this myself? scrapers are not that a dangerous beast after all ^^

As per the meeting on Slack, it's simpler to ingest the JSON file in mwoffliner but not so much in python scrapers as those use argparse-specific parameters and behaviours like groups, dest, etc. One way to navigate around this without putting too much information in the JSON file would be to have specific logic at the scraper level to handle this. The zimfarm doesn't really do anything beyond building the arguments from existing flags and any additional metadata/information would not be particularly helpful in how it does it's job.

benoit74:
Yes, I agree that we should probably not mind about the scrapers needing more logic to parse the JSON file for their own purpose. 

The fact that the JSON file will live inside scraper codebase is anyway already a big step forward, and scraper could even not parse it at all it will already help to sync changes in due time. 

Plus zimfarm anyway ignores extra fields, so scrapers are free to enrich the JSON file with "whatever needed" for their own purpose. We do not need to decide this at a central level.

Is this ready for a new review or do you still have open points to address?

elfkuzco:
> Is this ready for a new review or do you still have open points to address?

Finishing up the combo box selection and adding handlers. A bit tricky to get right as I have updates from multiple watchers.

elfkuzco:
- Added combobox to select version
- Added route to list all versions (used in populating combobox)
- Added handlers for when an offliner name changes and when it's version changes. Both cases lead to an update of the flags defintion.
- The watcher for the flags defintion rebuilds the flags by populating attributes that are set on the original schedule to the newer flags. This allows me to select an older version and still have some of the inputs across both versions to be used in populating the inputs as opposed to setting everything as empty. Same thing applies for offliners changed too although it almost always clears everything given offliners don't share fields.

elfkuzco:
Is it necessray to change the version if user changes the docker image/tag for the offliner? Currently, I fetch only on offliner name change.

elfkuzco:
> I finally wonder why we are asking for "Image name" in the UI since the backend is anyway forcing it to one specific value. Do I miss something?

The backend doesn't force it to a specific value. It only validates that the new image name is a valid image for the offliner. Plus, when the image name is changed, we update the tags for it. Do you want it to be removed?

elfkuzco:
Added watchers to synchronize version:
- If version is changed from dropdown, the flags definition is updated and the editFlags is populated with values from the original schedule that have keys in the new definition. This allows one to change to a version and have older fields removed and/or newer fields set to empty automatically. Essentially, a merge.
- If offliner name is changed from dropdown, versions are updated. A watcher which watches for the versions takes one of two branches:
    - If we changed from a "new" offliner to the original offliner, we simply set the version to the original version.
    - If we changed to a "new" offliner, we find the "preferred" version and set it to be the version because we don't know which version to use for the "new" offliner. The preferred version is selected as follows:
        - if the image tag exists in the new list of versions, select it
        - else, check if the original offliner version is in the list of versions and select it
        - otherwise, select the most recent version from the list of new versions
   - After detecting the new version, we fetch it's definitions, update the flags and perform the merge
- If image tag is changed, we check if the new tag is in the list of versions. if it is, we update the version to match it and update the flags

benoit74 (approved):
LGTM

Thank you for this "very big" PR, glad we achieved to make this work.

There is some open point left but I will move this to dedicated issues as already agreed.

PR #1271, opened 2025-09-04
use the same network stack to connect containers in development

elfkuzco:
## Rationale
During development, we often need to create a network as docker compose is not aware of the other containers spun up by the `task worker`. Rather than creating an external network, bind any new containers to use the same network as the running container (which translates to the one set up by `docker compose` originally).
This only applies if environement is set to `development`. Default is `production` which causes the networking to be via the `bridge` adapter (the default)
## Changes
- set up containers and intermediate containers to reuse the same network in development
- default environment is `production` which leaves things as it is


benoit74 (approved):
LGTM

PR #1272, opened 2025-09-04
update ssh keys for receiver and uploader communication

elfkuzco:
## Changes
- change Fingerprint hash function from md5 to sha256 for receiver sshd config (same hash function used by the new API)
-  newer API stores the full ssh in the `key` column. refactor `get_zimfarm_key` in `receiver` to return it as the key for authentication
- sometimes, there might be workers with the same fingerprint (or key) as there is no `unique` constraint. so, we return the `first` entry that matches the fingerprint. `one_or_none` raises a MultipleRows exception

### DIFF
--- a/receiver/apps/get_zimfarm_key.py
+++ b/receiver/apps/get_zimfarm_key.py
@@ -53,10 +53,7 @@ def print_keys_for(username, fingerprint):
     print("\n".join(keys))
 
 
-def fetch_public_keys_for(username, raw_fingerprint):
-    # convert fingerprint to dispatcher's format
-    fingerprint = "".join(raw_fingerprint.split(":")[1:])
-
+def fetch_public_keys_for(username, fingerprint):
     req = requests.get(
         url=f"{environ['ZIMFARM_WEBAPI']}/users/-/keys/{fingerprint}",
         params={"with_permission": ["zim.upload"]},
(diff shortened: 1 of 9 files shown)


receiver/apps/get_zimfarm_key.py
```diff
@@ -70,16 +67,14 @@ def fetch_public_keys_for(username, raw_fingerprint):
...
-        )
+        logger.warning(f"failed login attempt using {fingerprint}: {reason}")
         return
 
     logger.info(
         f"granted login for {response['username']} via {response['type']} key "
         f"{response['name']}"
     )
-    return [f"ssh-rsa {response['key']} {response['name']}"]
+    return [f"{response['key']}"]
```
benoit74, line 77: Do we really need to change this? I feel like previous response was more understandable / maintainable, even if I admit both probably work well

receiver/sshd_config
```diff
@@ -24,7 +24,7 @@
 
 # Logging
 #SyslogFacility AUTH
-#LogLevel INFO
+# LogLevel INFO
```
benoit74: Please revert this change, no need to do this

receiver/apps/get_zimfarm_key.py
elfkuzco: Main caveat with this approach is that it assumes the key will always be `ssh-rsa`. Given we now have support for keys generated with ECDSA algorithm, there could be parts with `ecdsa-sha2-nistp521`. Or even keys with `ssh-ed25519`.

elfkuzco:
created a second commit with the following changes:
- compute `name` on the backend and remove param from request body. remove such logic from frontend
- remove check from frontend which assumes keys will be generated with `rsa` algorithm. we now support `ecdsa` keys
- compute key type on backend
- `key` will contain the key as user sent it which is used by `get_zimfarm_key` on `receiver`. `receiver` also assumed key would always be `rsa` which is not always going to be the case now.

receiver/apps/get_zimfarm_key.py
benoit74: Got it

benoit74 (approved):
LGTM

PR #1274, opened 2025-09-05
expire ssh keys

elfkuzco:
## Rationale

As per NIST recommendations, SSH keys should be rotated once or twice per year. This PR adds features to flag a key as expired if the date which it was added is greater than `SSH_KEY_EXPIRY_DURATION`. This way, workers are forced to delete old keys and create new ones.

## Changes
- add `SSH_KEY_EXPIRY_DURATION` to control when keys are considered as expired
- exclude keys that have "expired" from `db_get_ssh_key_by_fingerprint` used by workers to authenticate with backend
- include has expired flag in users list response


elfkuzco:
Okay. I will close it now.

PR #1278, opened 2025-09-05
force refresh while calling getApiService

elfkuzco:
## Rationale
Vuetify's server side tables trigger before the app's `onMounted` is called. And this causes a call to fetch the requested tasks at which point the token has not been loaded from cookie. To mitigate this, we hide the `RouterView` till the applications' data is loaded. Accordingly, while retrieving axios request objects, we force refresh of token if current one has expired.


elfkuzco:
> And if so, what is the logic for forcing on every API service instantiation and not on App initialization (again, not saying it is wrong but I don't get it).

Nothing really. Just omission. I can add it

elfkuzco:
> A final concern is that I'm not sure about what would prevent the app to renew token many times in parallel when it becomes expired. Since we have many places requiring an API service all at once, they might all realize they have an expired token, each issuing a call to renew it before the one from previous service has completed and updated access token in memory / cookie.

Given there's almost always one major view/component mounted, it's rare there would be duplicate requests unless certain `setInterval` calls from a view are not cleaned up on component unmount. But that's not the case.

benoit74:
> Nothing really. Just omission. I can add it

But then, we would always force renewal, don't we?

PR #1290, opened 2025-09-08
retrieve event notification from task

elfkuzco:
## Changes
- retrieve each event's notifications from the task notification object instead of using the whole notifications object.


benoit74 (approved):
Nice catch!

Issue #1287, opened 2025-09-08
Enhance CI-CD jobs


benoit74:
Currently, we miss:
- `backend-Tests.yml`: should try building Dockerfile and starting it
- `frontend-Tests.yml`, `worker-Tests.yml` (for both manager + task, two distinct jobs is probably OK), `dnscache-Tests.yml`, `receiver-Tests.yml`, `uploader-Tests`, `watcher-Tests`: to add, should try building Dockerfile (and starting it if possible, at least frontend should be OK to start without problem)
- `frontend-QA.yml`: to add, should check prettier + eslint (at least)
- all besides `Publish.yml`: add conditional execution based on paths modifications (i.e. do not run again backend tests if only UI has been modified)

PR #1311, opened 2025-09-10
add precommit hooks for frontend and enhance ci/cd

elfkuzco:
# # Changes
- add prettier and eslint pre-commit hooks
- build docker images of backend, frontend and worker and run ci
- format frontend-ui files

This solves #1287

### DIFF
--- a/frontend-ui/tsconfig.app.json
+++ b/frontend-ui/tsconfig.app.json
@@ -5,6 +5,7 @@
   "compilerOptions": {
     "tsBuildInfoFile": "./node_modules/.tmp/tsconfig.app.tsbuildinfo",
     "types": ["vite/client"],
+    "lib": ["es2022", "dom", "esnext.intl"],
 
     "paths": {
       "@/*": ["./src/*"]
--- a/frontend-ui/yarn.lock
+++ b/frontend-ui/yarn.lock
@@ -2,22 +2,14 @@
 # yarn lockfile v1
 
 
-"@ampproject/remapping@^2.2.0":
-  version "2.3.0"
-  resolved "https://registry.npmjs.org/@ampproject/remapping/-/remapping-2.3.0.tgz"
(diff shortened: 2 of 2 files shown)


elfkuzco:
`prettier` from [mirrors-prettier](https://github.com/pre-commit/mirrors-prettier) seems to be conflicting with the one used on ci with `yarn format-check`. So, I decided to use the sytem one which should be the same as the one installed in the project. Just like pyright uses the system installed one which indirectly comes from `hatch install`

elfkuzco:
Regarding the tests for `watcher`, `uploader`, `receiver` and `dnscache`, I do not know what command would be used to check if the containers are successful. I think one thing we can do is probably create an environment that looks like what we have on `dev/docker-compose.yml` but would run with local builds of the images. Given the worker can be overriden with these images, we can set it to use the local and then test with the `wikipedia_fr_test` file.

WDYT?

benoit74:
As mentioned in issue, I do not expect it to be easy to start all these components Docker image without any context.

For all of them we can check that we achieve to build the Docker image (happened in the past that even the image does not build due to some upstream change typically).

For `watcher` and `uploader` it is probably too hard to test their startup.

For `receiver` and `dnscache`, we can probably check that container achieves to start and is still alive after few seconds. It is not a full test, but at least it checks that did not made some silly error in container which even prevent the container to start (happened in the past).

WDYT? If it is fine for you, please add this as well so that we have a full resolution of #1287 ; thank you

elfkuzco:
add for receiver, watcher, uploader and dnscache. seems something's wrong with the `split-by-grapheme` package and it's types. weird. will resolve

frontend-ui/tsconfig.app.json
```diff
@@ -5,6 +5,7 @@
   "compilerOptions": {
     "tsBuildInfoFile": "./node_modules/.tmp/tsconfig.app.tsbuildinfo",
     "types": ["vite/client"],
+    "lib": ["es2022", "dom", "esnext.intl"],
```
benoit74, line 8: Nice catch ! Not sure we need the `dom`, but it can not harm.

elfkuzco: i get a different class of errors if it isn't there either

PR #1312, opened 2025-09-10
add type for comma-seperated iso 6393 language codes

elfkuzco:
## Changes
Add custom type for comma-seperated iso 639-3 language codes used across multiple offliners


backend/src/zimfarm_backend/common/schemas/fields.py
```diff
@@ -56,6 +56,13 @@ def validate_language_code(value: str, info: ValidationInfo) -> str:
     )
 
 
+def validate_comma_seperated_zim_lang_code(value: str, info: ValidationInfo) -> str:
```
benoit74: Typo sep~~e~~rated => separated (same typo below and type and maybe in other places)

backend/tests/test_validators.py
```diff
@@ -355,3 +360,65 @@ def test_ted_flags_schema_links(links: str, expected: RaisesContext[Exception]):
...
             name="ted-talks_mul_football",
         )
+
+
+@pytest.mark.parametrize(
+    "languages,expected",
+    [
+        pytest.param("eng", does_not_raise(), id="single-valid-language-code"),
+        pytest.param("eng,fra,jpn", does_not_raise(), id="multiple-valid-languages"),
+        pytest.param(
```
benoit74, line 477: You could probably add a test for `eng,fr,jpn` to check that it fails even if only one value is wrong

PR #1314, opened 2025-09-11
fix bug in retrieval of worker resources

elfkuzco:
## Rationale
I introduced a bug while trying to get rid of the old way of setting resources from env vars.
The old code did this:
```py
ZIMFARM_MEMORY = None
try:
    ZIMFARM_MEMORY = as_pos_int(humanfriendly.parse_size(os.getenv("ZIMFARM_MEMORY")))
except Exception:
    physical_mem = psutil.virtual_memory().total
    if ZIMFARM_MEMORY:
        ZIMFARM_MEMORY = min([ZIMFARM_MEMORY, physical_mem])
    else:
        ZIMFARM_MEMORY = physical_mem
```

This PR refactors it to this:
```py
physical_mem = psutil.virtual_memory().total
ZIMFARM_MEMORY = as_pos_int(
    humanfriendly.parse_size(getenv("ZIMFARM_MEMORY", default=str(physical_mem)))
)
```

In the old code, `ZIMFARM_MEMORY` starts out as None and tries to read an env variable. If an exception happens, the block `if ZIMFARM_MEMORY` would never be entered because it starts out as `None`. So, it would always default to `physical_mem`.

The new code does pretty much the same thing but doesn't catch the exception which I think is a good thing. If a user sets a value that isn't an integer, then, the program shouldn't start.

The diff in the PR shows the flaw when I was to simplifying the logic.

### DIFF
--- a/worker/src/zimfarm_worker/common/constants.py
+++ b/worker/src/zimfarm_worker/common/constants.py
@@ -1,3 +1,4 @@
+import logging
 import multiprocessing
 import os
 import re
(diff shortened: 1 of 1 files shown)


elfkuzco:
This should solve the issue in the slack channel.

elfkuzco:
Couldn't resuse the app `logger` since that uses an env variable defined in constants. Decided to use the `logging` module directly.

PR #1315, opened 2025-09-11
show stats for tasks even if workers appear offline

### LINKED ISSUE #1308: Worker stats seem to take into account only first worker
Stats below seems to be wrong. Instead of taking into account all worker, it seems to take into account only first worker. Or only online workers, I don't know since it is the same currently. It should take into account all running tasks, no matter worker status.

Image

elfkuzco:
## Changes
- include stats of workers that  are not "online" in resource computation

This fixes #1308


benoit74 (approved):
LGTM

Issue #1307, opened 2025-09-10
Task worker has badly failed


benoit74:
Task worker for [eb357fe6-1d1b-4816-a943-20913bb9a17a](https://farm.openzim.org/pipeline/eb357fe6-1d1b-4816-a943-20913bb9a17a) task has badly failed.

Scraper container and dnscache were still running, I had to kill them manually (and I re-requested the recipe).

Logs:

```
[2025-09-09 13:49:19,776: INFO] starting zimfarm task-worker for eb357fe6-1d1b-4816-a943-20913bb9a17a.
[2025-09-09 13:49:19,776: INFO] configuration:
	username=wmf-worker
	webapi_uris=['https://api.farm.openzim.org/v2']
	workdir=/data
	task_id=eb357fe6-1d1b-4816-a943-20913bb9a17a
[2025-09-09 13:49:19,776: INFO] testing workdir at /data…
[2025-09-09 13:49:19,776: INFO] 	workdir is available and writable
[2025-09-09 13:49:19,776: INFO] testing private key at /etc/ssh/keys/zimfarm…
[2025-09-09 13:49:19,781: INFO] 	private key is available and readable (SHA256:p2ongNJ6EpbemgzDnJkdwjA5WheTw9Wbp4NZcyfasrg)
[2025-09-09 13:49:19,781: INFO] testing authentication with https://api.farm.openzim.org/v2…
[2025-09-09 13:49:20,344: INFO] 	authentication successful
[2025-09-09 13:49:20,344: INFO] testing docker API on /var/run/docker.sock…
[2025-09-09 13:49:20,377: INFO] 	docker API access successful
[2025-09-09 13:49:20,389: INFO] Host hardware resources:
	CPU : 6 (total) ;  6 (avail)
	RAM : 15 GiB (total) ;  15 GiB (avail)
	Disk: 200 GiB (configured) ;  210.79 GiB (avail) ; 50 GiB (reserved) ;  150 GiB (remaining)
[2025-09-09 13:49:20,389: INFO] registering exit signals
[2025-09-09 13:49:20,389: INFO] Fetching task details for eb357fe6-1d1b-4816-a943-20913bb9a17a
[2025-09-09 13:49:20,660: INFO] Updating task-status=started
[2025-09-09 13:49:21,027: INFO] Setting-up workdir
[2025-09-09 13:49:21,039: INFO] Starting DNS cache
[...]

PR #1316, opened 2025-09-11
handle exceptions in periodic scheduler

### LINKED ISSUE #1318: Do not block periodic scheduler when a requested tasks does not match constraints
Currently, when the periodic scheduler tries to create a requested task for a schedule which does not match expected constraints, the periodic scheduler fails completely.

I feel like not accepting to create a requested task which does not match our expected constraints is kind of a good way to fix current issues in recipe configuration.

But failing the whole periodic scheduling for one "bad" recipe is problematic.

There is also a flaw in current periodic scheduling behavior in that there must be only one periodic scheduler running because it commits requested tasks only at the end of all computations.

I suggest that we make periodic scheduling more resilient by:
- commit every new requested task to DB as soon as it is ready
- make it impossible to request the same schedule twice at DB level (unique constraint on schedule name / id probably)
- make it more resilient to any error during one schedule processing : if a schedule fails to be requested or if anything else bad happens during this schedule processing, then we log a warning and move on to next schedule

elfkuzco:
## Rationale
While viewing the logs associated with #1307 , I noticed that some exceptions where thrown by the server because `monitor` seems to be missing for some offliners. Also, I noticed the periodic scheduler errors out because some models violate validation while requesting for schedule, thus, causing the supervisor to crash

## Changes
- use `parse_bool` to convert `monitor` to boolean
- handle validation errors for periodic scheduler, thereby allowing it to proceed with other schedules.

EDIT: Incorporated specifications in #1318 to handle all kinds of errors

This fixes #1318


elfkuzco:
Running this locally, I noticed it happens for zimit (probably because of titles failing validation and not using relaxed schema). And a few other schedules. Could offer insights on how to fix recipes manually at db or via UI

backend/src/zimfarm_backend/utils/scheduling.py
```diff
@@ -65,13 +66,16 @@ def request_tasks_using_schedule(session: OrmSession):
...
+            try:
+                request_task(
+                    session=session,
+                    schedule_name=schedule.name,
+                    requested_by=requester,
+                    worker_name=worker,
+                    priority=priority,
+                )
+            except ValidationError:
+                logger.exception(f"could not request {schedule.name}")
```
benoit74: `logger.exception` expects an exception ; I assume you wanted to say `logger.error`

elfkuzco: `logger.exception` doesn't require an exception actually. It expects to be called from an exception handler and then logs with error. The exception info is added automatically to the logging message.

benoit74: Indeed, my bad. I learned something, thank you!

PR #1317, opened 2025-09-11
use paths filter for pull requests

elfkuzco:
## Rationale

Previously, if a pull request is submitted, all the CI workflows are triggered even though the change has nothing to do with certain directories. By specifiying `paths` filters, we potentially skip waiting for CI jobs that have nothing to do with the PR to complete.

## Changes
- use `paths` filter for pull requests in workflows


elfkuzco:
Added some comments to `backend` files and CI related for backend only runs. Dropping commit.

PR #1320, opened 2025-09-12
keep track of schedule changes

### LINKED ISSUE #498: Keep track of recipe's changes
For the moment, we don't really keep track of changes on recipes (or this is not visible for end-users). This has two major consequences:
* The transparency about who does what is not secured
* Considering that this is a tool which is a bit technical, there is no way to learn from the changes that other users made.

It would be great to have a kind of revision log (like on Mediawiki).

Keeping track of each revision in json format and having a diff display between revisions would be really helpful.

I'm in favour as well to keep a big more information about each revision like:
* Comment about the revision
* Author of the revision
* Date

elfkuzco:
## Rationale
As edits are made to a schedule, it becomes important to see the changes made, the author and most importantly the rationale behind the edit. This maintains transparency and learn from the changes that users have made. This PR introduces a `ScheduleHistory` model to keep track of the changes made to a schedule in order to bring these benefits.

## Changes
- create a model/table for schedule histories
- add route to fetch schedule history
- create a history on schedule create/update at the db
- add view to see the differences between a history and it's previous one.

This fixes #498


elfkuzco:
Screenshot_20250912_112659
Screenshot_20250912_112644

backend/src/zimfarm_backend/common/schemas/orms.py
```diff
@@ -186,6 +186,28 @@ class ScheduleDurationSchema(BaseModel):
...
+    comment: str | None
+    name: str
+    category: str
+    enabled: bool
+    language_code: str
+    tags: list[str]
+    periodicity: str
+    context: str
+    # entries are serialized as dict[str, Any] instead of ScheduleConfigSchema
+    # because the entry is possibly outdated and would fail validatiion as the
```
benoit74: `validatiion` => `validation`

backend/src/zimfarm_backend/db/models.py
```diff
@@ -226,6 +226,41 @@ class Schedule(Base):
...
+class ScheduleHistory(Base):
+    __tablename__ = "schedule_history"
+    id: Mapped[UUID] = mapped_column(
+        init=False, primary_key=True, server_default=text("uuid_generate_v4()")
+    )
+    schedule_id: Mapped[UUID] = mapped_column(
+        ForeignKey("schedule.id", ondelete="CASCADE"), init=False
+    )
+    author: Mapped[str]
+    created_at: Mapped[datetime] = mapped_column(index=True)
```
benoit74: Why all these indexed columns? I don't feel that we will heavily perform search filtering on these columns, besides `schedule_id`

benoit74 (changes requested):
I'm very impressed by the visual quality of the diffs, very nice job! And overall it is really moving in the right direction.

I have few technical remarks (inline comments).

From usability perspective, I think we can refine the PR a bit. Note that beside first point below, all can be moved to dedicated issues because they are not immediately mandatory. This would allow to merge this PR, get benefit from history immediately and refine UI/UX a bit later.

Please let me know what you wanna fix in this PR and open issues for every topic mentioned here you wanna fix later (because it is too complex or because it deserves a bit of discussions).

## DB initialization

We should automatically create a DB record in history table during the Alembic migration. This would allow to have diffs immediately at first modification after release.

## Display of history

*can be moved to another issue, this is a refinement*

When comments are a bit long and multiline, it gets very odd display (see second entry in the list below) ; I do hope we will get such long comments, at least we should have a UI ready for it.

image

I also feel like:
- the date is not very visible, making it hard to know in which order history is displayed / when things happened
- author is quite big / takes a lot of space, somehow putting a lot of emphasis on a "who should I blame"
- "No commit message" => "no comment" ; most Zimfarm users won't know what a commit is in a (near) future
- UI is not responsive / barely usable on small screens
- the first column (list of history records) should have a dedicated vertical scrollbar

## URLs

*can be moved to another issue, this is a refinement*

I would like to have a linkable URL (probably with the IDs of both versions being compared displayed in URL as query parameters) so that I can link to it in Github issues, Slack message, ...

## Diff any versions

*can be moved to another issue, this is a refinement*

I would like to be able to diff any versions (not only one version and the one just before it). Pretty sure we will have groups of small changes not always relevant to analyze.

Not exactly sure how user will be able to select these two versions, but mediawiki solution is pretty ok.

## Input of comment

*can be moved to another issue, this is a refinement*

The way change comment is asked for is not adequate. It is barely visible, most users will forget about it and you need to scroll to find it. I don't want to make this a mandatory input (this is just too annoying and anyway users will find ways to input silly value just to please the system)

My proposition (but I'm not convinced at all this is the best idea) is to not add the comment field in the regular fields but display a popup / confirmation screen after user has clicked "update" button. On this screen, user would be prompted for a comment about the changes he just made, and he will be displayed with the list of changes that will be saved (the diff). Two buttons (Cancel / Save) will be able in the top right corner.

elfkuzco:
I think we should move the remainder of the comments to a seperate issue. I have set up the DB initialization within Alembic. Review.

backend/src/zimfarm_backend/migrations/versions/60c546babd23_add_schedule_history.py
```diff
@@ -0,0 +1,82 @@
...
+            ondelete="CASCADE",
+        ),
+        sa.PrimaryKeyConstraint("id", name=op.f("pk_schedule_history")),
+    )
+    with session.begin():
+        for schedule in session.scalars(select(Schedule)):
+            history_entry = ScheduleHistory(
+                author="admin",
+                created_at=getnow(),
+                comment="Initial history created with Alembic",
```
benoit74: "Initial history created automatically" is probably clearer, most users have no idea what Alembic is.

backend/src/zimfarm_backend/routes/schedules/logic.py
```diff
@@ -239,6 +247,12 @@ def update_schedule(
         db_get_schedule(session, schedule_name=schedule_name)
     )
 
+    schedule_config = cast(ScheduleConfigSchema, schedule.config)
```
benoit74, line 250: Why duplicating this line twice?

benoit74 (approved):
LGTM ; please take into account my two comments, and then you can merge ; please also create one issue per "topic" in my comment, with more or less same level of details, so that we do not loose track of them

Thank you very much, this is clearly a long awaited feature and I'm glad we will move this to production very soon. It will allow to start recording history, and this is clearly the missing bit. And then we will have time to refine this with remaining smaller issues.

Issue #1321, opened 2025-09-12
500 error while retrieving task in API


benoit74:
See https://api.farm.openzim.org/v2/tasks/95df5141-69f1-49ef-9780-1aa94cdbe92e or https://api.farm.openzim.org/v2/tasks/74f1a9ac-ea66-4132-a4fa-eb6a628bedce or https://api.farm.openzim.org/v2/tasks/bd50028c-5896-4509-9c8b-44c9e2f2930 (among few others).

Reading a task data should always be possible.

elfkuzco:
last url does not appear to be a valid UUID

benoit74:
Correct URLS (with one more "gift" ^^):

https://api.farm.openzim.org/v2/tasks/bd50028c-5896-4509-9c8b-44c9e2f29307	
https://api.farm.openzim.org/v2/tasks/c28028c1-79ec-4526-a274-9721f8573201	
https://api.farm.openzim.org/v2/tasks/74f1a9ac-ea66-4132-a4fa-eb6a628bedce	
https://api.farm.openzim.org/v2/tasks/95df5141-69f1-49ef-9780-1aa94cdbe92e

elfkuzco:
What do you think about the one where the optimisationcache is not a valid URL. Should that be allowed to pass too?

benoit74:
Just saw that ... I don't really know, because this is something which should not have happened anyway. I will fix DB to not loose the rest of the task info.

benoit74:
https://api.farm.openzim.org/v2/tasks/c28028c1-79ec-4526-a274-9721f8573201 is now fixed

elfkuzco:
Also, the db model says upload should be a `dict[str, Any]` but somehow, it's getting validated against None

benoit74:
> Also, the db model says upload should be a dict[str, Any] but somehow, it's getting validated against None

For which task ID(s) ?

elfkuzco:
> > Also, the db model says upload should be a dict[str, Any] but somehow, it's getting validated against None
> 
> For which task ID(s) ?

https://api.farm.openzim.org/v2/tasks/bd50028c-5896-4509-9c8b-44c9e2f29307

benoit74:
I've fixed `74f1a9ac-ea66-4132-a4fa-eb6a628bedce` for which `upload` was `None`, but for `bd50028c-5896-4509-9c8b-44c9e2f29307` it looks like there is another issue.

benoit74:
[...]

PR #1324, opened 2025-09-12
define type for relaxing validation of gutenberg one-language-one-zim

elfkuzco:
## Rationale
While building the gutenberg flags, the `one_language_one_zim` is converted to a string which causes it to fail validation because our `skip_validation` does not apply to core types. By defining a custom type which wraps around the `bool` with the added benefit of skipping validation, we bypass this error while reading.

## Changes
- define custom type to skip validation of booleans

This fixes one of the cases mentioned in #1321


elfkuzco:
> LGTM ; I feel like `SkipableBool` better convey the meaning than `RelaxedBool` ; but this might be a personal feeling

It's better, I concur

PR #1326, opened 2025-09-12
define global app exception handlers

elfkuzco:
## Rationale

Across the application routes, there are calls to `try...except` and 90% of the time, they re-raise the same error message that the db raises. To simplify this and cut down redundancy, I've defined handlers to catch these exceptions and convert them to the same format clients expect

## Changes
- define global exception handlers to intercept application errors


benoit74 (approved):
LGTM

PR #1336, opened 2025-09-15
reload schedule on clone

### LINKED ISSUE #1329: Cloning recipe should not clone/copy/move all original associated tasks
I just have cloned recipe `wikipedia_zh_all` to a new `wikipedia_zh-hant_all` and I get:

Image

The whole history should be empty, but you see that this brand new recipe has got all the history of `wikipedia_zh_all`

elfkuzco:
## Rationale
After cloning a schedule (or changing it's name), reset the schedule reference to be null. This triggers a reload of the schedule and it's new properties.

## Changes
- add watcher for when a schedule name changes

This fixes #1329

### DIFF
--- a/frontend-ui/src/views/ScheduleDetailView.vue
+++ b/frontend-ui/src/views/ScheduleDetailView.vue
@@ -929,6 +929,17 @@ watch(
     }
   },
 )
+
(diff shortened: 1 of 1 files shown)


benoit74 (approved):
LGTM

PR #1337, opened 2025-09-15
read task status from timestamp

### LINKED ISSUE #1301: Display of time since recipe started is wrong
For https://farm.openzim.org/pipeline/2ffdea0b-aec4-49cc-aec3-409f705327ab

Correct data:

Image

But overview is wrong:

Image

(current local time is 16:15)

elfkuzco:
## Rationale
In #1277, the task `updated_at` was updated for `scraper_running` events in order to be able to cancel incomplete tasks that have not been updated. As a consequence, the `updated_at` does not reflect the actual timestamp which one of the desired timestamp for a status was set. By using the helper function `getTimestampStringForStatus`, we can retrieve the exact time the status was updated (defaulting to the `updated_at`).  This fixes the inconsistencies in the UI as decribed in #1301 

## Changes
- add status plus timestamp from backend responses
- use helper function to retrieve the timestamp for a particular status in frontend instead of the default `updated_at`

This fixes #1301

### DIFF
--- a/backend/src/zimfarm_backend/common/schemas/orms.py
+++ b/backend/src/zimfarm_backend/common/schemas/orms.py
@@ -145,6 +145,7 @@ class MostRecentTaskSchema(BaseModel):
     id: UUID
     status: str
     updated_at: MadeAwareDateTime
+    timestamp: list[tuple[str, datetime.datetime]]
 
 
 class ConfigOfflinerOnlySchema(BaseModel):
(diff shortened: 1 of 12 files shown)


benoit74 (approved):
I feel like modifying `TaskLink` component to get `timestamp`, `status`, and `updated_at` and calling `getTiemstampStringForStatus` from within the component would be cleaner, but I might be picky and/or miss the big picture.

Anyway, LGTM

PR #1338, opened 2025-09-15
default to reserved timestamp for task duration

### LINKED ISSUE #1332: Buggy duration when task has never really ran
See e.g. https://farm.openzim.org/pipeline/7657af2a-0a7e-4850-b336-ca7da0a2c9c2

Image

Image

700 months is quite a long duration 🤣 

We do have other occurence in the DB. Problem is linked to the fact that "chance" worker seems to never achieve to start task so the task stays "reserved" for some time and then the periodic task cleanup fails the task.

elfkuzco:
## Rationale
When a task is created, it always has `reserved` set by the `task_event_handler`. However, sometimes, a task has not been started by the worker and the `started` timestamp will default to `1970, January 1` on the frontend when used to compute the duration. This leads to buggy/incorrect duration values when task has never really run

## Changes
- default to `reserved` when computing the `startedOn`. old code did  `started_on() { return this.task.timestamp.started || this.task.timestamp.reserved; }`
- for task duration, return `Not actually Started` with falsy value as default
- remove unused `getTimestampForStatus`
- pipe duration now computes reliably between `requested` and `started` (or `reserved`) which should always be set.

This fixes #1332

### DIFF
--- a/frontend-ui/src/utils/timestamp.ts
+++ b/frontend-ui/src/utils/timestamp.ts
@@ -3,30 +3,15 @@
  */
 export const DEFAULT_TIMESTAMP = new Date('1970-01-01T00:00:00.000Z')
 
-export function getTimestampForStatus(
-  timestampList: [string, string][],
-  status: string,
-  defaultTimestamp: Date = DEFAULT_TIMESTAMP,
-): Date {
-  // Filter timestamps for the given status
-  const matchingTimestamps = timestampList
-    .filter(([statusStr]) => statusStr === status)
-    .map(([, timestamp]) => new Date(timestamp))
-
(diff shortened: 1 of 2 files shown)


elfkuzco:
> renaming defaultTimestamp to default and pass Not actually started ⌛ instead of '' and then checking it is empty?

- `default` seems to be a reserved keyword and eslint complains.
- I need to use the return value of the function to compute the `taskDuration`. If it is empty, I simply return that. If it isn't empty, I use that value and build the difference between it and the `updated_at` to show the task duration.

benoit74 (approved):
LGTM, thank you for the explanations!

PR #1339, opened 2025-09-15
define global styles for app

### LINKED ISSUE #1333: Colors of links
Links in the new UI do not have any special color assigned, so they inherit from default values. Which also means that color is changing when already visited.

Image

Zimfarm UI being kinda more a control dashboard than a regular website, I feel like we should have control over all colors being used. And I don't think it is helpful (at least it disturbs me) to have links changing color when already visited.

elfkuzco:
## Changes
- move `public/assets/styles.css` to `src/assets/styles.css` so it gets processed by vite
- define styles for links across the application

This fixes #1333


benoit74 (changes requested):
Looks like it broke the navigation bar display (blue on blue probably):

image

I've been also a bit surprised we changed tab colors like below, but in fact I think I like it, looks like it brings homogeneity to the app; well done.

image

image

benoit74 (approved):
LGTM

PR #1340, opened 2025-09-16
display zim metadata and counter from file info

### LINKED ISSUE #1220: Display ZIM metadata and counters
It would be very handy to display the actual metadata of ZIM files in the zimfarm. This would allow quick and easy checks.
The information is already in the database as the worker sends it (for use by the CMS).

elfkuzco:
## Changes
- use table to render file information showing zim metadata and counter statistics

This fixes #1220

### DIFF
--- a/frontend-ui/src/components/FileInfoTable.vue
+++ b/frontend-ui/src/components/FileInfoTable.vue
@@ -1,0 +1,170 @@
+<template>
+  <v-card max-width="600" max-height="400">
+    <v-card-title class="text-subtitle-2 pa-3">File Info</v-card-title>
+    <v-card-text class="pa-3 overflow-auto">
+      <div class="d-flex flex-column">
+        <!-- Top level entries -->
+        <div v-if="fileInfo.id" class="py-1">
+          <v-row dense class="align-start">
+            <v-col cols="4" class="text-body-2">ID</v-col>
+            <v-col cols="8" class="text-body-2 text-break">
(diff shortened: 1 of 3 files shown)


frontend-ui/src/components/FileInfoTable.vue
```diff
@@ -0,0 +1,154 @@
...
+                    <v-col cols="8" md="9" class="text-body-2 text-break">
+                      <!-- Handle illustration_ attributes as base64 images -->
+                      <div
+                        v-if="isIllustrationAttribute(key)"
+                        class="d-flex flex-column align-start"
+                      >
+                        <img
+                          :src="`data:image/png;base64,${value}`"
+                          :alt="key"
+                          style="
```
benoit74: We need the illustration to be exactly original illustration:
- if metadata key is `Illustration_48x48`, then image in info box must be 48x48, and so on (we only have 48x48 illustrations for now, but we do plan to add more in the future)
- no border, no radius
- `object-fit: contain` is probably correct since this is what most readers would do (image should already be 48x48 anyway)

benoit74 (approved):
LGTM ; see inline comment + would it be possible to select text in popup (typically to copy it) without having the popup automatically closing? Currently whenever I try to select text it closes the popup, which prevent me to copy it ; quite sure we will need to copy it sometimes to report issues.

elfkuzco:
> would it be possible to select text in popup (typically to copy it) without having the popup automatically closing?

Done

PR #1345, opened 2025-09-16
notify errors while requesting for tasks

### LINKED ISSUE #1114: UI: Better handle assignment of tasks to a single worker
Currently, when requesting a task we can assign it to a single worker.

User however misses information around how meaningful this assignment is. Typically, it occurs to me (and not only me) every now and then to assign a task which needs more resources than the worker is able to provide. One could also assign a task to a worker which is not processing this scraper. And there are probably other scenario to cover.

A UI enhancement indicating this kind of issues would help a lot to avoid stupidities and frustrations.

### LINKED ISSUE #1342: Issue when requesting a task on worker which cannot do it due to context
When I request a task needing a specific context on a worker which does not has this context, then the UI says that the task has been requested but it was not really requested in fact.

Probably going to be solved with https://github.com/openzim/zimfarm/issues/1114 anyway, but wanted to mention it so that we do not forget to check for it.

elfkuzco:
## Rationale
When we want to request a task for a worker, we should perform some checks to ensure that the task can actually be done by the worker and propagate errors accordingly to the client. These checks closely mirror those used by the `get_tasks_doable_by_worker` (done in SQL) but at Python level for each schedule that is requested.

## Changes
- show errors if a user cannot be able to request for a task
- sort workers by online status so that while requesting on the UI, one can easily see the workers that are online

This fixes #1342
This fixes #1114


backend/src/zimfarm_backend/db/requested_task.py
```diff
@@ -82,22 +87,59 @@ def request_task(
...
+    # Check if worker has enough resource to run a task from this recipe
+    if worker and (
+        worker.cpu < schedule.config["resources"]["cpu"]
+        or worker.memory < schedule.config["resources"]["memory"]
+        or worker.disk < schedule.config["resources"]["disk"]
+    ):
+        return RequestTaskResult(
+            requested_task=None,
+            error=(
+                "Worker does not have enough resources to run schedule "
```
benoit74, line 168: Maybe list resources required by the schedule vs the ones advertised by the worker so that it is easier for the user to understand the problem?

benoit74 (approved):
LGTM, see my comment before merging

Issue #1347, opened 2025-09-16
Schedule history URLs


elfkuzco:
It would be nice to have a linkable URL of schedule histories (probably with the IDs of both versions being compared displayed in URL as query parameters) so that I can link to it in Github issues, Slack message, ...

PR #1350, opened 2025-09-17
make history display responsive

### LINKED ISSUE #1346: Display of history
When comments are a bit long and multiline, it gets very odd display (see second entry in the list below) ; I do hope we will get such long comments, at least we should have a UI ready for it.

Image

Also,
  - the date is not very visible, making it hard to know in which order history is displayed / when things happened
  -  author is quite big / takes a lot of space, somehow putting a lot of emphasis on a "who should I blame"
    "No commit message" => "no comment" ; most Zimfarm users won't know what a commit is in a (near) future
   - UI is not responsive / barely usable on small screens the first column (list of history records) should have a dedicated vertical scrollbar

### LINKED ISSUE #1348: Schedule history diff versions
It would be nice to be able to diff any versions (not only one version and the one just before it). Pretty sure we will have groups of small changes not always relevant to analyze

elfkuzco:
## Changes
- show history items as list that can be selected in any order by the user
- show diff in a seperate window with button to go back to the list entries
- make history items to be responsive and usable on small screens
- make dates more visible

This fixes #1346
This fixes #1348

Screenshot_20250918_003438
Screenshot_20250918_003432
Screenshot_20250918_003402
Screenshot_20250918_003348

### DIFF
--- a/frontend-ui/src/components/DiffViewer.vue
+++ b/frontend-ui/src/components/DiffViewer.vue
@@ -5,23 +5,30 @@
         <v-card-text class="pa-0">
           <div class="diff-content">
             <div v-for="(diff, index) in flattenedDiffs" :key="index" class="diff-line">
-              <div
-                :class="getDiffLineClass(diff.type)"
-                class="d-flex align-center pa-2 font-monospace text-body-2"
-              >
-                <div class="diff-symbol mr-3">{{ getDiffSymbol(diff.type) }}</div>
-                <div class="diff-path mr-4 text-caption">{{ diff.path }}</div>
-                <div class="diff-value flex-grow-1">
-                  <template v-if="diff.type === 'modified' && diff.oldValue">
-                    <span class="text-error">{{ diff.oldValue }}</span>
-                    <span class="mx-2 text-medium-emphasis">→</span>
-                    <span class="text-success">{{ diff.value }}</span>
-                  </template>
-                  <template v-else>
-                    {{ diff.value }}
-                  </template>
-                </div>
-              </div>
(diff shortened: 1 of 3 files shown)


benoit74 (commented):
I like it, thank you.

My only concern is that the diff takes into account in which order versions have been selected which is kinda confusing.

I feel like we should always diff from oldest version to newest version, no matter in which orders versions have been selected.

WDYT? Was it done on purpose or is this just an oversight?

elfkuzco:
> WDYT? Was it done on purpose or is this just an oversight?

Yes, it was done on purpose. Because on #1347 , if a user wants to share a URL, whether they decide to share from previous to current or current to previous wouldn't matter. I just thought of it like how git shows diffs (not caring which came first). 

Arguably, most people would want to show from current to previous and this might just be overkill.

What do you think?

benoit74:
Indeed I doubt anyone would want to show from current to previous. In git this is a bit different because it is hard for git to be sure which commit come before the other commit.

Maybe we can implement a middle ground: in https://github.com/openzim/zimfarm/issues/1347 / https://github.com/openzim/zimfarm/pull/1352, respect the order passed in the URL. But when user is selecting versions, always sort them chronologically (oldest one first) before building "the URL". This way, most users would not be surprised. But should an someone want to do something weird, he will be able to manipulate the URL. And we could imagine in the future to add a "reverse order" button.

If it is easier, I don't mind if you merge this PR as-is and build this implementation on top of https://github.com/openzim/zimfarm/pull/1352

elfkuzco:
Would implement it here instead.

benoit74 (approved):
LGTM

PR #1351, opened 2025-09-17
ask for comments in dialog after editing schedule

### LINKED ISSUE #1349: Schedule history comments
The way change comment is asked for is not adequate. It is barely visible, most users will forget about it and you need to scroll to find it. I don't want to make this a mandatory input (this is just too annoying and anyway users will find ways to input silly value just to please the system)

My proposition (but I'm not convinced at all this is the best idea) is to not add the comment field in the regular fields but display a popup / confirmation screen after user has clicked "update" button. On this screen, user would be prompted for a comment about the changes he just made, and he will be displayed with the list of changes that will be saved (the diff). Two buttons (Cancel / Save) will be able in the top right corner.

elfkuzco:
## Changes
- use `ConfirmDialog` component to ask for comments after editing schedule

This fixes #1349 

Screenshot_20250918_004744

### DIFF
--- a/frontend-ui/src/components/ConfirmDialog.vue
+++ b/frontend-ui/src/components/ConfirmDialog.vue
@@ -9,7 +9,9 @@
       </v-card-title>
 
       <v-card-text class="text-body-1">
-        {{ message }}
+        <slot name="content">
+          <span v-if="message">{{ message }}</span>
+        </slot>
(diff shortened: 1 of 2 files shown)


benoit74 (approved):
LGTM, this is awesome, I love being displayed with the diff of changes before saving!

PR #1352, opened 2025-09-18
load schedule history diff from url

elfkuzco:
## Changes
- add functionality to load schedule diffs from urls. url looks like `?compare=<uuid>...<uuid>`
- avoid fetching history entries that have already been fetched (since they should be immutable) and reorder when we manage to get a new entry from the frontend
- add route to fetch a specific history item


benoit74 (approved):
LGTM, thank you

PR #1353, opened 2025-09-18
remove precision from filesizes formatting

### LINKED ISSUE #1105: Rendering problem around sizes
- Location: UI
- Task: https://farm.openzim.org/pipeline/48abf1a8-f355-43a2-9491-fd500150b0cc

## Problem

Image

Rendering of maxi size is "a bit" odd: `1.01e+3 MiB`. Looks like conversion to MiB is done after the decision to use MiB or GiB units, which causes this edge-case situation

Probably linked to https://github.com/openzim/zimfarm/pull/1086 which is recent change

elfkuzco:
## Changes
- remove `precision` while formatting filesizes. This allows the library to handle precision on its own. Default number of decimal places is 2 which is sufficient.
This fixes the issue where sizes show up as `1.01e+3 MiB` as in https://farm.openzim.org/pipeline/48abf1a8-f355-43a2-9491-fd500150b0cc

### Without precison
```node
> filesize.filesize(1058223158, {base: 2, standard: 'iec'})
'1009.2 MiB'
> filesize.filesize(529825922, {base: 2, standard: 'iec'})
'505.28 MiB'
> filesize.filesize(1024*1024, {base: 2, standard: 'iec'})
'1 MiB'
```

### With precison
```node
> filesize.filesize(1024*1024, {base: 2, standard: 'iec', precision: 3})
'1.00 MiB'
> filesize.filesize(1058223158, {base: 2, standard: 'iec', precision: 3})
'1.01e+3 MiB'
> filesize.filesize(529825922, {base: 2, standard: 'iec', precision: 3})
'505 MiB'
```

The library offers options like `round` and `roundingMethod` which we can use to get rid of decimal places if they are undesired.

### DIFF
--- a/frontend-ui/src/utils/format.ts
+++ b/frontend-ui/src/utils/format.ts
@@ -75,7 +75,7 @@ export function fromNow(value: string) {
 }
 
 export function formattedBytesSize(value: number) {
-  return filesize(value, { base: 2, standard: 'iec', precision: 3 }) // precision 3, display in KiB, MiB,... instead of KB, MB,...
+  return filesize(value, { base: 2, standard: 'iec' }) // display in KiB, MiB,... instead of KB, MB,...
 }
 
 export function getTimezoneDetails() {


benoit74:
Original request when we developed this code in https://github.com/openzim/zimfarm/issues/989 was to have 3 [significant figures](https://en.wikipedia.org/wiki/Significant_figures).

So from my understanding `1058223158` bytes should give either `1010 MiB` (1009.2 MiB with 3 significant figures) or `0.985 GiB`. I feel like `0.985 GiB` is easier to read / less misleading, but that's probably a personal bias. 

Either we have a bug in `filesize` library or we do not use it correctly or I don't get something.

I don't mind to merge this PR to have a least something more readable than `1.01e+3 MiB` (which is awful) but issue is not fixed.

elfkuzco:
Think I might have open an issue on the filesizes repo. Tried multiple params but still not yielding expected results.

elfkuzco:
Unlinking the PR so it doesn't automatically close the issue. But I have opened avoidwork/filesize.js#200

Issue #1358, opened 2025-09-18
Add new permission allowing a user to register as a worker


benoit74:
In https://github.com/openzim/zimfarm/issues/764, I've raised both a problem of impersonation and a problem of authorization.

This issue focus on the authorization issue.

With current system, once a user is authenticated, it can call `/v2/workers/{name}/check-in` worker endpoint without any special condition. If the name of the worker passed does not exists, a new worker will be created.

While impact is minimal (one record in a DB table per call) and condition is to be authenticated (i.e. not everyone can do this), we should probably restrict creation of new worker to users with a dedicated `worker` `create` permission (or something like that).

I've imagined more complex things in https://github.com/openzim/zimfarm/issues/764 but it seems unnecessary.

PR #1359, opened 2025-09-18
Fix worker check-in: impersonation + new worker

### LINKED ISSUE #764: Add more security around /worker/{name}/check-in endpoint
## Problem

The endpoint `/worker/{name}/check-in` requires an authenticated user but nothing more.

## Expected behavior

The endpoint should requires an authenticated user with appropriate rights, and for workers the authenticated user id must match the worker user id (a worker should not be allowed to check-in for another one, could cause serious problems/bugs).

## Proposition

- create a new role `WorkerPermissions` with two names : "checkin-all" (checkin for any worker) and "checkin-self" (checkin only for yourself)
- in the "ROLES" map, grant the "checkin-all" to admins (+ managers ?) + grant the "checkin-self" to workers
- update existing users scope in DB
- check these permissions in the checkin endpoint code

### LINKED ISSUE #1357: Unable to enroll new worker
I've just tried to enroll a new worker.

Call to `/v2/auth/ssh-authorize` works fine but call to `/v2/workers/{name}/check-in` fails with a 404.

benoit74:
## Rationale

Fix #1357 
Fix #764 

## Changes

- do not check existing worker status when there is none because it is the first check-in of this worker
- if worker already exists, systematically check that worker is associated with current user before performing any operation to avoid impersonation

Note that the authorization topic has been moved out of #764 to #1358

### DIFF
--- a/backend/src/zimfarm_backend/routes/workers/logic.py
+++ b/backend/src/zimfarm_backend/routes/workers/logic.py
@@ -12,8 +12,13 @@ from zimfarm_backend.db.models import User
 from zimfarm_backend.db.user import check_user_permission
 from zimfarm_backend.db.worker import check_in_worker as db_check_in_worker
 from zimfarm_backend.db.worker import get_active_workers as db_get_active_workers
-from zimfarm_backend.db.worker import get_worker as db_get_worker
+from zimfarm_backend.db.worker import (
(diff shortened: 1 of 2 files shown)


elfkuzco (approved):
LGTM

Issue #1031, opened 2024-10-12
Average Duration can (under certain conditions) be wrong


kelson42:
Looking at https://farm.openzim.org/recipes/cest-pas-sorcier_fr_astronomie, it's written `Duration: 40 minutes, 40 seconds`
![Image](https://github.com/user-attachments/assets/9d2f8dca-deb7-4859-a9fc-b09db9b1caf1)

But this recipe had only two (failed) attempts/tasks and each of them has taken only a few minutes.

Therefore the ~ 40 minutes duration can not be valid.

Considering that AFAIK this value should be the longest duration for the last tracked successful attempts, I guess that no value should be computed/displayed at all for the "Duration" in such a scenario.

rgaudin:
I believe this duration is coming from the schedule that has been cloned into this one. It's actually indicated by the `default` and the date. Scheduler  kinda requires an existing duration which is why we have a default.

benoit74:
Yep, I'm pretty sure it comes from the clone. I'm quite sure it is solved once we have a succesful run. Anyway, this should still be fixed.

elfkuzco:
@benoit74 , could you take a look at this at it should not happen anymore? In #1360, the scraper duration now computes between when a worker started and when it succeeded, thus making this unlikely.

benoit74:
I confirm values in DB :

| name | default | value | name |
| :--- | :--- | :--- | :--- |
| cest-pas-sorcier\_fr\_astronomie | true | 2678400 | null |
| cest-pas-sorcier\_fr\_astronomie | false | 117 | pixelmemory |
| cest-pas-sorcier\_fr\_astronomie | false | 1053 | badger1 |
| cest-pas-sorcier\_fr\_astronomie | false | 255 | athena18 |

And values in UI :

<img width="1512" height="544" alt="Image" src="https://github.com/user-attachments/assets/f9fd24fa-01fb-4b90-b706-f147e79857d7" />

Are aligned.

It is a bit hard to tell if the "2 minutes" on pixel memory worker is correct or not, we do not have a "2 minutes" succeeded run in the history. 

In order to close the task, @elfkuzco can you please double check in codebase when durations are created / updated in DB (it should be only when it task has succeeded)?

elfkuzco:
[...]

Issue #897, opened 2024-01-15
Average ¨Duration" seems wrong (for at least one recipe)


kelson42:
https://farm.openzim.org/recipes/thalesdoc_en_all
![image](https://github.com/openzim/zimfarm/assets/1029718/026ca406-6bf0-4276-990f-d9fc95b6bc8b)

Should be something around 5 days

elfkuzco:
@benoit74 , could you take a look at this at it should not happen anymore? In https://github.com/openzim/zimfarm/pull/1360, the scraper duration now computes between when a worker started and when it succeeded, thus making this unlikely.

benoit74:
I confirm thales recipe duration values are now correct in DB:

| default | value | name |
| :--- | :--- | :--- |
| true | 2678400 | null |
| false | 471991 | ondemand |
| false | 6887 | rhys |

And they are correctly displayed in UI.

Either we can't repro or bug is fixed.

PR #1360, opened 2025-09-19
show more columns in recipe history

### LINKED ISSUE #432: More columns in recipe history
It would be nice, if the History in a recipe could show more details like which worker and how long it needed:
<https://farm.openzim.org/recipes/wikipedia_tum_all>

## History
scraper_started | 5 minutes ago

worker        | status    | when     | duration
---------------+----------+------------+-----------------------------
mwoffliner2 | succeeded | a month ago | about an hour

instead of:
succeeded | 2 months ago

elfkuzco:
## Changes
- add columns for worker and duration in recipe history runs.
- add link to workers across ui since there is now a workers detail page
- use no-wrap on task detail links since they are mostly short and wrapping makes them look odd

This fixes #432

### DIFF
--- a/frontend-ui/src/utils/offliner.ts
+++ b/frontend-ui/src/utils/offliner.ts
@@ -2,7 +2,10 @@ import constants from '@/constants'
 import type { ScheduleDuration, WorkerScheduleDuration } from '@/types/base'
 import type { OfflinerDefinition } from '@/types/offliner'
 import type { ExpandedScheduleConfig, ScheduleConfig } from '@/types/schedule'
-import type { Task } from '@/types/tasks'
+import type { Task, TaskLight } from '@/types/tasks'
+import { formatDuration, formatDurationBetween } from '@/utils/format'
+import { DateTime } from 'luxon'
+import { getTimestampStringForStatus } from './timestamp'
 
 export function getSecretFields(offlinerDefinition: OfflinerDefinition[] | null) {
   if (offlinerDefinition === null) return []
@@ -66,65 +69,41 @@ export function buildCommandWithout(config: ExpandedScheduleConfig) {
 }
 
 interface SingleScheduleDuration {
-  single: boolean
+  single: true
   value: number
+  formattedDuration: string
   worker: string
   on: string
-  minValue: null
-  maxValue: null
-  minWorkers: null
-  maxWorkers: null
 }
 
 interface MultipleScheduleDuration {
   single: false
-  minValue: number
-  maxValue: number
+  formattedMinDuration: string
+  formattedMaxDuration: string
   minWorkers: WorkerScheduleDuration[]
   maxWorkers: WorkerScheduleDuration[]
+}
+export function single_duration(value: number, worker: string, on: string): SingleScheduleDuration {
+  return {
+    single: true,
+    value: value,
+    formattedDuration: formatDuration(value * 1000),
+    worker: worker,
+    on: on,
+  }
 }
 
 export function buildScheduleDuration(
   duration: ScheduleDuration | null,
 ): SingleScheduleDuration | MultipleScheduleDuration | null {
   if (!duration) return null
-  function single_duration(value: number, worker: string, on: string): SingleScheduleDuration {
-    return {
-      single: true,
-      value: value,
-      worker: worker,
-      on: on,
-      minValue: null,
-      maxValue: null,
-      minWorkers: null,
-      maxWorkers: null,
-    }
-  }
-
-  function multipleDurations(
-    minValue: number,
-    maxValue: number,
-    workers: Record<string, WorkerScheduleDuration>,
-  ): MultipleScheduleDuration {
-    const minWorkers = Object.values(workers).filter(function (item) {
-      return item.value == minValue
-    })
-    const maxWorkers = Object.values(workers).filter(function (item) {
-      return item.value == maxValue
-    })
-    return {
-      single: false,
-      minValue: minValue * 1000,
-      maxValue: maxValue * 1000,
-      minWorkers: minWorkers,
-      maxWorkers: maxWorkers,
-    }
-  }
 
   if (!duration.available && duration.default) {
     return single_duration(duration.default.value, 'default', duration.default.on)
   }
+
   if (!duration.workers) return null
+
   const workersArray = Object.values(duration.workers)
   const minWorker = workersArray.reduce(
     (min, worker) => (worker.value < min.value ? worker : min),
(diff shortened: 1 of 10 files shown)


elfkuzco:
Screenshot_20250919_020134
Screenshot_20250919_015630

elfkuzco:
Normally, I expected that the values in the duration columns would show up in the `between worker1 and worker2` statements but it doesn't always prove to be the case. For example, I see `1 minute 20 seconds` in the second image but it doesn't appear in any of the duration column values. 

This discrepancy is because the statement `between worker1...` uses values sent from the backend and it is calculated as https://github.com/openzim/zimfarm/blob/45d59709dfb6aec73b1faec1675bfa5259499e9c/backend/src/zimfarm_backend/db/schedule.py#L167-L181

and it is only invoked when scraper sends `scraper_completed` event. But the history run shows different statuses and as a consequence, the frontend computes the duration from when the scraper started and when it was last updated_at.

Not sure if the backend logic should be left as it  is.

benoit74:
I agree about your concern regarding the "Duration field". I feel like this is exactly the problem raised in #897 and #1031. I'm very open to suggestions on how to enhance this information. 

Original goal was to get an estimate of how fast the run was on fastest worker and how slow it was on slowest worker. Ultimate idea was to be able to pinpoint some workers which had slowness issue.

So far, it looks to me current implementation is a failure. Computation is so odd (not saying it is wrong, I don't know) that it is hard to trust. And variability of task duration can be induced by many external factors (number of modifications in the wiki invalidating cache size, current load on upstream server slowing execution down, temporary noisy neighbor on the worker ...) and we try to interpret it as coming from worker condition.

I'm starting to wonder if we should not simply get rid of this "workers_duration" data (everywhere, API, DB), and let the user do its own analysis based on historical values which are now displayed. Trying to automatically gain knowledge from a duration which could be impacted by tons of external factors is not going to lead anywhere. What I'm sure is that no real action or analysis are arising from this data, and no process are mandating it.

@kelson42 @rgaudin do you feel like we should keep this "Average duration" (between xxx on workerx and yyy on workery) and fix it ; or do you agree it is both too hard to trust and can be removed for the time being?

rgaudin:
First of all let's not call it average duration as it's the opposite of an average duration 😆

I'm leaning towards removal because as it was not trusted, it was not used and I don't think it was missed that much.
As you said, people look at the history table and figure something out.

On the other hand, it feels stupid that every user for every such screen needs to compute something like this while this is clearly a machine's job.

So either this information is useless (not my opinion but I'm no ZF user) and we drop it or it's useful and we fix it. I'd suggest the simple, consistency road of computing it client-side with the table's values.

benoit74:
I like a lot the idea of computing this client-side based on existing values. Thank you.

elfkuzco:
> I like a lot the idea of computing this client-side based on existing values. Thank you.

On it...

elfkuzco:
turns out the function `to_duration_obj` was not correct. it was erroneous in assuming days would always be 30 days in a month. removed it because I noticed a weird scenario where it was returning `3 days ..., ` while it was actually `2 days...`. Ended up setting the `formattedDuration` directly in the durations dict and let `luxon` handle the computation.

benoit74:
Looking a bit more into the codebase, I begin to get worried that we use something for scheduling (`ScheduleDuration`) which is different than what we use for display (last tasks duration).

Let me take some time to make a decision on this.

benoit74:
OK, I've now analyzed this in more depth and I feel like I have an "exit strategy" 🤣 

There is indeed a confusion between two different things (at least I got confused):
- the "scraper duration": this is the time it takes to run only the scraper part ; this is very important in current scheduling implementation because it is used to estimate tasks ETA ; computation takes into account that another task can start as soon as scraper completes, even if upload is still ongoing, and takes into account worker might need time to download scraper but this is rare ; both are correct (matches worker behavior at least)
- the "total duration": this is the time displayed in most places which takes into account upload duration

I hence propose to have two fields at "recipe-level":
- the "Scraper duration": "rollback" most recent changes and present data as it was but with updated field name
- the "Total duration": the new values you've just computed in most recent changes

The "table header" should be renamed "Total duration" as well for consistency.

WDYT?

Sorry for changing my mind in the middle of development, but this was definitely not clear enough in my mind.

And of course we will need to solve other issues around "Scraper duration" being wrong.

elfkuzco:
Okay. If I understand correctly, you are saying to use the data from the API as it is. Because, the one I computed recently uses the difference between when it was started/reserved and its current status. Is that correct?

The value from the API is always going to be `started` - `scraper_completed` as shown here 
 https://github.com/openzim/zimfarm/blob/45d59709dfb6aec73b1faec1675bfa5259499e9c/backend/src/zimfarm_backend/db/schedule.py#L167-L181

benoit74:
Yes, I'm saying we should use and display both values. They are different, they have different usage, and they are both correct. Value coming from the API is the one used for scheduling tasks so it is important. Value you've computed UI-side is a summary of duration shown below so it is useful to save user time.

elfkuzco:
Screenshot_20250923_161916

Distinguished between "Scraper Duration" and "Total Duration". 
"Scraper Duration" is set by backend and is computed as `scraper_completed` - `started` while "Total Duration" computes `current_status/updated_at` - `started/reserved`.
In the above image, we can see it takes `6 hours 30 minutes` to complete scraping but "Total Duration" takes additional 20 minutes.

benoit74:
Please also add a "Total duration" field below "Scraper duration"; this "Total duration" must be the summary of values in table.

elfkuzco:
> Please also add a "Total duration" field below "Scraper duration"; this "Total duration" must be the summary of values in table.

Summary as in sum of all the total durations or average of all the total durations? Or any other mathematical function?

benoit74:
Summary as what you have implement in https://github.com/openzim/zimfarm/commit/dabec69c9a46719f21a274a5aff7987fa45c3582 (from ... on worker xxx to ... on worker yyy) ; but this should take into account only "succeeded" tasks, I just realize it seems it tooks into account all tasks.

elfkuzco:
Screenshot_20250925_123416


Total Duration now computes between `started` and `succeeded`

frontend-ui/src/views/ScheduleDetailView.vue
```diff
@@ -951,6 +1053,15 @@ const shortId = (id: string | null): string => {
   return id ? id.substring(0, 8) : ''
 }
 
+const calculateTaskDuration = (task: TaskLight): string => {
+  if (!task.timestamp) return ''
+  const started = getTimestampStringForStatus(task.timestamp, 'started', '')
+  if (!started) return 'Not actually started ⌛'
+  const succeeded = getTimestampStringForStatus(task.timestamp, 'succeeded', '')
```
benoit74: We need to use either "succeeded" or "failed" to compute individual tasks duration ; knowing how long a failed task ran before failing is useful at individual task level.

frontend-ui/src/utils/offliner.ts
```diff
@@ -68,63 +71,48 @@ export function buildCommandWithout(config: ExpandedScheduleConfig) {
 interface SingleScheduleDuration {
   single: boolean
```
benoit74: Must be `single: true` ; this will allow to get rid of `formattedMinDuration`, `formattedMaxDuration`, `minWorker` and `maxWorker` on this interface, and `formattedDuration` on `MultipleScheduleDuration`. Please clean this, it makes code extra hard to read.

benoit74 (approved):
Two "minor" points and we are good.

PR #1361, opened 2025-09-19
enable recipe multi-selection

### LINKED ISSUE #718: Enable recipe selection
The current system only allows for the relaunch of all displayed recipes whereas it would make a lot more sense to be able to relaunch a selected few (basically only those that have failed)

Capture d’écran 2022-09-23 à 09 13 30

A selector/tick button at the end of each line would come in handy.

elfkuzco:
## Changes
- enable recipes to be selected with finer granularity using select boxes
- changed `Clear` button to `Clear Filters` since that is what it actually does.
- make buttons responsive and switch to column layout on small screens

This fixes #718 

Screenshot_20250919_043522

### DIFF
--- a/frontend-ui/src/components/RequestSelectionButton.vue
+++ b/frontend-ui/src/components/RequestSelectionButton.vue
@@ -9,7 +9,11 @@
     @click="emit('fetch-schedules')"
   >
     <v-icon size="small" class="mr-1">mdi-plus</v-icon>
-    {{ isRequesting ? requestingText : `Request those ${nbSchedules} recipes` }}
+    {{
+      isRequesting
+        ? requestingText
+        : `Request ${nbSchedules} selected recipe${nbSchedules !== 1 ? 's' : ''}`
+    }}
   </v-btn>
 </template>
 
--- a/frontend-ui/src/components/SchedulesTable.vue
+++ b/frontend-ui/src/components/SchedulesTable.vue
@@ -1,19 +1,34 @@
 <template>
   <div>
     <v-card v-if="!errors.length" :class="{ loading: loading }" flat>
-      <v-card-title class="d-flex align-center justify-end flex-wrap py-2">
-        <div class="d-flex align-center">
-          <RequestSelectionButton
-            :can-request-tasks="canRequestTasks"
-            :requesting-text="requestingText"
-            :count="paginator.count"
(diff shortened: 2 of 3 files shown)


elfkuzco:
I am going to need a bit of input to know if there are things to refine.
Currently, the selections you make persist even when you click the "Next page" button (assumed it would be desirable). Wanted to use `watch` to autoselect selections (especially whenever schedules change either from initial mount or "Next Button") but it has some edge cases that might be complex to work around. Such cases are:
- there's a setinterval that updates the page too. if I hook into the `watch` for schedules, I would need to need some flag to seperate when it's coming from `setinterval` or the intended areas
- even if I used the flag, the watch would mean that if I go to a previous page I had visited or changed the limit, everything would be re-selected (invalidating any de-selections I have made). I don't think this behaviour is desirable.

As a consequence, I decided to let the user manually make their selections. There's a quick button to select everything and then individually de-select. But then again, if I make a search after selecting, my searches are still included. I figured this simple approach would be sufficient given it's probably beneficial if you want to mix and match from various searches.

What do you think?

benoit74:
I feel like what you've implement is very sufficient and efficient. 

The fact that you can select all recipes on the page when you click the top check box is fine, I like the fact it does not select all recipes on all pages.

My only concern is that it is hard to deselect all recipes which have been selected so far, since unchecking the top check box will first select all recipes on the page (which is fine) and then unselect all recipes on the page (and not all recipes). You have to search in which page the remaining selected items are, which is kinda annoying.

I see two way forward to solve this last concern. Either when unchecking the top check box it unselects all recipes (clearing the whole selection), not only the ones on current page. But this is kinda weird because it selects only from current page but it unselects everything. Or we add a new button next to "Clear filters" saying "Clear recipes selection" (or something like that). 

I wonder if we still need a button allowing to select all recipes on all pages at once. I'm quite sure it is not worth it, one can set page size to 100 and then I don't see where it would need more than a handful of pages to select all required recipes. So I feel current UI is sufficient for this "select all recipes" scenario.

elfkuzco:
I can add a button to "Clear all recipes" since it's directly tied to the selected languages. should be an easy fix

elfkuzco:
Screenshot_20250919_093934

benoit74 (approved):
LGTM

PR #1364, opened 2025-09-19
upgrade filesize library to handle precision properly

### LINKED ISSUE #1105: Rendering problem around sizes

elfkuzco:
## Changes
- upgrade filesize library and reset precision to 3. upstream has closed avoidwork/filesize.js#200

This fixes #1105 
Screenshot_20250919_134755

### DIFF
--- a/frontend-ui/package.json
+++ b/frontend-ui/package.json
@@ -15,7 +15,7 @@
   "dependencies": {
     "axios": "^1.10.0",
     "deep-diff": "^1.0.2",
-    "filesize": "^10.1.6",
+    "filesize": "^11.0.7",
     "fuse.js": "^7.1.0",
     "jwt-decode": "^4.0.0",
     "luxon": "^3.6.1",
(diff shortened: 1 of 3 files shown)


benoit74 (approved):
LGTM ; `0.990` would have been more correct than `0.99`, but I feel like this is coming from my scientific background and I'm pretty sure mostly no-ones really care 🤣

elfkuzco:
> LGTM ; `0.990` would have been more correct than `0.99`, but I feel like this is coming from my scientific background and I'm pretty sure mostly no-ones really care 🤣

🤣🤣🤣
I felt the same way when I saw it but author had already made one wrong commit, reverted before fixing. Think I will work on it during the weekend.

PR #1372, opened 2025-09-23
Drop configuration by environment variable in frontend-ui

benoit74:
## Rationale

Configuring Frontend UI by environment variables to then transform them into a JSON is a pain. It forces to use variables prefixed by `ZIMFARM_` to differentiate them from other environment variable, configuration keys are not case sensitive, configuration is built at runtime which is never easy to debug, and it is hard to understand.

## Changes

- drop `entrypoint.sh` + associated `ENVIRON_PATH`
- drop `ENTRYPOINT` in Dockerfile


benoit74:
Nota: both frontend ui are already configured with plain configuration JSON file (see https://github.com/kiwix/operations/commit/82fceed9f56415ec13a55423002d0d887464c064), I "hacked" the entrypoint by writing to a random file. Will need to be cleaned-up, but at least we can confidently deploy this branch

PR #1376, opened 2025-09-25
compute single duration for total duration in milliseconds

### LINKED ISSUE #1375: Total duration is wrong on `ncase.me_en_all`
For some reasons I don't get at all, the "total duration" of https://farm.openzim.org/recipes/ncase.me_en_all is completely wrong.

Image

I checked multiple other recipes and they are still fine, so this must be a significant edge case.

elfkuzco:
## Rationale
the `single_duration` function assumes `duration` would be in `seconds` (the values come from the backend). So, it converts to milliseconds before forwarding to luxon. However, using it in `buildTotalDurationDict` proves buggy becuase those durations are already in milliseconds. This causes issuse like #1375 where 7,800,000 gets multiplied by 1000, causing it to give 90 days as return value. Without this extra multiplication, the value would be 2 hours, 10 minutes (as expected)

## Changes
- move `single_duration` helper function inside the `buildScheduleDuration` as that's where it's helpful
- compute the formatted duration directly from the duration in `buildTotalDurationDict`

This fixes #1375

### DIFF
--- a/frontend-ui/src/utils/offliner.ts
+++ b/frontend-ui/src/utils/offliner.ts
@@ -83,20 +83,21 @@ interface MultipleScheduleDuration {
   minWorkers: WorkerScheduleDuration[]
   maxWorkers: WorkerScheduleDuration[]
 }
-export function single_duration(value: number, worker: string, on: string): SingleScheduleDuration {
-  return {
-    single: true,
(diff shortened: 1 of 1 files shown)


benoit74 (approved):
LGTM

PR #1380, opened 2025-09-29
display characters limit on schedule editor

### LINKED ISSUE #1025: UI: enforce limit in the UI and display number of characters
For fields which have a limit on number of characters, it would be super helpful to:
- enforce the limit while editing the field (currently one has to save the recipe to have an error message)
- have a visual indicator of current number of characters (this is not my idea but one from @Popolechien in #777 and I can only agree)

elfkuzco:
## Changes
- show counter for text fields if they have a max length
- show flag description in hint instead so it aligns nicely with the counter
This fixes #1025 

Screenshot_20250929_115000

### DIFF
--- a/frontend-ui/src/components/ScheduleEditor.vue
+++ b/frontend-ui/src/components/ScheduleEditor.vue
@@ -404,36 +404,41 @@
             />
             <v-textarea
               v-else-if="field.component === 'textarea'"
-              v-model="editFlags[field.dataKey]"
+              :model-value="editFlags[field.dataKey]"
+              @update:model-value="(value) => handleInputWithGraphemeLimit(field, value)"
               density="compact"
               variant="outlined"
               :placeholder="field.placeholder"
               :required="field.required"
-              :counter="field.max_length ? true : undefined"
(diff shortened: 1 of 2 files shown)


kelson42:
@elfkuzco Can you please confirm that all the code related to string length measures graphemes and not usual characters? See https://wiki.openzim.org/wiki/Metadata#Graphemes

elfkuzco:
seems Vuetify's counter is using barebones length. Thanks for reminding me. I will add a workaround for this.

benoit74 (changes requested):
Just like in the backend, we need to count graphemes and not unicode characters.

For instance `विकिपीडिया से कंप्यूटरकंप्यू` should count for 14, not 29.

benoit74:
Oups, looks like I should have updated PR status before pushing my review 🤣

elfkuzco:
Screenshot_20250930_044825

used javascript to restrict the max length and also fill the counter as vuetify doesn't enforce the max length and if it is set on the HTML element directly, it doesn't count the number of graphemes but the number of characters

PR #1383, opened 2025-09-30
update top-level schedule config attributes from explicitly set values in request

### LINKED ISSUE #1382: Clearing `artifacts` setting is not working
It is not possible to clear the `artifacts` setting of a given schedule once set.

Image

elfkuzco:
## Changes
- update schedule config top-level attributes with values explicitly set in request (whether None or not) and let Pydantic handle validation internally.

This fixes #1382

### DIFF
--- a/backend/src/zimfarm_backend/routes/schedules/logic.py
+++ b/backend/src/zimfarm_backend/routes/schedules/logic.py
@@ -454,17 +454,20 @@ def update_schedule(
         ):
             raise BadRequestError("Image name must match selected offliner")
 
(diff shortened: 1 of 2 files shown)


elfkuzco:
I tried from the API (because you can't set to empty from UI) to set it to an empty string and it indeed failed. I believe pydantic has a way of detecting if a field was set to None or defaulted to None.

elfkuzco:
used `exclude_unset` to distinguish between set values and unset to update top-level config attributes. see tests

PR #1395, opened 2025-10-02
remove worker ip from api response

### LINKED ISSUE #1392: Hide IP everywhere (to anonymous users)
IPs are private information. Therefore they never should be publically available.

It seems that we have at least one place where this is violated

Image

elfkuzco:
## Changes
- remove worker ip from API response

This fixes #1392


benoit74:
awaiting answer to https://github.com/openzim/zimfarm/issues/1392#issuecomment-3360593955

benoit74 (approved):
For the time being, let's merge this as-is and then introduce new permission to show worker IP to admins.

PR #1396, opened 2025-10-02
remove pkcs8_key from ssh keys

### LINKED ISSUE #1275: Finalize support of SSH keys in PEM format
In `zimfarm-upgrade` branch, we've started to implement support of SSH keys in PEM format.

Work is however not complete, e.g. frontend assumes an SSH format.

elfkuzco:
## Changes
- remove pkcs8_key column from SSH key. Can always be recomputed using serialize_public_key function.
- verify signed messages with public key instead of pkcs8_key value

This closes #1275


backend/src/zimfarm_backend/migrations/versions/59de2c4dbaf0_remove_ssh_key_pkcs8_key_column.py
```diff
@@ -0,0 +1,31 @@
...
+depends_on = None
+
+
+def upgrade() -> None:
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.drop_column("sshkey", "pkcs8_key")
+    # ### end Alembic commands ###
+
+
+def downgrade() -> None:
```
benoit74, line 25: Is it possible add proper recomputation of column value on downgrade?

elfkuzco: Yes. But I would need to add it in the DB for all SSH keys with the function. Only caveat is that some ssh-keys did not have the `ssh-key ....` in them. I think we were splitting it in the older API. If all entries are valid SSH keys that I can pass to cryptography, then, it's doable. Otherwise, it would fail and the downgrade wouldn't succeed

benoit74: Let's live without this then

PR #1399, opened 2025-10-02
create and update offliner definitions

### LINKED ISSUE #1389: Add ability to create new offliner definition version
This is a follow-up to https://github.com/openzim/zimfarm/pull/1266

We need to add ability for scrapers CI to create new offliner definition.

I propose:
- add a new `POST /offliners/{offliner_id}/versions` endpoint
- add a `ci_secret` column in `offliner` DB table, initialized to a random secret value
- the endpoint should be called with offliner definition + appropriate secret value
- update the `update_scraper_version` maintenance script: modify offliner definition version together with docker image tag, and add a (commented-out) example where a flag is renamed from `oldName` to `newName` (showing how an admin can modify this script to migrate offliner flags when upgrading / downgrading versions)

The last point is a bit dirty, but it looks like the reasonable first step to do for now.

elfkuzco:
## Changes
- add ci_secret_hash column to offliner table
- add endpoint to create an offfliner definition version
- update `update_scraper_version` script to update an existing offliner defintion. By providing `--name-mappings`, users can add, delete or modify existing data in the DB to conform to the updated definition. 

This closes #1389


backend/maint-scripts/update_scraper_version.py
```diff
@@ -38,12 +109,102 @@ def update_scraper_tag(session: OrmSession, scraper: str, tag: str):
...
-    try:
-        scraper, tag = sys.argv[1:]
-    except (IndexError, ValueError):
-        logger.error(f"Usage: {sys.argv[0]} SCRAPER_NAME VERSION_TAG")
-        sys.exit(1)
-
-    with Session.begin() as session:
-        update_scraper_tag(session=session, scraper=scraper, tag=tag)
+    parser = argparse.ArgumentParser(
+        description="Update schedules and (unstarted) requested tasks image tag "
```
benoit74: Remove `(unstarted)`, requested tasks are always unstarted by definition, otherwise it is a task

backend/maint-scripts/update_scraper_version.py
```diff
@@ -38,12 +109,102 @@ def update_scraper_tag(session: OrmSession, scraper: str, tag: str):
...
+        "or offliner definition",
+    )
+
+    # Required offliner specification
+    parser.add_argument(
+        "-o", "--offliner", required=True, help="Specify which offliner to update"
+    )
+
+    # Mutually exclusive group for version specification (must be provided)
+    version_group = parser.add_mutually_exclusive_group(required=True)
```
benoit74: Why making them mutually exclusive ? Does not look like it is much more complex to do both in one call + this is generally what we will have to do (move to next image tag and offliner definition when scraper is released)

backend/maint-scripts/update_scraper_version.py
```diff
@@ -38,12 +109,102 @@ def update_scraper_tag(session: OrmSession, scraper: str, tag: str):
...
+                        name_mappings=args.name_mappings,
+                    )
+                    flag_modified(schedule, "config")
+                schedule.offliner_definition_id = offliner_definition.id
+                # Just needed to ensure our model still works
+                create_schedule_full_schema(schedule, offliner)
+                nb_schedules_modified += 1
+
+            logger.info(
+                f"updated {nb_schedules_modified} schedule(s) and "
```
benoit74: Remove ` and `

backend/src/zimfarm_backend/db/offliner_definition.py
```diff
@@ -161,3 +163,75 @@ def get_offliner_versions(
...
+    # - Update the field with name to the new name
+
+    for old_name, new_name in name_mappings.items():
+        if old_name == "+":
+            # Add the new name to the data with a default of None
+            new_data[new_name] = None
+            continue
+
+        if old_name not in data:
+            # Sometimes, the old name is does not exist in the DB probably because
```
benoit74: `is does` => `does`

backend/src/zimfarm_backend/routes/offliners/logic.py
```diff
@@ -87,6 +97,25 @@ async def get_offliner_versions(
...
+@router.post("/{offliner_id}/versions")
+async def create_offliner_version(
+    offliner_id: Annotated[str, Path()],
+    session: Annotated[OrmSession, Depends(gen_dbsession)],
+    request: Annotated[OfflinerDefinitionCreateSchema, Body()],
+) -> Response:
+    """Create a new version for a specific offliner"""
+    offliner = db_get_offliner(session, offliner_id)
+    if not check_password_hash(
+        offliner.ci_secret_hash or secrets.token_hex(32), request.ci_secret
```
benoit74: Why `or secrets.token_hex(32)`? If it is because `ci_secret_hash` might be `None`, then I feel like having a specific message saying that `ci_secret` is not set for this offliner would help debug issues (still raising an `UnauthorizedError`)

benoit74 (approved):
LGTM

PR #1402, opened 2025-10-03
enable archiving of recipes

### LINKED ISSUE #796: Never totally delete recipes
Because we might want to look at or restore an old recipe, we should not really delete a recipe, but instead put it in an `recipe_archive` table.

### LINKED ISSUE #1400: Improve recipe "Delete" button
For the moment it looks like this:

Image

The button "Delete" should be in red.

The task deletion button is called "Remove", any rational why one is called "Delete" and the other one "remove"?

See also https://github.com/openzim/zimfarm/issues/796

elfkuzco:
## Changes
- add column to mark schedules as archived
- prevent archived schedules from being requested as tasks
- add functions/route logic to toogle schedule archive status
- restrict endpoints to toggle schedule archive status
- refactor schedulesview into archived view and recipes view with a base view for shared logic
- hide edit, delete and clone tabs for archived recipes

This closes #796 
This closes #1400 

Screenshot_20251003_140426
Screenshot_20251003_140403
Screenshot_20251003_140356


backend/src/zimfarm_backend/routes/schedules/logic.py
```diff
@@ -628,6 +659,46 @@ def clone_schedule(
...
+@router.patch("/{schedule_name}/archive")
+def archive_schedule(
+    schedule_name: Annotated[ScheduleNameField, Path()],
+    session: OrmSession = Depends(gen_dbsession),
+    current_user: User = Depends(get_current_user),
+) -> JSONResponse:
+    """Archive a schedule"""
+    if not (
+        current_user
+        and check_user_permission(current_user, namespace="schedules", name="delete")
```
benoit74, line 680: In the future, we will probably want a specific "archive" right, but let's live with it for the time being.

backend/src/zimfarm_backend/routes/schedules/logic.py
```diff
@@ -628,6 +659,46 @@ def clone_schedule(
...
+@router.patch("/{schedule_name}/unarchive")
+def unarchive_schedule(
+    schedule_name: Annotated[ScheduleNameField, Path()],
+    session: OrmSession = Depends(gen_dbsession),
+    current_user: User = Depends(get_current_user),
+) -> JSONResponse:
+    """Unarchive a schedule"""
+    if not (
+        current_user
+        and check_user_permission(current_user, namespace="schedules", name="delete")
```
benoit74, line 707: Same remark, in the future, we will probably want a specific "unarchive" right, but let's live with it for the time being.

elfkuzco:
> we should probably prevent recipe deletion at API level is recipe is not already archived

This would require clients like wp1 that create a recipe to also archive before deleting. Upstream changes would be needed.

elfkuzco:
> I don't get why you bypass the history mechanism ; would be interested to know your PoV on this ; I feel like having a track record of who archived a given recipe and when (and optional why with the comment) would be valuable information

I had originally prevented editing when schedule is archived and I didn't want to add the option in the `ScheduleEditor`. I would add the request for comments when the user archives/unarchives in their respective tabs instead.

benoit74:
> This would require clients like wp1 that create a recipe to also archive before deleting. Upstream changes would be needed.

Indeed, nice catch. Let's forget about this comment 🤣

kelson42:
"shedule" or "recipe"? We should be coherent about the naming.

"restore" seems besser as "unarchive".

benoit74 (changes requested):
Please replace any occurrence of "unarchive" in the codebase / API with "restore" (and their variation). We already have too many discrepancies between what is shown in the UI and the technical term being used, we should not let the gap widen.

elfkuzco:
Okay. I think it's more consistent. Will refactor.

PR #1404, opened 2025-10-06
update offliner spec if create fails with unique constraint

elfkuzco:
## Changes
- Adapt create_offliner_definition endpoint to update the spec if insertion fails due to unique constraint violation. This allows overwriting the `dev` versions of scrapers by CI workflow


benoit74 (approved):
LGTM

PR #1406, opened 2025-10-07
update create worker contrib script

elfkuzco:
## Changes
- adapt contrib script to create a new worker to use new API calls and contracts


benoit74 (approved):
LGTM, not tested but can only be better than former code 🤣

elfkuzco:
Tested

PR #1407, opened 2025-10-07
add ip addresses to contexts

### LINKED ISSUE #1167: Store and use list of workers IPs whitelists
Currently we do not have a source of truth regarding which (worker) IPs have been whitelisted, where (in terms of "providers") and which recipes needs this to perform correctly. 

Some is in a Google Sheet (youtube). Some (vikidia and probably others I forgot about) is in my mailbox. The rest is lost. 

This is a problem in terms of:
- knowledge management
- automation (currently vikidia can run only on few hosts ... but the Zimfarm does not know it ; same for youtube)
- automated failure detection (should a worker change its IP, we will not be automatically informed that some "providers" need to update their whitelist)

elfkuzco:
## Changes
- change worker contexts column from list of strings to mapping of context name to whitelisted IP
- ensure that if IP is set on a particular context, worker `last_seen` matches specified IP
- redesign view to update worker contexts

This closes #1167


elfkuzco:
Screenshot_20251007_133741
Screenshot_20251007_132618

elfkuzco:
> same remark, the fact that we need to get to "edit" to see current contexts IPs is kinda "frustrating" ; same problem of permissions I imagine

yes, intentionally omitted it. waiting for the permissions review.

backend/src/zimfarm_backend/db/requested_task.py
```diff
@@ -423,6 +462,21 @@ def get_tasks_doable_by_worker(
                 RequestedTask.worker_id.is_(None),
             )
         )
+    tasks = session.scalars(query).all()
```
benoit74: This looks very verbose and memory consuming way of building the proper list

A simple function 

```python
    def filter_req_task_for_ip_issues(task: RequestedTask):
       # Filter out tasks where worker IP doesn't match the context requirement
        return (
            task.context is None # task has no context
            or worker.contexts[task.context] is None # context is not requiring a specific worker IP
            or worker.contexts[task.context] == worker.last_ip # worker IP is still matching the IP whitelisted for this context
        )
```

and then

```python
return sorted(
        [
            RequestedTaskWithDuration(
...
                ),
                updated_at=task.updated_at,
            )
            for task in filter(session.scalars(query).all(),filter_req_task_for_ip_issues)
...
```

PR #1408, opened 2025-10-08
disable workers from requesting new tasks

### LINKED ISSUE #876: Detect failing workers
We should not let a failing worker (considering that all its tasks are failing early) to continue to get new tasks

### LINKED ISSUE #1022: Add ability to pause worker scheduling
Currently when one want the manager to stop starting new task, it is needed to completely stop the container.

This has many drawbacks:
- worker seems dead in the UI while it is just paused from starting new tasks
- should the machine reboot or docker restart, the manager will restart automatically and begin again to accept new task
- all other duties of the manager are not performed anymore (e.g. atm cancelling tasks that need to be)

I think it would help to create a new status for the worker : `cordon` (to reuse k8s term) or `scheduling_disabled`

In this status the worker would just stop starting new tasks but will continue all the rest of its duties.

This should probably be configured in `/etc/zimfarm.config` (needing a manager restart, but this is not a problem)

This status should be reported during check-in, stored in DB, displayed in UI.

elfkuzco:
## Changes
- add column `cordoned` to allow worker refusing to scheudule (read from ZIMFARM_SCHEDULING_DISABLED) env
- add column `admin_disabled` to allow admins disable worker from scheduling
- prevent request of tasks or working requesting for new tasks if either column is true
- update zimfarm.sh script and worker manager to checkin with new config

This new columns effectively allow a worker to not be considered for scheduling with control either from the admin side or the worker side.

This closes #1022 
This closes #876


elfkuzco:
Screenshot_20251008_132508

frontend-ui/src/components/WorkersTable.vue
```diff
@@ -73,14 +73,37 @@
...
+
+              <!-- Admin scheduling status -->
+              <v-chip
+                v-if="item.admin_disabled"
+                color="error"
+                size="x-small"
+                variant="tonal"
+                prepend-icon="mdi-shield-alert"
+              >
+                Disabled
```
benoit74: `Disabled` => `Disabled by Zimfarm Administrator`

frontend-ui/src/components/WorkersTable.vue
```diff
@@ -73,14 +73,37 @@
...
+              </v-chip>
+              <!-- Local scheduling status -->
+              <v-chip
+                v-if="item.cordoned"
+                color="info"
+                size="x-small"
+                variant="tonal"
+                prepend-icon="mdi-pause-circle"
+              >
+                Cordoned
```
benoit74: `Cordoned` => `Cordoned by worker owner`

frontend-ui/src/views/WorkerDetailView.vue
```diff
@@ -2,16 +2,40 @@
...
+
+          <!-- Local scheduling status -->
+          <v-chip
+            v-if="metrics?.cordoned"
+            color="info"
+            size="small"
+            variant="tonal"
+            prepend-icon="mdi-pause-circle"
+          >
+            Cordoned
```
benoit74: Same changes

frontend-ui/src/views/WorkerDetailView.vue
```diff
@@ -2,16 +2,40 @@
...
+
+          <!-- Admin scheduling status -->
+          <v-chip
+            v-if="metrics?.admin_disabled"
+            color="error"
+            size="small"
+            variant="tonal"
+            prepend-icon="mdi-shield-alert"
+          >
+            Disabled
```
benoit74: Same changes

worker/contrib/zimfarm.config.example
```diff
@@ -61,6 +61,9 @@ USE_PUBLIC_DNS=""
 # might help with some online website or DNS servers
 DISABLE_IPV6=""
 
+# Set to `"y"` to disable scheduling for this worker.
```
benoit74: I would prefer we use the "cordon" term here and in worker variables as well for clarity ; we should definitely keep explananation that "cordon" means "disable starting new tasks"

benoit74: `ZIMFARM_` prefix seems useless as well. Maybe simply `CORDONED` is sufficient

elfkuzco:
> Do you recall why GET /workers/{worker_name} was not used in the first place? 

The route didn't exist in the first place probably because there wasn't any consumer for it in the first place.

Issue #740, opened 2023-02-01
Prevent creation of duplicate recipes


kelson42:
By increasing the number of recipes and the number of people able to create and modify recipes, we increase as well the number of (potential) duplicate recipes. See for example https://github.com/openzim/zim-requests/issues/560.

we should elaborate a solution to help users detect these duplicate early and warn them about the « danger ».

benoit74:
We need to refine this issue, adding the `discussion` tag.

From my understanding, the problem we have is two fold.

First, we do not want to create duplicate work, creating a new recipe for something which is already done. Second, we do not want two recipes producing the same ZIM filenames (and names obviously).

## Duplicate work

The problem of duplicate work is an "early" problem in terms of zim-request process. Whenever a request arrives, we need to find if this request already exists in the Zimfarm. This is hence a discoverability issue. We are not yet at the stage of creating a recipe, but only checking request validity.

Problem is probably that current schedule naming convention and current UI makes it too hard to find recipes "similar" to something we would consider to build.

I don't think schedule naming convention can be refined further, there is already many constraints on this naming and many decision made.

So the problem seems to be that it is hard to ask the Zimfarm the good question. I feel like in 99% of the cases, the question is "give me all recipes which are processing this web domain". This is hence a transverse question, how a given scraper configure the fact that it is processing a given web domain being different from scraper to scraper.  

This "webdomain processed by a given recipe" is probably something we can store in DB at schedule creation / modification time, in a dedicated `web_domain` column. The way to populate this column would be scraper-based, but this is something we know easily at schedule creation / modification time. And this would allow fast and easy searching / grouping of similar schedules, without having to dive into the offliner flags at every query.

With this `web_domain` column, we can at least add this to recipes search criteria on https://farm.openzim.org/recipes, which seems to help cover most problems.

We can also imagine informing the user at recipe creation and modification time: add a button next to the field allowing to modify the web domain to quickly open the list of similar recipes (https://farm.openzim.org/recipes screen with pre-populated search criteria).

## Conflicting ZIM filename

The second problem is that nothing prevents an editor to create two recipes creating the same ZIM filename (and/or ZIM name). And this currently leads to very silent overriding of each other files over and over again.

It is important to note that the recipes could be very different: different scrapers, totally different configuration besides the ZIM filename (or whatever leads to it). Comparing recipe configuration is hence of little help, it will not cover all cases at all.

[...]

PR #1410, opened 2025-10-10
store recipe similarity data on create/update

### LINKED ISSUE #1354: Implement a system to better view / detect similar recipe configuration
See "Duplicate work" in https://github.com/openzim/zimfarm/issues/740#issuecomment-3244566945

We should implement something to better detect recipes which are too similar.

Probably focusing on the web domain first is a good idea, but it has the drawback of completely ignoring stuff like Youtube recipes where the same playlist could end in two different ZIMs.

I also like the idea of being able to compare different scrapers for similarity, so a hash does not seems really adequate.

It seems important to both warn the user when a recipe is updated that there are other similar recipes found. 

But also to let users easily navigate between similar recipes created on-purpose:
- all WPEN recipes
- the low and high res versions of a given Youtube channel (it does not exists yet, but it could)

All while avoiding too many similarities to be found, for system to stay relevant.

Anyway it is clear we will not "prevent" any recipe modification / creation based on this criteria.

@elfkuzco do you have any suggestions to share on this topic?

elfkuzco:
## Rationale
Typically recipes might be different but they often end up doing the same work (typically because they are processing the same domain or their flags are the same). Given it's the scraper-specific code that often knows the flags that are bound to produce the same output, a `similarity_data` column is introduced on the schedule. This field is generated from the offliner definition schema (specified by the scraper). As a consequence, we can detect duplicate recipes even though their names are different.

### Changes
- add a `similarity_data` column to maintain the list of strings that make a recipe unique
- split the `name` value used while searching for recipes and check if there is an overlap with the `similarity_data` column instead of doing name searches only.
- update signatures to create/update schedule

This closes #1354


elfkuzco:
@benoit74 , in #740, you mentioned
> We can also imagine informing the user at recipe creation and modification time: add a button next to the field allowing to modify the web domain to quickly open the list of similar recipes (https://farm.openzim.org/recipes screen with pre-populated search criteria).

But on zimfarm, there's no view to create  a recipe and on wp1, the request is made via API. Could you provide more context on how this should be implemented?

benoit74:
> split the `name` value used while searching for recipes and check if there is an overlap with the similarity_data column instead of doing name searches only.

I don't get what you are trying to achieve with that. The split is mostly never going to have any impact on the current recipes names, so there will probably never be a match.

> But on zimfarm, there's no view to create a recipe and on wp1, the request is made via API. Could you provide more context on how this should be implemented?

WP1 and zimit-frontend are definitely out of scope, especially since we might have two users requesting the exact same ZIM and this should not be considered a problem at all. While it is somehow an issue on farm.openzim.org for ZIMs we are building for "ourselves".

The question we need to answer is "how do we present similarity data to creators in a useful way?". A very preliminary idea was to show something in the recipe "Edit" form when fields are updated. E.g. for mwoffliner, since the similarity data is based on `mwUrl`, every time `mwUrl` is edited we check if current recipe similarity_data (after edit) is similar to any other recipes in the database and we somehow display the list of similar recipes. I'm however really not sure how this would looks like in the UI.

Now that we have a popup asking user to confirm changes and input a comment, maybe this is the right place to also show similar recipes. Since user is about to update recipe configuration, changes are "complete" and this is probably the right time to check. I'm still not sure about how it would look like in the popup. Popup is going to be pretty big with the list of changes, list of similar recipes, input text for comment ... Maybe just a single phrase saying "Number of similar recipes: xxx" is sufficient. Because I also feel like we need somewhere in the UI where one can check which recipes are similar to the current one, without having to make an edit. Maybe a new tab after "Info", "Config": "Similar recipes".

I'm looking forward for your ideas and recommendations on this.

elfkuzco:
> I don't get what you are trying to achieve with that. The split is mostly never going to have any impact on the current recipes names, so there will probably never be a match.

It just makes the search a bit more robust in that it doesn't search for names only. For example, if i know a URL and search for it bu there's no recipe with that URL as name, I could still get a return because a certain recipe is configured to search for the URL

elfkuzco:
> Now that we have a popup asking user to confirm changes and input a comment, maybe this is the right place to also show similar recipes. Since user is about to update recipe configuration, changes are "complete" and this is probably the right time to check. I'm still not sure about how it would look like in the popup. Popup is going to be pretty big with the list of changes, list of similar recipes, input text for comment ... Maybe just a single phrase saying "Number of similar recipes: xxx" is sufficient. Because I also feel like we need somewhere in the UI where one can check which recipes are similar to the current one, without having to make an edit. Maybe a new tab after "Info", "Config": "Similar recipes".

This seems sufficient and simple enough for me. Would need to set  up additional endpoints for it. Proceeding with the implementation

elfkuzco:
- used [GIN Indexes](https://www.postgresql.org/docs/current/gin.html) for storing similarity data as the entry could be exceedingly large. e.g [ted_topic_all](https://farm.openzim.org/recipes/ted_topic_all) which has more than 3000 topics
- reuse scheduletable to display list of similar recipes
- show number of recipes at edit confirmation dialog
- add tab to view similar recipes

elfkuzco:
Screenshot_20251013_135752
Screenshot_20251013_135722

benoit74:
@elfkuzco is this ready for a new review?

elfkuzco:
> @elfkuzco is this ready for a new review?


Yes

benoit74:
I feel like we miss a migration script to add the "similarity_data" to all existing offliner definitions already in the database, and recompute all recipes similarity data. I need this to test locally and we will need this in production. This is only a temporary migration "tool" obviously. Or do I miss something?

elfkuzco:
I updated the `update_scraper_version.py` script to do this. Ran it locally by simply updating the offliners to the same version they already were on. Could write a shell script for that but after #1411 , I am trying to avoid doing upgrades in alembic migrations

benoit74:
It does not work on my end, there is not similarity data configured on offliner definitions which are already in DB.

I agree we should avoid doing data upgrades in alembic migrations, it is a pain.

elfkuzco:
Run
```sh
ZIMIT_USE_RELAXED_SCHEMA=true JWT_SECRET=DH8kSxcflUVfNRdkEiJJCn2dOOKI3qfw POSTGRES_URI=postgresql+psycopg://zimfarm:zimpass@localhost:2345/zimfarm WORKER_OFFLINE_DELAY_DURATION=30d ./update_scraper_version.py -d initial -o youtube
```

The `update_scraper_version.py` script isn't mounted in the container, thus, I need to set some env variables on the commandliner. You can pass the different offliners after the `-o` flag in subsequent commands. This assumes the version is `initial`

benoit74:
> The update_scraper_version.py script isn't mounted in the container

It is available in `zf_backend` in `/app/maint-scripts` (build in Dockerfile, so static after build). But running `/app/maint-scripts/update_scraper_version.py -d initial -o mwoffliner` does run

But anyway, you got me wrong, my problem is not running this script (which I already did). My problem is that the offliner definitions I have in DB do not have any "similarityData" keys. And it will be the same on production database.

elfkuzco:
Aaah, I get what you mean. I have updated the initial_offliner_defintions json files in this PR. They serve as a good base to create the PRs against the offliner repos. That was what I used as the file for the update. For prod, I will have to create PRs against the respective offliners. The route that updates the schemas via CI will update the offliner after which the python script can be invoked. didn't want to update the schedules in the request that updates the offliner definition. Should it be done there?

benoit74:
The problem is that in production we already have an `initial` version, and some scrapers might already have created other versions, and I don't want to update these versions with the CI (too much risk of breaking something). 

Updating `initial_offliner_definition` does not help because this migration has already been ran in production.

I feel like we need a "script / tool / ..." to update all versions already existing in DB adding the same "similarityData" to all these versions, since I know we can compute this similarity the same way and hence with same config on all versions for the time being.

In addition to that, you're right we will have to create PRs on the various offliners, but this will update only `dev` and next releases.

I'm fine with the fact that we also have to call `update_scraper_version` script to really compute the similarity data, but we need "something" to properly configure the way it has to be done.

elfkuzco:
updated `update_scraper_version.py` to take a flag and update the offliner defintions and then update recipes similarity data. See example in script.

backend/maint-scripts/update_scraper_version.py
```diff
@@ -174,24 +226,56 @@ def update_entries(
...
+            ):
+                # update the spec for the offliner definition to use the new
+                # similarity data transformers list
+                offliner_spec = OfflinerSpecSchema.model_validate(
+                    offliner_definition.schema
+                )
+                offliner_spec.similarity_data = similarity_data
+                # update the db offliner definition and all of its associated recipes
+                offliner_definition.schema = offliner_spec.model_dump(mode="json")
+                nb_schedules_modified = update_entries(
```
benoit74: Looks like this is wrong, it is modifying the offliner definition used by schedules while we want to keep the ones currently used. 

We just want to update similarity data of every schedule (this is working well). And only once per schedule, not once per schedule and per offliner definition present in the DB.

frontend-ui/src/components/ScheduleEditor.vue
```diff
@@ -508,6 +508,28 @@
...
+          Checking for similar recipes...
+        </v-alert>
+      </div>
+
+      <div v-else-if="similarRecipesCount !== null && similarRecipesCount > 0" class="mb-4">
+        <v-alert type="warning" variant="tonal" density="compact">
+          <template #prepend>
+            <v-icon>mdi-information</v-icon>
+          </template>
+          There are {{ similarRecipesCount }} similar recipe{{
```
benoit74: `There are {{ similarRecipesCount }} similar recipe{{` => `There are {{ similarRecipesCount }} recipe(s) similar {{` 

I feel like this formulation is more correct (with plural by default), WDYT?

backend/maint-scripts/update_scraper_version.py
elfkuzco: Okay. But the caveat is this. If I update the schedules only and it's offliner definition has a different `similarity_data` or it is not set (empty list), anytime the schedule is updated, perhaps from an editor in the UI, the similarity data is recomputed. This would mean, that any ones set by the script would be undone.

So, should the similarity data not be recomputed at the `update` endpoint?

elfkuzco: https://github.com/openzim/zimfarm/blob/0447df9aa40d6bf37122bc558b73fb43219bece1/backend/src/zimfarm_backend/db/schedule.py#L573-L584

elfkuzco:
separated script to update simalrity data

backend/src/zimfarm_backend/common/schemas/offliners/models.py
```diff
@@ -78,10 +78,29 @@ class ModelValidatorSchema(CamelModel):
     model_config = ConfigDict(extra="forbid")
 
 
+class TransformerSchema(CamelModel):
+    # the name of the transformer function to use for the field. If None, the field
+    # will be used as it is
+    name: Literal["split", "hostname"] | None = None
+    # the operator to use for the transformer function (if the function
+    # takes an operator)
+    operator: str | None = None
```
benoit74: We should name this `operand`, `operator` is the function itself (`split`, `hostname`, ...). Sorry for realizing this only now ... still better than once spread across scrapers ^^

Issue #1397, opened 2025-10-02
Show worker IP to "admins"


benoit74:
In https://github.com/openzim/zimfarm/pull/1395, we have completely removed worker IP for API/UI to quickly fix https://github.com/openzim/zimfarm/issues/1392

We should now implement a solution to show worker IP to Zimfarm admins since this a really important and convenient information to have in the UI.

elfkuzco:
What permission should a user have to able to view a worker's detail? I am thinking to use `check_user_permission(current_user, namespace="users", name="read")`. What do you think?

benoit74:
This is significantly unrelated to users, even if worker 

We need a new permission like `namespace="workers", name="write"` which would mean have write permission on worker config, hence ability to read sensitive data.

Hence the need to tackle it together with https://github.com/openzim/zimfarm/issues/1393

elfkuzco:
we need to display somewhere in the UI when there is a discrepancy between last_ip and contexts IPs: probably some icon in https://farm.openzim.org/workers to indicate this worker has an issue ; and then changing contexts colors in the "contexts" of workers "info" tab and in the "edit" tab as well ; but for this to work we need the last_ip info, which is in https://github.com/openzim/zimfarm/issues/1397

we also need to display the IPs next to the contexts in the worker detail view

PR #1419, opened 2025-10-14
add new permissions

### LINKED ISSUE #1393: Revisit permissions view
The permissions view is not going to scale:

Image

The table makes it looks like we could have `ssh_keys` permission on `zim` and `upload` on `schedules` and so on.

I feel like we should find another more "open" representation. Probably simply a kind of "nested list".

Especially since we will implement even more permissions in the future.

elfkuzco:
## Changes
- add new permissions to further distinguish resources access/update
- add permissions to reveal secrets
- redesign view for permissions list

Screenshot_20251014_113617

This closes #1393

### DIFF
--- a/backend/src/zimfarm_backend/common/roles.py
+++ b/backend/src/zimfarm_backend/common/roles.py
@@ -16,6 +16,8 @@ class Permissions:
 
 class TaskPermissions(Permissions):
     names: ClassVar[list[str]] = [
+        "read",
+        "secrets",
         "request",
         "unrequest",
         "create",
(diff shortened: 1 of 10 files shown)


elfkuzco:
This lays the groundwork for #1397 with the "secrets" permission. However, there is a bit of an issue on how to proceed with merging the PR because the db stores the permissions as a dictionary, not as the enum. This means, that existing scope objects in the DB would need to be updated to have the new permissions, otherwise they won't be able to perform certain actions.

benoit74:
It is indeed not really possible to test this correctly. And I do not really have any idea about how to properly move this into production.

Should we split this PR in two:
1. add a "role" column in DB `user` table so that it is clearer which role we want to assign to each user ; we only have few users with custom scope ATM. And they probably don't need new permissions, at least an admin can check this by hand. And a user can either have a role assigned, or a scope (meaning a custom scope)
2. add new permissions just like you did in this PR, users will automatically inherit new permissions through their roles

PR #1421, opened 2025-10-14
define model for task files

### LINKED ISSUE #1416: "Created After" value is missing in task details
See https://farm.openzim.org/pipeline/e88a81b8-1616-4884-b873-6f34c72b6e82

Image

elfkuzco:
## Rationale
The task `files` schema was a `dict[str, Any]`. This meant that deep within it, the `created_timestamp` and `uploaded_timestamp` which are serialized to datetime objects by the custom `zimfarm_loads` function are  not detected by the API datetime serializer https://github.com/openzim/zimfarm/blob/0b2638bad42a6a2038267f1d6af3a75b6564fd03/backend/src/zimfarm_backend/common/schemas/__init__.py#L8-L23

This further meant that when we want to compute the duration between when a file was created, it compares a `naive` datetime (without the `Z` suffix) to the `events` datetime which has the `Z` suffix. This leads to no data shown for `Created After` on the UI as the `end` time is greater than the `start` time. This is the cause of #1416 and is evident in other task details too

By building a model of the task files, Pydantic can detect that the field is a datetime object and serialize it accordingly. This means that on the UI, when the `formatBetween` function is called, both datetime strings are datetime-aware

## Changes
- define pydantic schema for Task files with defaults set to None as data in the DB may be null or not and we don't want validation errors

This fixes #1416

### DIFF
--- a/backend/src/zimfarm_backend/common/schemas/orms.py
+++ b/backend/src/zimfarm_backend/common/schemas/orms.py
@@ -82,6 +82,34 @@ class TaskLightSchema(BaseTaskSchema):
     config: ConfigWithOnlyResourcesSchema
 
 
+class CMSStatusSchema(BaseModel):
+    status_code: int | None = None
+    succeeded: bool | None = None
+    on: datetime.datetime | None = None
+    book_id: UUID | None = None
+    title_ident: str | None = None
+
+
+class TaskFileSchema(BaseModel):
+    """
+    Schema for reading the files associated with a task
+    """
+
+    name: str
+    size: int | None = None
+    status: str
+    cms: CMSStatusSchema | None = None
+    created_timestamp: datetime.datetime | None = None
+    uploaded_timestamp: datetime.datetime | None = None
+    failed_timestamp: datetime.datetime | None = None
+    check_timestamp: datetime.datetime | None = None
+
+    check_result: int | None = None
+    check_log: str | None = None
+    check_details: dict[str, Any] | None = None
+    info: dict[str, Any] = Field(default_factory=dict)
+
+
 class TaskFullSchema(BaseTaskSchema):
     """
     Schema for reading a task model with all fields
@@ -94,7 +122,7 @@ class TaskFullSchema(BaseTaskSchema):
     container: dict[str, Any]
     priority: int
     notification: ScheduleNotificationSchema | None
-    files: dict[str, Any]
+    files: dict[str, TaskFileSchema]
     upload: dict[str, Any]
     offliner_definition_id: UUID = Field(exclude=True)
     offliner: str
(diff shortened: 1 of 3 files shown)


benoit74:
Did you validated this is going to work well on all tasks we have in DB? We should try to read all of them and check we do not get validation errors and still get all files as expected. I'm a bit worried about the impact of this change should some data be wrong in DB.

Do you want a fresh data export with the bug from #1416 to be sure to reproduce the issue?

elfkuzco:
Yeah, I did. Tried to default to None for most fields I wasn't sure of and mirrored it closely to the same model on zimit-frontend. You could send me a fresh dump just to double-check.

benoit74:
Just sent you a new dump through Slack

elfkuzco:
Ran the migrations and it works for the files schema of existing tasks in the db

elfkuzco:
However, running the model validation against the entries in the dump exposes some issues in the DB dump. The columns `files`, `container` and `debug` should not be nullable. But it appears they contain `null` as values in the column. For example, task `eefedc7a-e312-455d-8867-661ec0fe05c5`. 

A SQL script should probably be used to coalesce these values to `{}` as they violate data integrity.

benoit74:
Can you prepare such an SQL script to coalesce these values to {} as they violate data integrity? And check when is last task with this problem (to confirm these are just old tasks and we are not still creating tasks with problems).

elfkuzco:
New code currently sets them to `{}` while creating the task. It's the old tasks that violate it
https://github.com/openzim/zimfarm/blob/19915be6b7dd1e8a82cf06d9f8dc785959ec19e0/backend/src/zimfarm_backend/db/tasks.py#L220-L242

elfkuzco:
```sql
UPDATE task SET debug = '{}'::jsonb WHERE debug IS NULL OR debug = 'null'::jsonb;
UPDATE task SET container = '{}'::jsonb WHERE container IS NULL OR  container = 'null'::jsonb;
UPDATE task SET notification = '{}'::jsonb WHERE notification IS NULL OR  notification = 'null'::jsonb;
UPDATE task SET files = '{}'::jsonb WHERE files IS NULL OR  files = 'null'::jsonb;
```

elfkuzco:
From dump, oldest task that has this fault is updated at 2023

backend/src/zimfarm_backend/common/schemas/orms.py
```diff
@@ -82,6 +82,34 @@ class TaskLightSchema(BaseTaskSchema):
     config: ConfigWithOnlyResourcesSchema
 
 
+class CMSFileSchema(BaseModel):
```
benoit74: This is not a file but rather the CMS status, so I suggest we rather use `CMSStatusSchema` class name.

PR #1422, opened 2025-10-15
remove route to get initial offliner definiton

### LINKED ISSUE #1391: Cleanup transition of Zimfarm "dependencies" to offliner definition
This is a follow-up to https://github.com/openzim/zimfarm/pull/1266

Once https://github.com/openzim/zimit-frontend/issues/167 and https://github.com/openzim/wp1/issues/1024 have been merged and deployed, we can clean up the codebase:

- drop "initial" defaults in some endpoints used for the transition
- drop `GET /offliner/{offliner_id}` endpoint

elfkuzco:
## Changes
- remove `GET /offliners/{offliner_id}` endpoint
- add route to create offliner via API

This closes #1391

### DIFF
--- a/backend/src/zimfarm_backend/api/routes/offliners/logic.py
+++ b/backend/src/zimfarm_backend/api/routes/offliners/logic.py
@@ -58,32 +58,6 @@ async def get_offliners(
     )
 
 
-@router.get("/{offliner_id}")
-async def get_initial_offliner_version(
-    offliner_id: Annotated[str, Path()],
-    session: Annotated[OrmSession, Depends(gen_dbsession)],
-) -> JSONResponse:
-    """Get a specific offliner"""
-
-    # find the schema class that matches the offliner
-    offliner = db_get_offliner(session, offliner_id)
-    offliner_definition = db_get_offliner_definition(
-        session, offliner_id=offliner_id, version="initial"
-    )
-    schema_cls = build_offliner_model(offliner, offliner_definition.schema_)
-
-    flags = schema_to_flags(schema_cls)
-
-    return JSONResponse(
-        content={
-            "flags": [flag.model_dump(mode="json", by_alias=True) for flag in flags],
-            "help": (  # dynamic + sourced from backend because it might be custom
-                f"https://github.com/openzim/{offliner}/wiki/Frequently-Asked-Questions"
-            ),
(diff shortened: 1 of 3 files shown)


elfkuzco:
@benoit74 , I am going to need your input on the permissions required for the offliner creation via API. I avoided using `schedules, create=True` because there are other roles with that permission that could not be prvileged to create an offliner. Left a todo comment on it. i don't know if this should be paused till #1419  is reviewed and possibly new permissions set.

WDYT?

elfkuzco:
WP1 does not really have a UI to "select the version" they want to run and I figured they would also want to do similar thing. Because the API returns the versions as a list of strings in descending order of created_at. So, if they are going to select the version they would use by "choosing the first" because it should be the most recent, I figured it would be simpler to do it here and avoid one more API call. Same thing applies to Zimit frontend

Do you have an idea on how the selection should be done?

benoit74:
You probably saw that after your comment, just for the reference: https://github.com/openzim/wp1/issues/1024#issuecomment-3409984147 (we will configure the version in WP1 config file)

benoit74 (approved):
LGTM, please do not merge before https://github.com/openzim/zimit-frontend/pull/176 and https://github.com/openzim/wp1/issues/1024 are merged / solved

Moving to draft to avoid accident ^^

benoit74:
@elfkuzco shall we update and merge this now that zimit and wp1 have switched to "other" endpoints / behaviors?

elfkuzco:
Had to cherry-pick the changes and apply  on top of main as code directory layout has changed significantly amongst other things like migrations, etc

elfkuzco:
Should I proceed to merge or do you want to have another look?

benoit74:
In the production logs, everything looks good on api.farm.openzim.org but I see unexpected things on api.farm.zimit.kiwix.org. While most calls are to `GET /v2/offliners/zimit/initial` as expected, I do see some occasional calls to `GET /v2/offliners/zimit`. Where could this come from? Could it be someone with an old UI somewhere or something else?

elfkuzco:
Most likely looks like a call from an older UI. Ran a trace of the `query_api` calls and the offliners own passes the definition. Nothing comes from the UI directly AFAIK

PR #1431, opened 2025-10-17
Do not validate anything in offliner configuration when asked for

benoit74:
When we `skip_validation` in `create_offliner_instance`, we need to mostly completely by-pass pydantic validation logic because we can end-up with a very old / odd offliner config which is not aligned with current offliner definition. We should not mind about this since the goal is to display / return config "as it was".


benoit74:
This is probably going to be obsolete by another approach from @elfkuzco ... let's see ^^

benoit74:
Superseed by #1433

Issue #1411, opened 2025-10-10
Zimfarm backend auto migrate in dev docker graph not working


audiodude:
[//]: # (Pick appropriate classifiers)
- Location: __API__
- Schedule Name/ID: __n/a__
- Task ID: __n/a__

## Problem

Can't start dev docker graph after `git pull origin main`. There seems to be a problem with a database migration:

```
[2025-10-10 16:14:25,180: INFO] Upgrading database schema with config in /usr/local/lib/python3.13/site-packages/zimfarm_backend
INFO:backend:Upgrading database schema with config in /usr/local/lib/python3.13/site-packages/zimfarm_backend
INFO  [alembic.runtime.migration] Context impl PostgresqlImpl.
INFO  [alembic.runtime.migration] Will assume transactional DDL.
INFO  [alembic.runtime.migration] Running upgrade 60c546babd23 -> 12096716adfd, add offliner definitions
INFO  [alembic.runtime.migration] Running upgrade 12096716adfd -> 0607dff50f80, add offliner model
Traceback (most recent call last):
  File "/usr/local/lib/python3.13/site-packages/sqlalchemy/engine/base.py", line 1963, in _exec_single_context
    self.dialect.do_execute(
    ~~~~~~~~~~~~~~~~~~~~~~~^
        cursor, str_statement, effective_parameters, context
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    )
    ^
  File "/usr/local/lib/python3.13/site-packages/sqlalchemy/engine/default.py", line 943, in do_execute
    cursor.execute(statement, parameters)
    ~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.13/site-packages/psycopg/cursor.py", line 97, in execute
    raise ex.with_traceback(None)
psycopg.errors.UndefinedColumn: column "ci_secret_hash" of relation "offliner" does not exist
LINE 1: ...(id, base_model, docker_image_name, command_name, ci_secret_...
                                                             ^

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/usr/local/bin/alembic", line 7, in <module>
    sys.exit(main())
             ~~~~^^
  File "/usr/local/lib/python3.13/site-packages/alembic/config.py", line 636, in main
    CommandLine(prog=prog).main(argv=argv)
[...]

PR #1432, opened 2025-10-17
alembic migrations cleanup

elfkuzco:
## Changes
- remove model updates in alembic migrations to prevent issues like #1411 due to the model fields not being fully set as the migration runs incrementally.
- add route to create offliner via API 
- create shell script to pull offliner definitions from offliner repos and populate fresh database with offliner and definitions
- remove `backend-tools` from compose (don't think it really serves anymore)
- update container for running tests
- update `dev/README` to get any new dev started with new DB and updated instructions


benoit74 (approved):
LGTM

PR #1433, opened 2025-10-17
allow extra fields while reading db

### LINKED ISSUE #1430: Allow extra keys in config object
Calling `model_validate` on config object removes extra keys. This has the consequence that when reading from the config, old values are removed because that's the default behaviour of pydantic. 

This is often desirable as such fields could simply be updated within the config to a new  name and keep the schema still valid.

elfkuzco:
## Rationale
Previously, offliner models had a default of `ignore` (default Pydantic extra settings). This leads to very old tasks that do not meet the definition to fail because they do not contain the newer fields which are required. By hooking into the `create_offliner_instance` model and setting a default of `allow`, we allow extra fields by default since most actions are read operations. However, `extra` is set to `ignore` when creating/updating schedules since this is the source of truth for tasks/requested tasks.

## Changes
- override ConfigDict in `create_offliner_instance` with default of `allow`
- `ignore` extra fields while creating/updating schedules
- use `model_construct` to skip validation entirely. This means, we need to further customize how secrets are shown because they can be raw strings or `SecretStr`

This closes #1430

### DIFF
--- a/backend/src/zimfarm_backend/common/constants.py
+++ b/backend/src/zimfarm_backend/common/constants.py
@@ -100,8 +100,8 @@ SLACK_USERNAME = getenv("SLACK_USERNAME")
 SLACK_EMOJI = getenv("SLACK_EMOJI")
 SLACK_ICON = getenv("SLACK_ICON")
 
-# string to replace hidden secrets with
-SECRET_REPLACEMENT = "--------"  # nosec
+# length of secret strings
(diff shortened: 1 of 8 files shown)


benoit74 (changes requested):
We should add the two additional tests of https://github.com/openzim/zimfarm/pull/1431 + 1 for getting obsolete task (task with flags which are unknown in the offliner definition)

elfkuzco:
added tests for obsolete tasks and schedules

Issue #557, opened 2020-11-20
Create a healtcheck


kelson42:
Like on the Cardshop:
* API
* Web site
* Mandory workers
* S3/Youtube/... access

kelson42:
We just have a scenario in youzim.it where the only one crawler (worker) has been dead for 4 months and we had no way to detect this. We probably need a way to say that a worker is mandatory, or a certain level of worker power has to be available, otherwise the healthcheck will returns an error.

rgaudin:
4 days, not 4 months

benoit74:
Removed from zimit2 project unfortunately

benoit74:
To be clarified: who is monitoring what and how.

kelson42:
@benoit74 Do you have a few specific questions because to me the situation is clear and would glad to answer them. Considering this is a PRIO1 (still believe in this priorisation), it would be also worth to invest the time on that.

benoit74:
I thought we've agreed last week we will discuss this kind of matter together for efficiency reasons.

Anyway, if you already have all answers because it is clear for you, here are my specific questions:
- what is responsible to monitor the various components? We can't have the API monitoring the API or it means all monitoring will stop as soon as API will stop
- what is responsible to display the healthcheck results? If it is only the API, then it means we will have a blackout on this status as soon as the API is down
- what does it means to monitor the various component? how do we intend to implement it? 
  - we should have clear definition of what we currently want to perform as a check to declare a given component is UP or DOWN 
  - for API and Website, I can imagine we just want to check that a query to root path returns a valid HTTP response
  - for workers, does "UP" means we've recently got a query for tasks to start, just like the "offline" status in the UI? (there is no HTTP endpoint exposed by workers)
   - or only some "mandatory" workers?
   - or like you said we want some "worker power" available? but what is it?
  - for S3, I have no idea
  - for Youtube, I don't know either
- how / where do we define what / where the various components are?
  - API does not know about the Website at all
  - There is no such things as a **mandatory** worker
  - There is no really such things as S3 and Youtube defined yet
    - We have an S3 endpoint and policy defined in global config, to update the policy whenever worker IP is changing
    - We have many S3 URLs configured in recipes
    - We use "Youtube" but this is a mix of the official Youtube API (if there is such a thing) and unofficial endpoints used by yt-dlp
- shouldn't we also monitor (and again, what does it means to be UP/DOWN for each of them):
  - DB availability/responsiveness?
[...]

PR #1442, opened 2025-10-21
set up healthcheck API

elfkuzco:
# Rationale
This PR sets up the healthcheck service to monitor zimfarm components like the API and database. The results are displayed in a HTML page listing the status of each component. As of now, the checks done are:
- authentication can be done with API
- database can still receive connections
- at least one worker is still online (to be revised in a later PR/issue)

Screenshot_20251021_160218


elfkuzco:
Screenshot_20251021_160218

kelson42:
@elfkuzco Please share the details about what does this PR in the PR description (Focus on the user, the product.... not need to say with words how your code works). A screenshot is also welcome.

elfkuzco:
@kelson42 , I have updated the PR description to give an overview of what the PR achieves. What do you think about using creating sub-issues to properly discuss the remainder of the components like S3, Youtube, etc... and allowing this PR to close the original issue? I have discussed this with @benoit74

healthcheck/README.md
```diff
@@ -0,0 +1,10 @@
+## Healthcheck
+
+This service monitors the status of Zimfarm services and components and displays results as HTML
+
+### Environment variables
+
+- ZIMFARM_API_URL: Zimfarm backend API URL
+- ZIMFARM_USERNAME: Username to authenticate with on Zimfarm
+- ZIMFARM_PASSWORD: Password of user to authenticate with on Zimfarm
+- ZIMFARM_DATABASE_URL: Zimfam database URL
```
benoit74: `Zimfam` => typo + specify this should be full database URL including credentials

healthcheck/src/healthcheck/main.py
```diff
@@ -0,0 +1,42 @@
...
+from healthcheck.router import router
+
+
+def create_app(*, debug: bool = True):
+    app = FastAPI(
+        debug=debug,
+        docs_url="/",
+        title="Zimfarm Healthcheck Service",
+        version="1.0.0",
+        description="Service for checking health status of Zimfarm services",
```
benoit74: `Zimfarm services` => `Zimfarm components and dependencies`

healthcheck/src/healthcheck/main.py
```diff
@@ -0,0 +1,42 @@
...
+    )
+
+    if origins := getenv("ALLOWED_ORIGINS", default=None):
```
benoit74: Do we really need to setup CORS on this very basic UI? I don't get in which scenario we would get cross-origin requests.

healthcheck/src/healthcheck/router.py
```diff
@@ -0,0 +1,63 @@
...
+
+    global_status = all(
+        [
+            auth_check.success,
+            db_check.success,
+            workers_check.success,
+        ]
+    )
+
+    return templates.TemplateResponse(
```
benoit74, line 55: HTTP Status code returned must reflect the `global_status` value so that we can easily know with our UptimeRobot monitoring when the `global_status` is KO. We should use same HTTP codes than on imager health-check.

benoit74 (changes requested):
Besides few comments below, I feel like we've reached an acceptable "first version" which can be moved in production and already provide value.

@elfkuzco please proceed with opening issues with everything left to be done by this PR so that we can close #557 while knowing we have issues to tackle everything which is next. Include functional issues (add status for X, Y, Z to healthcheck, refine "worker status" healthcheck, ...) and more technical ones (cache statuses in memory, and maybe others). 

As agreed together, I would like we merge this soon so that we avoid ending with a never-ending PR. Especially since it looks we miss some details on some of these "next-issues".

benoit74:
@rgaudin I'm not asking for a formal review, but your feedback on this first implementation (which does not yet include the caching we've spoken about) is warmly welcomed, to be sure we are on the same page.

Also worth to mention: I imagine we will deploy this at `https://status.farm.openzim.org` or something like that, and run it as a "standalone" pod so that we can scale or shutdown the API without impact on the status monitoring, is that OK for you?

benoit74:
@elfkuzco please also update `Publish` CI job to publish a Docker image to GHCR repo.

elfkuzco:
added publish job but didn't add the k8s part used on backend since there's no deployment for it

rgaudin:
> @rgaudin I'm not asking for a formal review, but your feedback on this first implementation (which does not yet include the caching we've spoken about) is warmly welcomed, to be sure we are on the same page.

skimmed briefly through it ; looks OK. One thing that we've always missed on the cardshop is dedicated (custom) status code for individual tests failing. It can be some numbers on top of 500 that are sumed. The goal is for us to know what it's about directly from the `#uptime` alert in case it can be deduced. It seems like it's not that important but with the years of imager experience, I can guarantee it will be helpful. Especially for alerts that come and go at night. Should we try to _standardize_ this and thus have matching codes for imager service?

> Also worth to mention: I imagine we will deploy this at `https://status.farm.openzim.org` or something like that, and run it as a "standalone" pod so that we can scale or shutdown the API without impact on the status monitoring, is that OK for you?

Absolutely

benoit74:
Thank you @rgaudin!

> One thing that we've always missed on the cardshop is dedicated (custom) status code for individual tests failing. It can be some numbers on top of 500 that are sumed. The goal is for us to know what it's about directly from the #uptime alert in case it can be deduced. It seems like it's not that important but with the years of imager experience, I can guarantee it will be helpful. Especially for alerts that come and go at night. Should we try to standardize this and thus have matching codes for imager service?

Let's move this to a dedicated issue. Using HTTP return code is a hack and I feel like it is pretty fragile. Would like to discuss this further. I'm also not sure the need once #1455 is implemented: currently it is painful to get details about the alert on imager service because status is not cached, and so you have to wait, might get different results, ... ; once #1455 is implemented, the details are going to be only one click away. @elfkuzco can you please open an issue about this topic?

elfkuzco:
Done. See #1460

benoit74 (approved):
LGTM

PR #1450, opened 2025-10-22
add roles to users

elfkuzco:
## Rationale
Previously, user permissions for a particular resource were computed by indexing into their `scope` dictionary. However, we need to further extend the scope object for users in order to implement #1397, #1393. But it turns out the existing permissions system isn't particularly easy to extend as it would require changing each users scope dictionary at the database level. 

By introducing a column called `role` for users which maps directly to the `ROLES` dictionary, we make it easier to extend permissions for users by updating this dictionary at the application code. Newer permission checks would retrieve the "current" dictionary and only fallback to the one defined in the DB only if it's not a recognized role e.g "custom"

## Changes
- add role column to users in database
- retrieve permissions from application level role or fallback to the one defined in the DB

This PR is one part of #1419 which has been closed to apply changes incrementally.


elfkuzco:
This snippet updates the `roles` of users in the DB to the one in the enum or "custom"
```py
from zimfarm_backend.db import Session
from zimfarm_backend.common.roles import get_role_for
from zimfarm_backend.db.models import User
from sqlalchemy import select

with Session.begin() as session:
    for user in session.scalars(select(User)).all():
        role = get_role_for(user.scope)
        if role != "custom":
            user.role = role
            user.scope = None
        session.add(user)
```

This works correctly as is because the dictionary returned by `get_role_for` is the same one as the `user.scope`. Future PRs that extend this scope in the application code will be returning `custom`
https://github.com/openzim/zimfarm/blob/9cc38d42c494947613c448eafa2585928028c8cd/backend/src/zimfarm_backend/common/roles.py#L83-L87

benoit74 (changes requested):
Thank you.

When a user has a given `role` assigned, I would expect its `scope` value to be `null` inside DB, so that it does not get outdated when we modify the `role` assignments. I don't feel like this is what your code snippet does.

We should also add more tests ensuring user gets proper permission whether it comes from role or from (custom) scope. No need to test all permissions at all, but at test the `check_user_permission` for both situations (role vs custom scope).

backend/src/zimfarm_backend/db/models.py
```diff
@@ -69,6 +69,7 @@ class User(Base):
     password_hash: Mapped[str | None]
     email: Mapped[str | None]
     scope: Mapped[dict[str, Any] | None]
+    role: Mapped[str | None]
```
benoit74: Let's say `role` should never be `None`, and use `custom` when we want to use the `scope` to have a custom role. I feel like this will simplify code forward (especially in the UI)

elfkuzco:
- regenerated alembic migration script so `role` is non-nullable with server default of `custom`
- updated snippet to use the `get_role_for` to unset users that are not  `custom`
- added tests for check_permissions to verify it retrieves from application defined permissions for known roles and db scope for custom users

elfkuzco:
fixed

backend/src/zimfarm_backend/db/user.py
```diff
@@ -67,9 +68,11 @@ def check_user_permission(
...
-    if user.scope is None:
+    # Select the scope that comes with their role enum or scope from the DB
+    scope = ROLES.get(user.role, user.scope)
+    if not scope:
         return False
-    return user.scope.get(namespace, {}).get(name, False)
+    return scope.get(namespace, {}).get(name, False)
 
 
 def update_user_password(
```
benoit74, line 78: Method name is now misleading

backend/src/zimfarm_backend/db/user.py
```diff
@@ -119,14 +122,16 @@ def create_user(
     username: str,
     email: str,
     password_hash: str,
-    scope: dict[str, Any],
+    scope: dict[str, Any] | None,
+    role: str,
 ) -> User:
     """Create a new user"""
     user = User(
```
benoit74, line 131: We should check here that if role is not `custom`, then scope must be `None`

backend/src/zimfarm_backend/routes/auth/logic.py
```diff
@@ -49,7 +50,7 @@ def _access_token_response(db_session: OrmSession, db_user: User, response: Resp
             user_id=str(db_user.id),
             issue_time=issue_time,
             username=db_user.username,
-            scope=db_user.scope,
+            scope=ROLES.get(db_user.role, db_user.scope),
```
benoit74: Looks like you've spotted a bug which was not catch in any test (I'm not sure since you've force-pushed). If you confirm, please add a test to ensure that scope in access token is properly populated even if user only has a role.

elfkuzco:
> Btw, you should also really avoid to force-push after first review has been requested and until final approval (where you can of course force-push to rebase + squash).

Okay

backend/src/zimfarm_backend/db/user.py
elfkuzco: Sorry, what should we call it?

benoit74: `update_user` ? (we do not update only the password)

elfkuzco: This function only updates user password
https://github.com/openzim/zimfarm/blob/8ecdaf0690518f430169a1b92eaeb28112ae3271/backend/src/zimfarm_backend/db/user.py#L75-L84

This one updates other details
https://github.com/openzim/zimfarm/blob/8ecdaf0690518f430169a1b92eaeb28112ae3271/backend/src/zimfarm_backend/db/user.py#L139-L148

They are both used by different endponts:
https://github.com/openzim/zimfarm/blob/8ecdaf0690518f430169a1b92eaeb28112ae3271/backend/src/zimfarm_backend/routes/users/logic.py#L135-L142

and 
https://github.com/openzim/zimfarm/blob/8ecdaf0690518f430169a1b92eaeb28112ae3271/backend/src/zimfarm_backend/routes/users/logic.py#L230-L237

benoit74: OK, sorry, never mind, looks like I got confused while reading Github diff, everything is fine. Sorry.

PR #1451, opened 2025-10-23
redefine permissions to access resource

### LINKED ISSUE #1358: Add new permission allowing a user to register as a worker

### LINKED ISSUE #1393: Revisit permissions view

elfkuzco:
## Rationale
Previous permissions overlapped e.g permission to check in worker verified if a user could create another user. By defining new permissions specifically for each resource, the permissions system is further fine-grained. It further lays the foundation for #1397 where we need specific permissions to be able to view a worker's secret details in the UI.

## Changes
- add new permissions to further distinguish resources access/update
- add permissions to reveal secrets
- redesign view for permissions list

Screenshot_20251023_035527

This closes #1358 
This closes #1393

### DIFF
--- a/backend/src/zimfarm_backend/common/roles.py
+++ b/backend/src/zimfarm_backend/common/roles.py
@@ -16,8 +16,8 @@ class Permissions:
 
 class TaskPermissions(Permissions):
     names: ClassVar[list[str]] = [
-        "request",
-        "unrequest",
+        "read",
+        "secrets",
         "create",
         "update",
         "cancel",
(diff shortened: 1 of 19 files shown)


benoit74 (approved):
LGTM

elfkuzco:
Linking #1358 as it would be closed by this PR too

PR #1453, opened 2025-10-23
strip spaces from input

### LINKED ISSUE #1425: Most (all?) config values should be trimed
Image

I have just put a trailing whitespace to a MWoffliner recipe article list... and it seems it will be saved. The problem here is that trailing whitespace is "invisible", does not make sense, but can bring quite a few of problem later in the toolchain.

We should avoid this.

elfkuzco:
#Rationale
Configuration values can contain leading/trailing spaces which are undesirable. This also leads to the editor detecting an additonal space as a change.
## Changes
- strip away leading/trailing spaces from `NotEmptyString` type which is the underlying type for string values. Due to the `model_construct` used while reading fields from DB, existing entries that have leading/trailing spaces won't be trimmed
- use vue's `v-model.trim` to trim spaces in text and text area inputs

This closes #1425

### DIFF
--- a/backend/src/zimfarm_backend/common/schemas/fields.py
+++ b/backend/src/zimfarm_backend/common/schemas/fields.py
@@ -56,7 +56,7 @@ def not_empty(value: str) -> str:
     if not value.strip():
         raise ValueError("String value cannot be empty")
 
-    return value
+    return value.strip()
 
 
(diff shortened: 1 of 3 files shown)


benoit74 (approved):
LGTM

benoit74:
Oups, there is a test to fix first, missed that.

PR #1454, opened 2025-10-23
fix bug in schedule editor reset button

elfkuzco:
## Rationale
The reset button contains a button of type `reset` which [resets](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/input/reset) the form's values in an undesirable way as it clears everything. Also, previously, while handling reset, offliner flags were not reset to original before checking if there was a version change. That call cached the http request and didn't update the flag definitions. This meant that if the version didn't change, the edited flags were still in place instead of being reverted

## Changes
- change reset button type from `reset` to `button`
- reset schedule edit flags to original flags while handling reset

### DIFF
--- a/frontend-ui/src/components/ScheduleEditor.vue
+++ b/frontend-ui/src/components/ScheduleEditor.vue
@@ -10,7 +10,7 @@
         Update Offliner details
       </v-btn>
       <v-btn
-        type="reset"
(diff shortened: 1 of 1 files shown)


benoit74 (approved):
LGTM

PR #1464, opened 2025-10-23
support http for S3 uploads

elfkuzco:
## Rationale
The uploader uploads items to S3 using only HTTPS. This proves undesirable especially while testing locally with another object store like Minio and you don't want to set up HTTPS. This PR aims to address this challenge by providing two additional ways of specifying the HTTP scheme to be used with S3: `s3+http` and `s3+https`. Default of `s3` is HTTPS for backwards compatibility.

## Changes
- Define S3 urls with http scheme


benoit74 (approved):
LGTM

PR #1466, opened 2025-10-24
use artifacts expiration globs during task creation

### LINKED ISSUE #1035: Really use artifacts expiration setting
Currently, the artifacts expiration setting is saved in the task, together with the upload link. But in the worker, the expiration setting is never used by the task worker which should pass this information to the uploader to set proper retention rule.

elfkuzco:
## Rationale
Previously, the artifacts globs were ignored while creating tasks. This meant that no artifacts would ever be created because the list was always empty. By using the same artifacts globs when the requested task was created, we ensure the globs goes through the task worker

## Changes
- use artifacts globs during task creation

This fixes #1035


benoit74:
This is weird, because it worked in the past. 

So from my understanding what you've fixed here looks more like a "recent" regression (linked to past month changes), and #1035 needs no fix besides correctly configuring the bucket.

Is that possible?

elfkuzco:
Yes, I think so.

PR #1467, opened 2025-10-24
use diskcache to memoize check results

### LINKED ISSUE #1455: Cache healthcheck service checks
The healthcheck service should use some form of caching mechanism while doing the checks on the zimfarm components. Suggested library for caching is [diskcache](https://grantjenks.com/docs/diskcache/).

Several component checks should have a timeout for cache key expiry as they have different response times.

elfkuzco:
## Rationale
The healthcheck service performs all the checks each time which might be overwhelming for the zimfarm and its components. By caching the results of each call with a TTL, we can avoid making checks in quick succession.

## Changes
- use diskcache [FanoutCache](https://grantjenks.com/docs/diskcache/tutorial.html#fanoutcache) to store results of each check. As per the docs, results are cached using pickle
- add decorator function to cache checks only if they are successful
- provide default ttl allowing different checks to have different ttl for their results

This closes #1455

### DIFF
--- a/healthcheck/src/healthcheck/cache.py
+++ b/healthcheck/src/healthcheck/cache.py
@@ -52,6 +52,9 @@ def memoize(
         async def wrapper(*args: P.args, **kwargs: P.kwargs) -> R:
             cache = init_cache()
             location = f"{CACHE_KEY_PREFIX}:{key}"
+            # Types other than the basic types like floats, ints, bytes, strings are
+            # are stored using pickle by default. Thus, we can save our results
+            # (pydantic models) directly to the cache and get it back as is.
             if (result := cache.get(location)) is not None:
                 return result
 


elfkuzco:
PR is not yet complete but more of a PoC to demonstrate the implementation. Will need some inputs on what TTL to use for each check

healthcheck/src/healthcheck/cache.py
```diff
@@ -0,0 +1,58 @@
...
+from typing import ParamSpec, TypeVar
+
+from diskcache import FanoutCache
+
+from healthcheck.constants import CACHE_KEY_PREFIX, CACHE_LOCATION, DEFAULT_CACHE_TTL
+
+P = ParamSpec("P")
+R = TypeVar("R")
+
+_cache: FanoutCache | None = None
```
rgaudin, line 20: Why a `FanoutCache` and not another one? We should comment hbere to record rationale.

healthcheck/src/healthcheck/cache.py
```diff
@@ -0,0 +1,58 @@
...
+def close_cache() -> None:
+    """Close the disk cache instance."""
+    global _cache  # noqa: PLW0603
+    if _cache is not None:
+        _cache.close()
+        _cache = None
+
+
+def memoize(
+    key: str, ttl: float = DEFAULT_CACHE_TTL
```
rgaudin: Not a fan of the `ttl` abbreviation because we'll use it in a request context where TTL has a different meaning in network than _expiration_

healthcheck/src/healthcheck/cache.py
```diff
@@ -0,0 +1,58 @@
...
+    """Memoize successful results of functions with results at CACHE_KEY_PREFIX:key.
+
+    Results are considered successful if ther success attribute is truthy.
+    """
+
+    def decorator(func: Callable[P, Awaitable[R]]) -> Callable[P, Awaitable[R]]:
+        @wraps(func)
+        async def wrapper(*args: P.args, **kwargs: P.kwargs) -> R:
+            cache = init_cache()
+            _key = f"{CACHE_KEY_PREFIX}:{key}"
```
rgaudin: Leading `_` is convention for private, here you'd want `key_` or something more explicit I believe

healthcheck/src/healthcheck/cache.py
```diff
@@ -0,0 +1,58 @@
...
+
+            if (result := cache.get(_key)) is not None:
+                return result
+
+            result = await func(*args, **kwargs)
+            if (
```
rgaudin: I like that it's clearly indicated that we only cache success. I'm wondering if that's what we always want. Maybe that could be a parameter as well…

elfkuzco:
>  but it's unclear to me how we're caching the Result object. Is is serialized? How? How are we unserializing it?

This is handled by the library. The default implementation uses pickle to store data types that are not natively supported like bytes, string, floats, etc. This is how we store the pydantic objects and retrieve them directly since each check is a `Result` object

elfkuzco:
> Why a FanoutCache and not another one? We should comment hbere to record rationale.

I opted to use it because it's the same as what the DjangoCache uses. Figured since it is used in a web framework, it should suffice here. It has support for sharding than the default Cache. The docs say:

"While readers and writers do not block each other, writers block other writers. Therefore a shard for every concurrent writer is suggested. This will depend on your scenario."

So, I figured since we call the results with `asyncio.gather`, we don't want another writer to block another. I suppose I should add the rationale in the code too.

healthcheck/src/healthcheck/cache.py
elfkuzco: Renamed parameter to `expire`

healthcheck/src/healthcheck/cache.py
elfkuzco: Added comment in code

healthcheck/src/healthcheck/cache.py
elfkuzco: Named variable to `location`

healthcheck/src/healthcheck/cache.py
elfkuzco: Added a parameter to `cache_only_on_success` with default of `True`

rgaudin:
> > but it's unclear to me how we're caching the Result object. Is is serialized? How? How are we unserializing it?
> 
> This is handled by the library. The default implementation uses pickle to store data types that are not natively supported like bytes, string, floats, etc. This is how we store the pydantic objects and retrieve them directly since each check is a `Result` object

Good ; please add a small comment to make this clear

rgaudin:
> > Why a FanoutCache and not another one? We should comment hbere to record rationale.
> 
> I opted to use it because it's the same as what the DjangoCache uses. Figured since it is used in a web framework, it should suffice here. It has support for sharding than the default Cache. The docs say:
> 
> "While readers and writers do not block each other, writers block other writers. Therefore a shard for every concurrent writer is suggested. This will depend on your scenario."
> 
> So, I figured since we call the results with `asyncio.gather`, we don't want another writer to block another. I suppose I should add the rationale in the code too.

Good, a comment with this would be perfect

elfkuzco:
> I don't see why we want to store data on disk using an SQLite DB. Check results do not need to be persisted from my understanding so far, and adding such persistence layer is only unneeded accidental complexity from my perspective so far (which can only lead to useless bugs if we confirm it does not add value). Can you develop why we would need this?

It seems to be an [implementation detail](https://grantjenks.com/docs/diskcache/tutorial.html#implementation) of the library and I'm not really in control of it.

elfkuzco:
> I finally do not get how this PR achieves to fulfill requirements I've expressed. With this PR, as soon as the cache has expired if we have multiple parallel calls to the check I feel like the code will still run multiple checks in parallel (cache is expired so code will be ran). I would expect that checks are periodically updated but we continue to serve old status until new status has been retrieved.

Do you mean this should be run periodically by a cron job and not from the HTTP request? Because it's probably only a caching mechanism (even if not this one) that prevents the checks from happening each time a request is made. The checks can be configured with different intervals but at some future time, all the intervals will meet and they will be run in parallel. Running "periodically" would still be susceptible to parallelism.

elfkuzco:
> I also don't get why we cache only on success. In case of failure, this is where everyone will "rush" (to a moderate extend, we do not have that much success, but still) to check status, and we do not want to have parallel / too much repeated calls to the check which would only overwhelm the upstream system which is already failing.

This is quite parametrizable and I can flip the switch to false if that shouldn't be the normal.

elfkuzco:
> I finally miss the capability to force refreshing the check status.

I will add a query parameter to the checks that will force refresh

benoit74:
@elfkuzco please wait for discussion on https://github.com/openzim/overview/issues/72 to settle, we've discussed a bit with @rgaudin and confirm we miss the proper big picture to correctly implement things in this PR.

kelson42:
@benoit74 Just want to say that I mostly share the opinion here. Should be possible to have a better software architecture than this.

PR #1475, opened 2025-10-28
make role and scope mutually exclusive

### LINKED ISSUE #1465: Enhance API and UI to allow customizing user permission scope
Currently, API and UI only allows to update user to a given role. It is not possible to set a custom permission scope.

We need to allow admins to set custom scope through the UI.

Custom scope will be input in a free text area, directly as JSON, for simplicity. This seems sufficient since only used by admins.

The text area should only be editable when the user scope is set to "custom". 

It should contain user scope for all roles (so that it is easier to copy/paste from a standard role to a custom one).

No validation of the JSON scope will be done in a first step.

### LINKED ISSUE #1471: Permissions: display all possibles permissions
See https://farm.openzim.org/users/healthcheck

Image

It looks like only the `workers:read` permission exists in DB.

We should find a way to display all existing permissions, even if these permissions do not exist in current user scope. Probably just explore all keys which exists on pre-defined roles when fetching user scope and populating it with these keys set to False.

### LINKED ISSUE #1472: Permission: set scope to `None` when assigning a pre-defined role to a custom
When I assign in pre-defined role to an existing user which had a custom role, the pre-existing "scope" is still there in DB. 

It must be set to `None`.

Image

elfkuzco:
## Rationale
This PR makes enhances the user roles by making sure that known roles have no scope at DB level while custom scopes contain specific scope. It also provides the UI with the capability of updating a specific user not just from a known role but fine-tuning their permissions for custom role.

## Changes
- update user role/scope from API/UI
- list all user permissions for a resource by merging their permissions with the one from the ADMIN role instead of returning the subset that they have
- set scope to `None` when assigning a pre-defined role
- set role to custom when assigning scope

Below is a screenshot listing all permissions for a user even when they don't have that permission set on their scope. This happens because of the merge functionality that merges a user's scope or the scope attached to their role to the admin scope which contains every permission. 

Screenshot_20251028_165314

This fixes #1472 
This closes #1465
This closes #1471

### DIFF
--- a/backend/src/zimfarm_backend/common/roles.py
+++ b/backend/src/zimfarm_backend/common/roles.py
@@ -84,7 +84,7 @@ class RoleEnum(StrEnum):
     PROCESSOR = "processor"
 
 
-ROLES: dict[str, dict[str, Any]] = {
+ROLES: dict[str, dict[str, dict[str, bool]]] = {
     RoleEnum.ADMIN: {
         "tasks": TaskPermissions.get_all(),
         "schedules": SchedulePermissions.get_all(),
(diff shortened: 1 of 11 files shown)


elfkuzco:
- Added `secrets=True` for editor-requester on `tasks`, `schedules`, `requested_tasks`. The rationale is that if they can create schedules, then they need secrets. By extension, `tasks` and `requested_tasks`

PR #1482, opened 2025-10-30
allow requested tasks for disabled recipes to be unrequested from UI

### LINKED ISSUE #1474: Allow to unrequest disabled recipe
It is possible to have a requested task pending for a disabled recipe. UI should allow to unrequest this task. Currently it is not possible, button is not shown.

How to reproduce:
- request a schedule
- disable it
- unrequest button is gone

elfkuzco:
## Rationale
The action buttons to request/unrequest tasks were only shown if a recipe was enabled and not archived. This meant that if a recipe is requested and later disabled, it couldn't be unrequested from the UI. This is undesirable. By relaxing the conditions to show the button but pass the information to the component, we can control the granularity of the display of the buttons. Using this information, we only show the request button if a schedule is archived and the unrequested button if there is a requested task irrespective of if it is enabled.

## Changes
- show action buttons irrespective of schedule enabled or archived status
- show unrequest button irrespective of status

This fixes #1474

### DIFF
--- a/frontend-ui/src/components/ScheduleActionButton.vue
+++ b/frontend-ui/src/components/ScheduleActionButton.vue
@@ -14,7 +14,7 @@
 <template>
   <div v-if="visible" class="d-flex justify-end">
     <!-- Worker Selection Dropdown -->
-    <v-menu v-show="canSelectWorker" location="bottom end" offset-y>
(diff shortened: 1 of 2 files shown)


benoit74 (approved):
LGTM

PR #1484, opened 2025-10-30
document worker shutdown process

### LINKED ISSUE #1479: Add documentation about process to shutdown a worker
We need a clear documentation in `worker/README.md` about how to cleanly shutdown a worker:

- explain usage of `CORDONED` env variable
- explain that once cordoned, no more new tasks will be started on the worker and that worker owner should wait for running tasks to complete
- explain that on a regular basis we expect worker to not be shutdown more than once a month
- explain that we don't mind much about the duration of the shutdown, that it is way more important to avoid too many shutdowns rather than long shutdowns (to limit risk of tasks being interrupted)

elfkuzco:
## Changes
- add documentation about worker manager shutdown

This closes #1479

### DIFF
--- a/worker/README.md
+++ b/worker/README.md
@@ -129,8 +129,9 @@ We do recommend to rotate this SSH key once a year. Zimfarm allows to have multi
 ### Create worker account on the Zimfarm
 
 Send your worker information to to Zimfarm Admins via mail [contact+zimfarm@kiwix.org](mailto:contact+zimfarm@kiwix.org) or on Slack:
-- worker name (only lowercase letter, see already in-use worker names at https://farm.openzim.org/workers, avoid generic names like "cloud-vm", "zimfarm-worker", ...) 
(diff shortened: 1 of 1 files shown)


worker/README.md
```diff
@@ -215,3 +216,71 @@ If you feel like you want to meet these expectations, feel free to tell Zimfarm
...
+### Waiting for Tasks to Complete
+
+Once your worker is cordoned:
+
+1. Monitor running tasks with `zimfarm ps`
+2. Check task logs if needed with `zimfarm logs <task-name>`
+3. **Wait patiently** for all running tasks to complete naturally
+4. Do not force-stop tasks unless absolutely necessary
+
+Some tasks can take days to complete. It's critical to let them finish rather than canceling them, as this wastes resources and the task will need to be restarted on another worker.
```
benoit74: `critical` => `important`

`to be restarted on another worker` => `to be manually restarted on another worker`

PR #1488, opened 2025-10-30
prevent overrriding cancel_requested events

### LINKED ISSUE #937: Task cancellation can be overriden by other task events
When a task is marked for cancellation, this request might be overridden by another task even. See screenshot below for one occurence. First `cancel_requested` has been overridden by `scraper_started` and I had to request the cancellation again.
 
image

Not sure yet if this is only a UI bug or if the cancellation is really lost in such a situation.

elfkuzco:
## Rationale
Currently, task events can be overriden by other task events. While most updates happen in defined manner once started by a worer, it is problematic if an already picked up task is requested for cancellation because the worker will continue with the task as if things were normal. By returning to the worker, the status of the task each time they make a patch, it serves as a form of communication to let the worker know that the update went according to what they expected. For now, we simply check if the status is not cancel_requested because that's one not sent by the worker and the point of conflict. 

## Changes
- return status of tasks after `PATCH` request
- exit gracefully from task worker if after patching, status is cancel_requested.

This closes #937

### DIFF
--- a/worker/src/zimfarm_worker/manager/worker.py
+++ b/worker/src/zimfarm_worker/manager/worker.py
@@ -265,11 +265,16 @@ class WorkerManager(BaseWorker):
                 continue  # already handling cancellation
 
             self.update_task_data(task_ident)
-            if self.tasks.get(task_ident, {}).get("status") in [
-                CANCELED,
-                CANCELING,
-                CANCEL_REQUESTED,
-            ]:
+            logger.debug(f"Checking if task {task_ident} should be cancelled...")
+            event_codes = {
+                event["code"]
+                for event in self.tasks.get(task_ident, {}).get("events", [])
+            }
+            cancel_events = {CANCEL_REQUESTED, CANCELED}
+            if (
+                self.tasks.get(task_ident, {}).get("status") in cancel_events
+                or event_codes & cancel_events
+            ):
                 self.cancel_and_remove_task(task_ident)
 
     def cancel_and_remove_task(self, task_ident: TaskIdent):


benoit74 (commented):
I don't feel like this is the correct implementation because:
- it introduces a new way for the task manager to exit, which seems risky (pretty sure we will have side-effects, I would prefer task manager always terminates the same way when cancellation is requested)
- it completely drops `started`, `scraper_started`, `scraper_running` and`scraper_completed` events, while they represent the reality (e.g. scraper has started or completed) and could be useful to understand the situation in some cases

After some thought, I feel like we could fix this issue only at the worker side by looking for the existence of a `cancel_requested` event in the task events to decide to stop a given task. WDYT?

https://github.com/openzim/zimfarm/blob/64899d923a6dec8927144acd6beadf40edd501dd/worker/src/zimfarm_worker/manager/worker.py#L268-L273

elfkuzco:
> After some thought, I feel like we could fix this issue only at the worker side by looking for the existence of a cancel_requested event in the task events to decide to stop a given task. WDYT?

You mean, omitting the API side of the check and new response? Because what happens is that (from the linked issue and  a simulation I ran) is that the manager maintains a stale list of tasks in `self.tasks`. So, by the time it probably started that task, it got canceled and the first thing the worker is going to do is update with `scraper_started`.

benoit74:
Yes, this is what I mean. Because the scraper has really started so we should record "scraper_started" in the list of events when it is received, as well as any other event received in fact (including potential logs, ...). Processing the list of task events in the worker manager to check if it contains "cancel_requested" seems to be sufficient to cancel scraper. It will take a bit more time, but we should probably not care because it always takes time to cancel a task. And it seems way simpler in terms of code. Or do I miss something?

worker/src/zimfarm_worker/manager/worker.py
```diff
@@ -265,11 +265,16 @@ def check_cancellation(self):
...
-                CANCELED,
-                CANCELING,
-                CANCEL_REQUESTED,
-            ]:
+            logger.debug(f"Checking if task {task_ident} should be cancelled...")
+            event_codes = {
+                event["code"]
+                for event in self.tasks.get(task_ident, {}).get("events", [])
+            }
+            cancel_events = {CANCEL_REQUESTED, CANCELING, CANCELED}
```
benoit74: CANCELING does not exists at DB level, it cannot be among event codes.

PR #1489, opened 2025-10-30
update zimfarm shell script update URL and worker README

elfkuzco:
## Changes
- update source url for zimfarm update command
- remove `id_rsa` mentions since keys are now expected to be Ed25519 keys
- update README to specify minimum script version while attempting to CORDON worker

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,7 +28,9 @@ Anybody can use it in read-only mode.
 
 ### workers
 
-Workers are always-running computers which gets assigned ZIM creation tasks by the dispatcher. If you are interested in providing us worker resources, please [read these instructions](https://github.com/openzim/zimfarm/blob/main/workers/README.md).
+Workers are always-running computers which gets assigned ZIM creation tasks by the dispatcher. If you are interested in providing us worker resources, please [read these instructions](https://github.com/openzim/zimfarm/blob/main/worker/README.md).
+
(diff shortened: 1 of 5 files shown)


elfkuzco:
Aaah. It will empty the script. Apparently, `wget` doesn't fail with the `-O` flag. Should we send them a slack message to update with the proper URL?

elfkuzco:
Or we could update the repo side by simply renaming the `worker` name of the tree to `workers`? Their `zimfarm update do` would still work.

benoit74:
We've already made the move from `workers` to `worker` and it was a painful one since we have too many places to update (code but also wiki, ...). Going back to `workers` is a bit sad. Maybe we should just temporarily store the `zimfarm.sh` in both `worker` and `workers` tree, with a clear README about the reason in `workers` tree. Ask for all owners to update their scripts. And then drop the `workers` tree in something like 1 month or so.

I also propose we replace the URL in `zimfarm.sh` with something more stable / easy to manipulate like `https://farm.openzim.org/contrib/zimfarm.sh` which would in turn would redirect to github or anything else. @rgaudin does it makes sense to you as well? Do you have suggestions in terms of stable URL to use?

We also need to find a way so that `zimfarm upgrade do` does fail if URL is not properly getting a bash script.

rgaudin:
Makes sense 👍

benoit74:
https://farm.openzim.org/contrib/zimfarm.sh is ready to be used.

@elfkuzco please proceed with:
- update zimfarm.sh to use this URL
- host zimfarm.sh in both worker and workers folder temporarily with an explanation about why workers folder still exist
- enhance zimfarm.sh `update do` and `update` command to fail if script is not found at given URL (quite sure it is possible with wget)

elfkuzco:
Ended up using `curl` because wget still pipes input to output file even on failure.

PR #1491, opened 2025-11-01
measure task resources

### LINKED ISSUE #1423: Measure and report all tasks resources usage
Currently, we have no clue about how well fit the recipes resources are.

We typically:
- over-reserve memory because this is going to cause OOM errors otherwise
- have no clue about CPU and disk

One typical problem is that we increase memory requirements when the scraper demands more resources, but never decrease it should it have been a bug / transient behavior which has been fixed.

While the "monitor" system is great to have details about resource consumption, we cannot activate it at all times because of permission issues on workers and huge associated data volume.

I feel like we could implement something a bit simpler and lighter by having the task manager regularly (once every 2 secs for instance) probe the scraper container usage and store maximum usage seen during the whole scrape.

And display clear warning in task details when there is discrepancies between task reservation and task usage.

And we could then on a regular basis review and adjust tasks requirements.

@kelson42 WDYT?

elfkuzco:
## Rationale
This PR adds support for measuring resources used by the scraper. For the CPU stats, it uses the [Exponentially Weighted Moving Average](https://corporatefinanceinstitute.com/resources/career-map/sell-side/capital-markets/exponentially-weighted-moving-average-ewma/) to measure the percentage of CPU usage, along with the maximum CPU percentage used.

Also, max disk usage is computed by adding the filesizes of the files in the scraper mount directory.
Screenshot_20251114_133311
Screenshot_20251114_133301

## Changes
- add functions  to compute CPU and Disk stats
- show stats in UI

This closes #1423

### DIFF
--- a/backend/src/zimfarm_backend/common/schemas/orms.py
+++ b/backend/src/zimfarm_backend/common/schemas/orms.py
@@ -113,14 +113,18 @@ class TaskContainerProgressSchema(BaseModel):
     total: int | None = None
 
 
-class TaskContainerMemoryStatsSchema(BaseModel):
-    max_usage: int | None = None
+class TaskResourceUsageSchema(BaseModel):
+    max_usage: int | float | None = Field(default=None, alias="max")
 
 
-class TaskContainerStatsSchema(BaseModel):
-    memory: TaskContainerMemoryStatsSchema = Field(
-        default_factory=TaskContainerMemoryStatsSchema
-    )
+class TaskCPUUsageSchema(TaskResourceUsageSchema):
+    avg_usage: float | None = Field(default=None, alias="avg")
+
+
+class TaskStatsSchema(BaseModel):
+    memory: TaskResourceUsageSchema = Field(default_factory=TaskResourceUsageSchema)
+    cpu: TaskCPUUsageSchema = Field(default_factory=TaskCPUUsageSchema)
+    disk: TaskResourceUsageSchema = Field(default_factory=TaskResourceUsageSchema)
 
 
 class TaskContainerSchema(BaseModel):
@@ -130,7 +134,7 @@ class TaskContainerSchema(BaseModel):
 
     log: str | None = None
     image: str | None = None
-    stats: dict[str, Any] | None = None
+    stats: TaskStatsSchema | None = None
     artifacts: str | None = None
     stderr: str | None = None
     stdout: str | None = None
--- a/worker/src/zimfarm_worker/task/worker.py
+++ b/worker/src/zimfarm_worker/task/worker.py
@@ -40,6 +40,9 @@ from zimfarm_worker.common.worker import BaseWorker
 from zimfarm_worker.task.zim import get_zim_info
 
 SLEEP_INTERVAL = 60  # nb of seconds to sleep before watching
+CPU_EWMA_ALPHA = 0.01  # EWMA smoothing factor for CPU percentage samples (0..1)
+
+
 PENDING = "pending"
 DOING = "doing"
 DONE = "done"
@@ -132,6 +135,9 @@ class TaskWorker(BaseWorker):
         self.scraper_succeeded: bool | None = None  # whether scraper succeeded
 
         self.max_memory_usage: int = 0  # maximum memory used by scraper
+        self.max_disk_usage: int = 0  # maximum disk used by scraper
+        self.avg_cpu_usage: float = 0.0  # cpu exponential moving weighted average
+        self.max_cpu_usage: float = 0.0  # maximum cpu percentage used by scraper
 
         # register stop/^C
         self.register_signals()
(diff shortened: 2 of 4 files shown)


elfkuzco:
Implementation of disk usage is yet to be added as it would rely on the approval of docker/docker-py#3370

benoit74:
As discussed on Slack, I propose we wait few days for https://github.com/docker/docker-py/ maintainers to give us an answer. 

If they don't reply soon enough, I propose two possible plans:
- consider only mounts size in disk usage stat for the time being, since this is in general the "core" of disk usage, at least for big ZIMs which are primary concern
- reimplement our own very-limited Docker SDK for only the operations we need ; could make since because Python Docker SDK seems to receive little attention from Docker, plus we only use few methods and could plug directly to the REST API just like the SDK does for these few methods we are using

benoit74:
@elfkuzco looks like I was right been concerned about not getting any feedback on your upstream PR.

Please advise which plan B (among the two I've proposed, or one you can propose) makes more sense to you so that we can move on and have CPU measure and at least a first estimate of Disk used.

elfkuzco:
> consider only mounts size in disk usage stat for the time being, since this is in general the "core" of disk usage, at least for big ZIMs which are primary concern

this would be simpler to implement plus, given almost everything is written to the mount point, i don't know if there's going to be any real metric obtained from writable layer. possibly .pyc files  or __pycache__ files but those really shouldn't be big enough, right?

benoit74:
Let's go for this alternative: consider only mount point for the time-being + open an issue about the fact that we might want to better track disk usage. Goal would be not only to capture the writable layer (which in general is supposed to be small, but this is not the case on all scrapers, not even speaking about bugs) but also the image size itself (which is then an slight overestimation of disk usage since image is shared across tasks).

elfkuzco:
Updated PR description.

worker/src/zimfarm_worker/task/worker.py
```diff
@@ -39,6 +39,9 @@
 from zimfarm_worker.task.zim import get_zim_info
 
 SLEEP_INTERVAL = 60  # nb of seconds to sleep before watching
+CPU_EWMA_ALPHA = 0.25  # EWMA smoothing factor for CPU percentage samples (0..1)
```
benoit74: How did you choose this `0.25` value? Worth explaining, at least in the PR

worker/src/zimfarm_worker/task/worker.py
```diff
@@ -191,17 +250,32 @@ def submit_scraper_progress(self):
...
+        self.max_disk_usage = max([disk_usage, self.max_disk_usage])
+
         stats: dict[str, Any] = {
             "memory": {
                 "max_usage": self.max_memory_usage,
-            }
+            },
+            "cpu": {
+                "max_usage": self.max_cpu_usage,
+                "ewma": round(self.cpu_ewma, 2),
```
benoit74: `ewma` => `avg_usage`

worker/src/zimfarm_worker/task/worker.py
```diff
@@ -119,6 +122,9 @@ def __init__(
         self.scraper_succeeded: bool | None = None  # whether scraper succeeded
 
         self.max_memory_usage: int = 0  # maximum memory used by scraper
+        self.max_disk_usage: int = 0  # maximum disk used by scraper
+        self.cpu_ewma: float = 0.0  # cpu exponential moving weighted average
```
benoit74: I would prefer `avg_cpu_usage` (the fact that we currently usage EWMA is an implementation detail at this stage)

backend/src/zimfarm_backend/common/schemas/orms.py
```diff
@@ -110,14 +110,18 @@ class TaskContainerProgressSchema(BaseModel):
...
+class TaskResourceUsageSchema(BaseModel):
+    max_usage: int | float | None = Field(default=None, alias="max")
 
 
-class TaskContainerStatsSchema(BaseModel):
-    memory: TaskContainerMemoryStatsSchema = Field(
-        default_factory=TaskContainerMemoryStatsSchema
-    )
+class TaskCPUUsageSchema(TaskResourceUsageSchema):
+    ewma: float | None = Field(default=None, alias="avg")
```
benoit74: `ewma` => `avg_usage`

worker/src/zimfarm_worker/task/worker.py
elfkuzco: Read from a few links and articles that choosing lower values are used to detect smaller mean changes. Not entirely familiar with it and randomly chose 0.25. Figured I would get your opinion on what to use.

benoit74: Let's use `0.01` for the time being then. According to maths, it looks like `0.01` with EWMA updates every minutes (default `SLEEP_INTERVAL`) means the EWMA will respond to variations on a hour-scale (mostly totally ignoring variations which do not take as long), which is probably what I would want for a scraper.

PR #1493, opened 2025-11-03
parse resource values as binary multiples of bytes

elfkuzco:
## Rationale
Worker resources are often parsed using `parse_size` function of the humanfriendly library. By default, these sizes are parsed in base 10 which might make a worker unable to execute a task. For example:
```py
from humanfriendly import parse_size
>>> parse_size('20Gb')
20000000000

>>> parse_size('20Gb', binary=True)
21474836480
```
This seems to be the cause of openzim/wp1#1027 where a worker has 20Gb but can't execute a task with 20Gb


audiodude:
Do I just re run the docker compose graph with always pull and I should pick up this change?

elfkuzco:
> Do I just re run the docker compose graph with always pull and I should pick up this change?


Yes. With --build

PR #1497, opened 2025-11-05
split background tasks from backend api

### LINKED ISSUE #1461: Move periodic task to another container
Currently, periodic tasks to automatically request tasks and to cleanup the Zimfarm (gone tasks, ...) are ran in the `backend` (API) container.

This is not a ideal solution because it means that:
- when we scale this container because we get more incoming API traffic, we get duplicate periodic tasks in parallel
- when the periodic task runs, it consumes load on the API while these periodic tasks are not urgent at all and way less important than responding to API calls in a timely manner

https://github.com/openzim/zimfarm/issues/1436 will also need a new kind of periodic task to (re)send CMS notifications.

I suggest it is time to implement a background tasks "manager" for the Zimfarm.

elfkuzco:
## Rationale
As expressed in #1461, there are cons to having the periodic task and the backend API in one container. This PR splits the backend into two, a REST API server and the background tasks. These tasks are configured with various timeouts. In doing so, the API can be scaled without having the background tasks effect happening in parallel.

## Changes
- seperate backend into api and background_tasks folder
- add seperate Dockerfiles for both services
- add both services to the dev compose
- uncomment services in the dev compose. With [Docker profiles](https://docs.docker.com/compose/how-tos/profiles/), users should be able to selectively run services
- modify CI jobs for backend API
- add CI jobs for background tasks
- remove supervisord which was used to run the tasks in parallel. However, this means that tasks are going to be run sequentially (which is desirable atm)

EDIT:
- Created `TaskConfig` dataclass to encapsulate logic for when to run a task based on it's last run.
- Moved the `jwt` code into the api folder and renamed the top level `token.py` to `cryptography.py` since that's what it mostly does now.

This closes #1461


.github/workflows/backend-Tests.yml
```diff
@@ -48,18 +48,22 @@ jobs:
...
           "host=localhost port=5432 dbname=zimtest user=zimfarm password=zimpass"
 
-      - name: Build backend Docker image
+      - name: Build API Docker image
         working-directory: backend
-        run: docker build -t zimfarm-backend:test .
+        run: docker build -t zimfarm-backend:test -f api.Dockerfile .
 
-      - name: Run backend container
+      - name: Build bacground tasks Docker image
```
benoit74: `bacground` => `background`

.github/workflows/backend-Tests.yml
```diff
@@ -68,14 +72,22 @@ jobs:
...
             fi
             echo "Waiting for backend..."
             sleep 3
           done
 
+      - name: Run background tasks container
+        run: |
+          docker run --rm --name zimfarm-background-tasks \
+            -e POSTGRES_URI=$POSTGRES_URI \
+            -e JWT_SECRET=DH8kSxcflUVfNRdkEiJJCn2dOOKI3qfw \
```
benoit74, line 86: Do we really need the JWT_SECRET in background tasks container? Maybe it is just too complex to make it explicit that this is mandatory only for API, but might be worth to try if feasible.

elfkuzco:
> My only concern is that with supervisord, we had a mechanism to ensure that periodic-tasks.py will run once every 10 minutes, and periodic-scheduler.py will run once every hour. I don't get how this is now working in this PR. It looks to me that we tasks will continuously run, with a BACKGROUND_TASKS_SLEEP_DURATION (1 minute) pause between them.


I suppose I could keep track of `last_run` with each task to mimic this behaviour.

.github/workflows/backend-Tests.yml
elfkuzco: Fixed now. Moved the jwt code to api folder since it's only used there. Beginning to think I can further separate the deps needed for the api and background task. Might contribute marginally to image size.

elfkuzco:
- Created `TaskConfig` dataclass to encapsulate logic for when to run a task based on it's last run.
- Moved the `jwt` code into the api folder and renamed the top level `token.py` to `cryptography.py` since that's what it mostly does now.

PR #1498, opened 2025-11-06
Upgrade monitor

### LINKED ISSUE #1102: Upgrade monitor
This is a subtask of https://github.com/openzim/zimfarm/issues/919

Monitor (used to monitor Zimfarm tasks consumption) is still using Netdata 1.38 (Feb 2023) while we are now at 2.5.1 (few days ago)

We should aim to upgrade it at some point, even if it will probably have breaking consequences given the major number change.

This also means that we must add this to the dev stack for local testing.

Not that urgent given it is not really a core tool + it is probably going to be painful.

elfkuzco:
## Rationale
Monitor (used to monitor Zimfarm tasks consumption) is still using Netdata 1.38 (Feb 2023) while we are now at 2.8.0. Also, monitor is currently not in the dev docker graph making it hard to view the statistics for local tasks. This PR aims to address these points.

## Changes
- upgrade netdata version to 2.8.0
- add `monitor` service to dev docker graph with profile `worker`
- uncomment out worker-related components and rely on docker profiles to selectively include them
- included `netdata` subfolder in `dev` directory. This is what would be hosted on https://github.com/kiwix/container-images/blob/main/netdata/netdata.conf
- use `urllib` in regen-stream because netdata base image doesn't have pip installed and we can't install requests
- netdata v2.8.0 doesn't come with wget/killall installed. So, I installed them via the package manager as those are either used in the `Dockerfile` or shell scripts
- disable all forms of metrics collection for the parent
- allow only container metrics collections with filter to send charts only for zimscraper and zimtask containers.
- change the format for cronjobs to debian style
- use `uuid5` function to build worker stream key from SHA256 fingerprint
- use the scraper name directly in development mode since we cannot retrieve it's IP address. This works because the containers all share the same network and the `go.d` plugin which communicates to the redis instance of the scraper can connect to it via it's name
Screenshot_20251127_141848

This closes #1102

### DIFF
--- a/monitor/parent/Dockerfile
+++ b/monitor/parent/Dockerfile
@@ -27,7 +27,7 @@ RUN printf "enabled: no\n" | tee /etc/netdata/python.d.conf /etc/netdata/go.d.co
     && chmod 0644 /etc/cron.d/update-stream-whitelist \
     && touch /etc/netdata/stream.conf \
     && chmod a+x /usr/local/bin/entrypoint \
-    && wget https://github.com/Yelp/dumb-init/releases/download/v1.2.5/dumb-init_1.2.5_x86_64 \
+    && wget -q https://github.com/Yelp/dumb-init/releases/download/v1.2.5/dumb-init_1.2.5_x86_64 \
     && mv dumb-init_1.2.5_x86_64 /usr/bin/dumb-init \
     && chmod +x /usr/bin/dumb-init
 
--- a/monitor/parent/regen-stream-conf.py
+++ b/monitor/parent/regen-stream-conf.py
@@ -48,7 +48,7 @@ def get_token() -> str:
     req = urllib.request.Request(url, data=data, headers=headers, method="POST")
 
     try:
-        with urllib.request.urlopen(req) as response:
+        with urllib.request.urlopen(req) as response:  # nosec B310
             data = json.loads(response.read().decode("utf-8"))
             return data.get("access_token")
     except urllib.error.HTTPError as e:
@@ -85,7 +85,7 @@ def query_api(
         req = urllib.request.Request(
             url, data=data, headers=req_headers, method=method.upper()
         )
-        with urllib.request.urlopen(req) as response:
+        with urllib.request.urlopen(req) as response:  # nosec B310
             status_code = response.getcode()
             response_text = response.read().decode("utf-8")
 


elfkuzco:
It seems the previous way of building the chart name doesn't work anymore. When I click on it, it says "Chart Not Found". So, what I did was to remove it from the URL. The image in the PR description is what will be shown on page view. Haven't figured out to construct the query param to point to a specific chart and the docs/communities haven't been helpful on it.

benoit74 (commented):
I'm not sure I get where I should have a look in the dashboard to see only stats from the "current node" (scraper). Everywhere I go I only see stats from my host machine, not only the scraper container.

There is also a bad warning (probably because I did not get how to tweak things to have the task manager start my local monitor image instead of the officially published one):

image

And finally, it looks like I have not been precise enough. It is important to also update the image at https://github.com/kiwix/container-images/tree/main/netdata since this is the one we use to run https://monitoring.openzim.org ; note that it would maybe be easier to host this image Dockerfile inside the zimfarm repo since it is so tight to the Zimfarm ; at least I don't mind if you temporarily host it there for the review to complete, and then we move it again over there.

elfkuzco:
I have updated the PR description. Still, it appears there's significant changes in the netdata operates now. I have tried disabling their cloud offering but it seems it doesn't respect my decision. Hence, I have to configure the UI to point me to `/v3` which is a UI that doesn't give the cloud pop-up netdata/netdata#15640

elfkuzco:
Also had to rebase the branch on top of main because of #1521

benoit74:
I still don't get what I'm seeing in the graphs. I expect to see resource usage (mostly CPU and memory) of the scraper container itself. Here it looks like I see resource usage of my whole machine, not only the scraper. Do I miss something?

elfkuzco:
> I still don't get what I'm seeing in the graphs. I expect to see resource usage (mostly CPU and memory) of the scraper container itself. Here it looks like I see resource usage of my whole machine, not only the scraper. Do I miss something?

Could you show a picture?

benoit74:
Sure. For instance, I'm quite sure the scraper I've started did not consumed 16G used and 42G cache memory.

image

elfkuzco:
Applied some updates. See updated README. Notable changes since last PRs are:
- upgraded to 2.8.0
- only show container level metrics. Sending only `zimscraper*` and `zimtask*` metrics. https://github.com/openzim/zimfarm/blob/5822e5ae03159cb4f1d714b581f8f3af5650caf3/monitor/entrypoint.sh#L6-L10

The system-level collection collects entire metrics for my host machine making it hard to discern worker-related metrics. Should we use this as an opportunity to make all containers started within the manager start with a known prefix and collect only their stats. Collecting stats of the zimfarm backend and postgresdb (because I am in development) is quite unnecessary and reducing the amount of data collected could speed up streaming speed. What do you think?

elfkuzco:
You will of course need to build the monitor image with the same name set as `MONITOR_IMAGE` in the compose file

benoit74:
I'm now a bit confused. How did it worked in the past? Did we collected metrics from all containers on the worker host, no matter which task they belong to? And also host metrics?

elfkuzco:
> Did we collected metrics from all containers on the worker host, no matter which task they belong to? And also host metrics?

Yes

elfkuzco:
Previously, netdata listed each container/cgroup in its own section making it easy to visualize. But that is not the case in the v2 where everything is aggregated under the **Containers & VMs** section. Thus, it is up to the viewer to selectively group the stats by instance and apply filters by themselves to view per-instance metric. The first image shows CPU usage but with Grouping done by cgroup instance and only `zimscraper_mwofflienr....` selected (see the first row in the image to see the filters).  But the second image shows the stats for all the containers running on my machine.


Also, in the older UI, query parameters were constructed to take the user to the specific cgroup chart https://github.com/openzim/zimfarm/blob/95b232b793817100f4d9e73e06c28aecd1650a5b/frontend-ui/src/views/TaskDetailView.vue#L654-L664

However, this is not possible in the UI, so, you will have to scroll to the **Containers & VMs** section.

Screenshot_20251202_121507
Screenshot_20251202_121439

elfkuzco:
Here are some useful links with people clamoring for these older features that made life simpler but it appears the netdata team is heading in a new direction

https://community.netdata.cloud/t/6-years-experience-but-can-not-use-netdata/4994
https://github.com/netdata/netdata-cloud/issues/654
https://learn.netdata.cloud/docs/dashboards-and-charts/charts
https://github.com/netdata/netdata-cloud/issues/894

benoit74 (commented):
Thank you! Are we really collecting parent host or containers data in production at https://monitoring.openzim.org? I don't see where it would be configured, and I'm really not sure it is useful (there are many services running on the same machine, and I don't see them in current charts).

Please also include some details in a README.md about how to select a given container to view its CPU / memory consumption.

elfkuzco:
> Are we really collecting parent host or containers data in production at https://monitoring.openzim.org?


Nope. The parent doesn't collect statistics partly because everything is disabled https://github.com/kiwix/container-images/blob/f48ea9cec98b744ca0f2067acf9f225faf49ea77/netdata/netdata.conf#L15-L28

I extended these to ignore more since there are even more plugins

It's the child which doesn't ignore these plugins that collects everything from where it is running.

elfkuzco:
I have added jobs to build parent and child images from zimfarm. However, in contrast to kiwix/containier-images that published with image name `kiwix/netdata`, I published this as `openzim/zimfarm-netdata`. This would probably require upstream changes in how the service is deployed.

https://github.com/kiwix/container-images/blob/f48ea9cec98b744ca0f2067acf9f225faf49ea77/.github/workflows/netdata_ci.yml#L13-L28

.github/workflows/Publish.yml
```diff
@@ -291,7 +293,27 @@ jobs:
           image-name: openzim/zimfarm-monitor
```
benoit74: Let's rename this to `zimfarm-monitor-child` for clarity

.github/workflows/Publish.yml
```diff
@@ -291,7 +293,27 @@ jobs:
...
+    needs: paths-filter
+    if: ${{ needs.paths-filter.outputs.monitor == 'true' || inputs.force || contains(inputs.force_components, 'monitor') }}
+    steps:
+      - name: Retrieve source code
+        uses: actions/checkout@v5
+
+      - name: Build and publish Docker Image
+        uses: openzim/docker-publish-action@v10
+        with:
+          image-name: openzim/zimfarm-netdata
```
benoit74: Let's rename this to `zimfarm-monitor-parent` for clarity

dev/docker-compose.yml
```diff
@@ -33,7 +33,7 @@ services:
       JWT_SECRET: DH8kSxcflUVfNRdkEiJJCn2dOOKI3qfw
       POSTGRES_URI: postgresql+psycopg://zimfarm:zimpass@postgresdb:5432/zimfarm
       ALEMBIC_UPGRADE_HEAD_ON_START: "true"
-      # upload logs to receiver but logs, zimcheck results and artificats to s3
+      # upload zims to receiver but logs, zimcheck results and artificats to s3
```
benoit74: `artificats` => `artifacts`

dev/docker-compose.yml
```diff
@@ -142,6 +142,49 @@ services:
...
+      - /etc/os-release:/host/etc/os-release:ro
+      - /var/log:/host/var/log:ro
+      - /var/run/docker.sock:/var/run/docker.sock:ro
+    depends_on:
+      backend:
+        condition: service_healthy
+    profiles:
+      - worker
+
+  task_monitor:
```
benoit74: Add comment explaining this is just meant to build the child monitor image locally so that it is available to be used by the task manager

benoit74: + rename container, image and container name to `monitor_child` and `zf_monitor_child` for clarity

dev/docker-compose.yml
```diff
@@ -142,6 +142,49 @@ services:
     profiles:
       - worker
 
+  monitor:
```
benoit74: rename to `monitor_parent` and `zf_monitor_parent` for clarity

dev/docker-compose.yml
```diff
@@ -142,6 +142,49 @@ services:
...
+    environment:
+      - ZIMFARM_API_URL=http://backend:80/v2
+      - ZIMFARM_USERNAME=admin
+      - ZIMFARM_PASSWORD=admin
+      - NETDATA_HOSTNAME=monitor
+    volumes:
+      - netdataconfig:/etc/netdata
+      - netdatalib:/var/lib/netdata
+      - netdatacache:/var/cache/netdata
+      - /etc/passwd:/host/etc/passwd:ro
```
benoit74: Why do we need to mount all these host folders since we do not monitor anything on the host? This is not done in production at least, so I'm not sure it really helps to reproduce production setup, it at least deserve a comment

monitor/parent/README.md
```diff
@@ -0,0 +1,65 @@
+# maintenance-netdata
```
benoit74: why `maintenance-` ?

worker/src/zimfarm_worker/common/utils.py
```diff
@@ -17,8 +19,5 @@ def format_size(value: float) -> str:
 
 
 def format_key(fingerprint: str) -> str:
```
benoit74, line 21: Do you confirm this is used only for monitoring? No impact on running tasks (assuming if necessary they don't use monitoring, which is the case anyway)?

benoit74 (changes requested):
Few remarks, we are getting close to being able to merge this. I was aware of image renaming consequences, but thank you for mentioning it.

I just noticed initial comment also deserves a bit of rewriting to match latest changes.

worker/src/zimfarm_worker/common/utils.py
elfkuzco: Yes

dev/docker-compose.yml
elfkuzco: This is what the netdata docker documentation used in the docker setup guide. I think the first three are just mounts of the configuration files and cache. Don't really know why they required much of the host files but I guess if production works without them, I can safely remove them. Will try to re-test without them and see if things work as expected.

elfkuzco: seems to work without those mounts

elfkuzco:
created upstream PR on kiwix/container-images#284 to remove the netdata folder there

PR #1503, opened 2025-11-07
publish specific components by names

elfkuzco:
## Changes
- change workflow dispatch to publish specific component to be a string instead of maintaining seperate `force-<service` input for each service. Also, this avoids the error associated with a workflow input being greater than 10.


benoit74 (approved):
LGTM

PR #1504, opened 2025-11-07
add task to send notification for checked files to CMS

### LINKED ISSUE #1436: CMS - enhance reliability of CMS notifications
Currently, CMS is informed about ZIM file creation when the ZIM has been checked (with zimcheck) and status sent to Zimfarm.

The problem with this situation is that if for any reason the CMS is not available at that moment, then notification is lost.

We also miss the capability to resend some or all notifications should it be needed. This might be needed for maintenance reason e.g. if CMS has lost some data. This is useful to be resend because it avoids to have to rerun zimcheck and extract ZIM metadata.

We need a better system capable to retry to send a given notification on a regular basis (probably with a regular background task) if previous attempts have failed and to mark some notifications as needing to be sent again.

On CMS side, the notification API will be idempotent (i.e. it is possible to notify multiple times the same file without any impact on the CMS), so we should not be worried about that.

I feel like the task files probably deserve their own table, so that we can have a `cms_notified` column on each of them.

elfkuzco:
## Rationale
This PR implements a retry mechanism for CMS notifications using a new background task that runs periodically.

## Changes
- create a seperate `File` table which is going to be populated with entries from `task.files` (a JSONB column)
- add a `cms_notified` flag on the new table to indicate if CMS has been successully notified about a file
- add task to retry notifying cms about checked files

This closes #1436


backend/src/zimfarm_backend/background_tasks/entrypoint.py
```diff
@@ -68,6 +78,9 @@ def main():
     logger.info(f"Configured {len(tasks)} tasks:")
 
     while True:
+        # First, upgrade the database schema
+        if ALEMBIC_UPGRADE_HEAD_ON_START:
+            upgrade_db_schema()
```
benoit74: This DB upgrade should be outside (before) the while loop, otherwise we will continuously run upgrades (which would be mostly a noop, but no need to put such strain on the DB).

backend/src/zimfarm_backend/background_tasks/retry_cms_notifications.py
```diff
@@ -0,0 +1,65 @@
...
+from sqlalchemy.orm import Session as OrmSession
+
+from zimfarm_backend.background_tasks import logger
+from zimfarm_backend.common.constants import INFORM_CMS
+from zimfarm_backend.common.external import advertise_book_to_cms
+from zimfarm_backend.db.models import File, Task
+from zimfarm_backend.db.tasks import get_task_by_id
+
+
+def retry_cms_notifications(session: OrmSession):
```
benoit74: naming this `retry_cms_notification` is kinda ... pessimistic ^^

In general this code responsibility is to send notifications, and we can hope it will succeed on first attempt ^^

I would rename the function + fix docstring to be more optimistic (rename filename as well)

backend/src/zimfarm_backend/background_tasks/retry_cms_notifications.py
```diff
@@ -0,0 +1,65 @@
...
+        return
+
+    logger.info(f"::: found {nb_files} file(s) needing CMS notification")
+
+    nb_notified = 0
+    nb_failed = 0
+
+    for row in files_to_notify:
+        file = cast(File, row.File)
+        try:
```
benoit74: We should stop on first failed notification, there is no reason for one notification to fail and next ones to succeed. At least this is the target design, better to stick to it and avoid overwhelming CMS if there is a serious bug.

backend/src/zimfarm_backend/common/external.py
```diff
@@ -177,21 +177,19 @@ def advertise_books_to_cms(task_id: UUID, session: so.Session):
     Safe to re-run as successful requests are skipped"""
```
benoit74: This `advertise_books_to_cms` method in now unused, better to remove it

backend/src/zimfarm_backend/common/external.py
```diff
@@ -177,21 +177,19 @@ def advertise_books_to_cms(task_id: UUID, session: so.Session):
...
 
     file_data = task.files[file_name]
 
     # skip if already advertised to CMS
-    if file_data.cms:
-        # cms query already succeeded
-        if file_data.cms.succeeded:
-            return
+    if file_data.cms and file_data.cms.succeeded:
+        return
```
benoit74, line 186: we should log a warning to not be surprised

backend/src/zimfarm_backend/common/utils.py
```diff
@@ -157,24 +159,28 @@ def add_to_debug_if_present(task: Task, kwargs_key: str, debug_key: str) -> None
     if kwargs.get("file", {}).get("name"):
         fkey = kwargs["file"]["name"]
         fstatus = kwargs["file"].get("status")
+
+        values: dict[str, Any] = {
```
benoit74: I don't get how this isy preserving keys which have already been set.

E.g. first call will be with fstatus "created", it will set the "size" value.

Second call will be with fstatus "uploaded". The call to create_or_update_task_file will not have the "size" in values, so it's value will be None inside this function. The insert statement will fail (conflict detected), update will be ran on the DB and will set the "size" to None.

Unless I miss something.

backend/src/zimfarm_backend/common/utils.py
```diff
@@ -397,7 +403,7 @@ def task_checked_file_event_handler(
         if task is None:
             raise RecordDoesNotExistError(f"Task {task_id} does not exist")
         if filename := file.get("name"):
-            advertise_book_to_cms(task, filename)
+            advertise_book_to_cms(session, task, filename)
```
benoit74: We need to remove this, we do not want to advertise book to the CMS from the API anymore. 

But what it makes me realize is that we miss an information in DB (new column probably) about the fact that file is ready to be notified to the CMS. We do not want to until this checked event is received. We could reuse the fact that `check_result` is populated ... but it makes it harder to use in query of the background task.

backend/src/zimfarm_backend/common/external.py
```diff
@@ -230,6 +228,18 @@ def advertise_book_to_cms(task: TaskFullSchema, file_name: str):
             )
 
     # record request result
+    create_or_update_task_file(
+        session,
+        task_id=task.id,
+        name=file_name,
+        status=file_data.status,
```
benoit74: Why setting again the status, I don't see what would have modified it.

benoit74: And then I don't get why we set the file_data.cms property to then pass it to create_or_update_task_file and then return it.

I would rather:
- build a local `CMSStatusSchema` object 
- pass it to create_or_update_task_file
- retrieve task_file again from DB (to be sure to have the "real" value)

benoit74: And errors when notifying CMS should be fatal to the background task (we should not try to notify other books), CMS is meant to accept all notifications, if one notification fails we want to stop sending notifications

backend/src/zimfarm_backend/db/models.py
```diff
@@ -190,6 +189,48 @@ class Task(Base):
...
+    __tablename__ = "file"
+    id: Mapped[UUID] = mapped_column(
+        init=False, primary_key=True, server_default=text("uuid_generate_v4()")
+    )
+    name: Mapped[str]
+    status: Mapped[str]
+    size: Mapped[int | None] = mapped_column(default=None, type_=BigInteger)
+
+    # CMS status fields (nested in TaskFileSchema.cms)
+    cms_status_code: Mapped[int | None] = mapped_column(default=None)
```
benoit74: You should drop:
- `cms_status_code`: no need to keep failure status code in DB, it should not happen, will never be analyzed from UI, ops will go to the logs
- `cms_succeeded`: it is confusing and a duplication of `cms_notified` ; let's keep only `cms_notified` for the time being 
- `cms_book_id` and `cms_title_ident`: these information will not be supplied by the CMS anymore (this is a new info I've never shared with you)

In the migration upgrade operation, you should keep `cms_notified` and `cms_on` to None, we will restart CMS with a blank DB soon

benoit74: Consequence is that downgrade will not be capable to repopulate the "cms_book_id", "cms_title_ident" and "cms_status_code", but this is only garbage ATM anyway so I'm fine with it.

benoit74 (approved):
See inline comments + add `INFORM_CMS=0` in dev docker-compose on the background task (there is no CMS in this graph)

Makes me realize there are really many things moving ATM ^^

Note that I've also set `INFORM_CMS=0` in production (on API and on background tasks).

I did not made any real test for now since there is already too much to discuss / fix

backend/src/zimfarm_backend/common/utils.py
elfkuzco: Thanks for spotting this. Resolved.

backend/src/zimfarm_backend/common/utils.py
elfkuzco: I don't see how it is difficult to use in the query. It's an `int` and it would just be an additional where clause to determine if it's `1` or `0` with 0 being success. Is that not what you mean?

benoit74: Regarding my second point in original comment, `File.status == "checked" condition is just perfect in fact.

But still we do not want to advertise book to CMS from the API, and unless I'm mistaken this is what this method is doing right?

backend/src/zimfarm_backend/db/tasks.py
```diff
@@ -41,28 +45,33 @@ class TaskListResult(BaseModel):
...
+        status=file.status,
+        created_timestamp=file.created_timestamp,
+        uploaded_timestamp=file.uploaded_timestamp,
+        failed_timestamp=file.failed_timestamp,
+        check_timestamp=file.check_timestamp,
+        check_result=file.check_result,
+        check_log=file.check_log,
+        check_details=file.check_details,
+        info=file.info,
+        cms=CMSStatusSchema(
```
benoit74: Let's unnest these properties, it is not like they are really used by any of our external systems ATM, only our own UI, so better to do it now than regret it later 🤣 (and you can then change comment in backend/src/zimfarm_backend/db/models.py)

benoit74 (changes requested):
Few last remarks, getting closer, that's a nice move, a like it.

backend/src/zimfarm_backend/common/utils.py
elfkuzco: Oh, yes, you are correct. Removing...

backend/src/zimfarm_backend/background_tasks/send_cms_notifications.py
```diff
@@ -0,0 +1,55 @@
...
+        for row in files_to_notify:
+            file = cast(File, row.File)
+            task_full = get_task_by_id(session, row.task_id)
+
+            advertise_book_to_cms(session, task_full, file.name)
+            nb_notified += 1
+
+            logger.debug(f"Notified CMS for file {file.name} from task {row.task_id}")
+    except Exception:
+        logger.exception("Failed to send CMS notification")
```
benoit74: Log number of notification already sent before failure

backend/src/zimfarm_backend/background_tasks/entrypoint.py
```diff
@@ -67,6 +77,8 @@ def main():
     logger.info("Starting background tasks...")
     logger.info(f"Configured {len(tasks)} tasks:")
 
+    if ALEMBIC_UPGRADE_HEAD_ON_START:
+        upgrade_db_schema()
```
benoit74, line 85: Also add  `check_if_schema_is_up_to_date()` and `create_initial_user()` calls (unconditional, no matter the `ALEMBIC_UPGRADE_HEAD_ON_START` value), this might be handy for situations where we want to only start the background_tasks for whatever reason.

backend/src/zimfarm_backend/background_tasks/entrypoint.py
```diff
@@ -67,6 +77,8 @@ def main():
     logger.info("Starting background tasks...")
     logger.info(f"Configured {len(tasks)} tasks:")
```
benoit74: Remove the trailing `:`

benoit74 (approved):
LGTM ; please implement minor modifications and we are good to merge

Please merge only tomorrow morning so that if something goes wrong in the migration we have a bit of time ahead of us to fix.

PR #1507, opened 2025-11-10
use request timeout for wasabi requests

### LINKED ISSUE #863: API: No timeout for Wasabi operations
Operations performed on Wasabi (typically updating the S3 policy when a worker IP changes) have no timeout.

This means that DB locks arise when Wasabi is not responding because all subsequent worker requests are trying to do the same thing which is blocked

We should:
- implement a timeout for Wasabi operations (maybe just monitor the operation as a whole) with a nice log message
- add a Grafana alert for these timeouts
- probably commit the DB changes before calling Wasabi, so that we do not mind if this works or not, or at least add a try/catch around Wasabi operations

On the long term, just like CMS operations, this should be moved to an asynchronous job

elfkuzco:
## Rationale
This PR adds request timeouts to wasabi requests which allows us to control the duration of requests made to Wasabi without waiting unnecessarily.

## Changes
- configure `KiwixStorage` to use timeouts before making requests to Wasabi

This fixes #863


backend/src/zimfarm_backend/common/constants.py
```diff
@@ -116,6 +116,7 @@ def getenv(key: str, *, mandatory: bool = False, default: Any = None) -> Any:
 WASABI_WHITELIST_STATEMENT_ID = getenv(
     "WASABI_WHITELIST_STATEMENT_ID", default="ZimfarmWorkersIPsWhiteList"
 )
+WASABI_REQUEST_TIMEOUT = parse_timespan(getenv("WASABI_REQUEST_TIMEOUT", default="1m"))
```
benoit74: I suggest we set a more aggressive default value like 10s, this is already a lot and should be sufficient (or will be customized by env var if needed).

PR #1511, opened 2025-11-11
Zimcheck results to s3

### LINKED ISSUE #1456: Some `zimcheck` results are too big
See https://github.com/openzim/zimfarm/issues/1307#issuecomment-3428989229

I ran `zimcheck` on this `mathoverflow.net_en_all_2025-10.zim` ZIM and it produced a **103M** zimcheck output (see also https://github.com/openzim/sotoki/issues/348 for underlying reason).

This proves we have following issues:
- task manager should not try to send these results in an API call
- zimfarm API should not send these results to CMS in the notification
- such big data should not be stored inside the database

I propose we change the architecture to:
- store `zimcheck` results on a publicly available S3 bucket
- send S3 url ~~+ summarized statistics (`status`, number of errors, warnings, ...)~~ to zimcheck results in API calls (from task manager to Zimfarm and from Zimfarm to CMS)
- store S3 url ~~+ summarized statistics~~ in DB
- modify UI to retrieve zimcheck results from S3 directly (probably needs a warning about size of download)
- ~~cleanup S3 bucket appropriately (probably needs a daily background job doing an inventory of files on S3 bucket + an inventory of existing URLs in DB and deleting everything which is not necessary anymore)~~

This means we will need a maintenance script to fix existing situation (compute statistics, upload DB zimcheck results to S3 and store S3 URL in place).

@rgaudin do you have some insight to provide here? Did I misunderstood or oversight something?

elfkuzco:
## Rationale
This PR removes the direct upload of the zimcheck results to the API but instead sends the data to an S3 bucket or external service that can be used with uploader. This way, the bottlenec of having to upload such large files to the API is avoided as it only keeps track of the zimcheck file to be reconstructed at any given time.

## Changes
- add zimcheck upload configuration variables, `ZIMCHECK_UPLOAD_URI` and `ZIMCHECK_EXPIRATION`
- add new event `check_results_uploaded` to detect when check results have been uploaded. This allows the checker and upoad to continue as different events. As a consequence of this event being distinct, we make the `status` field nullable allowing us to insert/update because we don't need to set a status everytime
- reconstruct zimcheck upload url before advertising to CMS
- add fields `check_filename` and `check_upload_timestamp` to `File` in the DB
- check for `check_result` in task to send cms notifications instead of `status` as that can easily be overwritten
- fetch check results from frontend and show to user when they want
- start different uploader after a check is succesful with its own list of files.

This closes #1456


backend/src/zimfarm_backend/common/external.py
```diff
@@ -186,13 +186,17 @@ def advertise_book_to_cms(session: so.Session, task: TaskFullSchema, file_name:
         return
 
     # prepare payload and submit request to CMS
-    download_prefix = f"{CMS_ZIM_DOWNLOAD_URL}{task.config.warehouse_path}"
+    download_prefix = f"{CMS_ZIM_DOWNLOAD_URL}/{task.config.warehouse_path}"
```
benoit74: Why? warehouse_path is supposed to start with a `/`

backend/src/zimfarm_backend/common/utils.py
```diff
@@ -159,21 +160,25 @@ def add_to_debug_if_present(task: Task, kwargs_key: str, debug_key: str) -> None
...
             setattr(values, f"{fstatus}_timestamp", timestamp)
         elif fstatus == "checked":
+            values.status = fstatus
             values.check_result = kwargs["file"].get("check_result")
-            values.check_log = kwargs["file"].get("check_log")
             values.check_timestamp = timestamp
-            values.check_details = kwargs["file"].get("check_details", {})
             values.info = kwargs["file"].get("info", {})
+        elif fstatus == "check_results_uploaded":
+            # Don't update file status here as it must have been created/uploaded
```
benoit74: I don't get the rationale for not saving file status, sorry, could you develop?

backend/src/zimfarm_backend/db/models.py
```diff
@@ -200,7 +200,7 @@ class File(Base):
         init=False, primary_key=True, server_default=text("uuid_generate_v4()")
     )
     name: Mapped[str]
-    status: Mapped[str]
+    status: Mapped[str | None] = mapped_column(default=None)
```
benoit74: This seems risky, I don't like we could have a file with status set to `None`

backend/src/zimfarm_backend/db/models.py
```diff
@@ -210,12 +210,14 @@ class File(Base):
...
     uploaded_timestamp: Mapped[datetime | None] = mapped_column(default=None)
     failed_timestamp: Mapped[datetime | None] = mapped_column(default=None)
-    check_timestamp: Mapped[datetime | None] = mapped_column(default=None)
-
     # Check fields
+    check_timestamp: Mapped[datetime | None] = mapped_column(default=None)
     check_result: Mapped[int | None] = mapped_column(default=None)
     check_log: Mapped[str | None] = mapped_column(default=None)
     check_details: Mapped[dict[str, Any] | None] = mapped_column(default=None)
+    # if filename exists, then check result was uploaded successfully, otherwise false
```
benoit74: I don't get what `otherwise false` means. Either it is important and we need to be more precise or this is an irrelevant comment.

worker/src/zimfarm_worker/task/worker.py
```diff
@@ -450,6 +473,100 @@ def start_checker(self, filename: str):
...
+                    del self.zimcheck_files[zimcheck_filename]
+                    if self.task_workdir:
+                        (self.task_workdir / zimcheck_filename).unlink(missing_ok=True)
+
+            self.zimcheck_uploader.remove()
+            self.zimcheck_uploader = None
+
+        # Start an uploader instance
+        if (
+            self.task
```
benoit74: why are all these checks so different than for other files containers (upload and zimcheck)?

worker/src/zimfarm_worker/task/worker.py
```diff
@@ -727,24 +844,45 @@ def check_files(self):
...
+            )
+
             try:
                 zimcheck_result = ujson.loads(zimcheck_log)
             except Exception as exc:
+                # Failed to parse JSON, send raw log
                 zimcheck_result = None
                 logger.warning(f"Failed to parse zimcheck output: {exc}")
-            else:
-                zimcheck_log = None
```
benoit74, line 737: Why not setting zimcheck_log to None anymore? Doesn't it means we will upload both raw and parsed logs?

worker/src/zimfarm_worker/task/worker.py
```diff
@@ -727,24 +844,45 @@ def check_files(self):
                 self.docker,
                 self.checker.name,  # pyright: ignore[reportArgumentType]
             ).strip()
+
+            zimcheck_info = get_zim_info(
```
benoit74: This is not info about zimcheck, but info about the zim, so better named `zim_info`

worker/src/zimfarm_worker/task/worker.py
```diff
@@ -727,24 +844,45 @@ def check_files(self):
...
+                info=zimcheck_info,
             )
+            # zimcheck results could be too big, so, we write them to a file
+            # https://github.com/openzim/zimfarm/issues/1456
+            zimcheck_filename = f"{zimcheck_info['id']}_zimcheck.json"
+            try:
+                with open(self.task_workdir / zimcheck_filename, "w") as f:
+                    json.dump(zimcheck_metadata, f)
+            except Exception as exc:
+                logger.error(f"Failed to write zimcheck metadata: {exc}")
```
benoit74: zimcheck metadata is confusing, it looks like metadata about the zimcheck itself. I would suggest to use "zimcheck file content" / `zimcheck_file_content` or something like this

backend/src/zimfarm_backend/common/external.py
elfkuzco: Oh, forgot...

backend/src/zimfarm_backend/common/utils.py
elfkuzco: Okay. I thought perhaps it wouldn't be important and leave it as `created`, `failed`, `uploaded`, `checked`.

backend/src/zimfarm_backend/db/models.py
elfkuzco: I mean, if there is no filename, then, it means the check result was never uploaded. Will reword comment...

worker/src/zimfarm_worker/task/worker.py
elfkuzco: The `start_uploader` function just blinldy indexes into task to find the upload uri. For the zim and logs, defaults are always provided at the application level but not so for the artifacts url. That's why the artifacts uploader checks if an upload uri was provided. I didn't want to set a default for the zimchecks but I suppose I should have and avoid hte check to see if an upload URI was set. Will set the default to be `sftp://uploader@warehouse.farm.openzim.org:1522/zimchecks` so, it is always provided.

backend/src/zimfarm_backend/background_tasks/send_cms_notifications.py
```diff
@@ -23,7 +23,7 @@ def notify_cms_for_checked_files(session: OrmSession):
         sa.select(File, Task.id.label("task_id"))
         .join(Task, Task.id == File.task_id)
         .where(
-            File.status == "checked",
+            File.check_result.is_not(None),
```
benoit74: I realize this is not exactly correct. On one end we need to wait for zimcheck results to be uploaded to S3, but on another end we need to cope with the fact that zimcheck might never run or notification being lost (bad bug, worker interrupted, ...).

Since the logic is that we want notify CMS of any ZIM file produced, we need to give up waiting for zimcheck after some delay. CMS is notified that a file has been created but is missing check_result and/or check_filename, and then CMS can decide what needs to be done.

I propose that we amend a bit this logic:
- do not filter for file status at all in the SQL statement
- filter "afterwards" in Python code

The Python filtering should remove any file which:
- has not a `check_result`or has not a`check_filename`
- and is not older than a given delay (24h by default)

frontend-ui/src/utils/offliner.ts
```diff
@@ -272,7 +272,7 @@ export function artifactsUrl(task: Task) {
   return uploadUrl(task.upload.artifacts.upload_uri, task.container.artifacts)
 }
 
-function uploadUrl(uri: string, filename: string) {
```
benoit74, line 275: For consistency, prefer to not export this function and move getCheckUrl function to this file.

worker/src/zimfarm_worker/task/worker.py
```diff
@@ -104,10 +105,21 @@ def __init__(
...
+        )  # ZIM files registry
+        self.zim_upload_retries: defaultdict[str, int] = defaultdict(
+            int
+        )  # ZIM files with upload errors (registry)
+        self.zim_uploader: Container | None = None  # zim-files uploader container
+        self.checker: Container | None = None  # zim-files checker container
+
+        self.zimcheck_uploader: Container | None = None  # zimcheck result uploader
+        # mapping of zimcheck result filenames to zim filename. This allows multiple
+        # checkers to run independent of the rate that upload for the check results
```
benoit74: I don't get this. We still have only one `self.checker` and one `self.zimcheck_uploader`, so we will not have multiple checkers runing at the same time

worker/src/zimfarm_worker/task/worker.py
```diff
@@ -104,10 +105,21 @@ def __init__(
...
+        # happen
+        self.zimcheck_files: dict[str, str] = {}
+        self.zimcheck_retries: defaultdict[str, int] = defaultdict(int)
```
benoit74: For consistency, group `self.zimcheck_retries` with `self.zim_upload_retries`

worker/src/zimfarm_worker/task/worker.py
```diff
@@ -450,6 +477,101 @@ def start_checker(self, filename: str):
             filename=filename,
         )
 
+    def start_zimcheck_uploader(self, filename: str):
+        if not self.task or not self.host_task_workdir:
+            logger.error("No task or host task workdir to start uploader")
+            return
+
+        logger.info(f"Starting zimcheck uploader for {filename}")
```
benoit74: log this at the very beginning of the method for consistency + easier debuging

worker/src/zimfarm_worker/task/worker.py
```diff
@@ -424,7 +451,7 @@ def start_uploader(self, upload_dir: Path, filename: str):
         if not self.task or not self.host_task_workdir:
             logger.error("No task or host task workdir to start uploader")
```
benoit74: `to start uploader` => `to start ZIM uploader` (and fix method name to `start_zim_uploader`, and move the stupid `Gathering ZIM metadata ...` log (WTF?) to proper place (start of `get_zim_info` method), and fix other log to mention `ZIM uploader`

worker/src/zimfarm_worker/task/worker.py
benoit74: My point was more:
- why do we check for `self.task` (this is not needed elsewhere)
- why do we check for `self.host_task_workdir` (this is not needed elsewhere)
- why don't we check for `self.pending_zim_files(CHK_UPLOAD)` (or maybe even better `len(self.zimcheck_files) > 0`)

worker/src/zimfarm_worker/task/worker.py
benoit74: Please fix log messages here and below as well

worker/src/zimfarm_worker/task/worker.py
```diff
@@ -720,48 +852,72 @@ def check_files(self):
...
+                zim_file, _ = self.pending_zim_files(ZIM_CHECK).pop()
             except Exception:
                 # no more pending files,
                 logger.debug("failed to get ZIM file: pending_zim_files empty")
             else:
                 self.start_checker(zim_file)
-                self.zim_files[zim_file][CHK] = UPLOADING
+                self.zim_files_actions_status[zim_file][ZIM_CHECK] = DOING
 
     def upload_files(self):
```
benoit74: Fix method name

worker/src/zimfarm_worker/task/worker.py
```diff
@@ -772,74 +928,77 @@ def upload_files(self):
...
                 # no more pending files,
                 logger.debug("failed to get ZIM file: pending_zim_files empty")
             else:
                 self.start_uploader(self.task["config"]["warehouse_path"], zim_file)
-                self.zim_files[zim_file][UP] = UPLOADING
+                self.zim_files_actions_status[zim_file][ZIM_UPLOAD] = DOING
 
     def handle_files(self):
         self.upload_files()
         self.check_files()
```
benoit74: also rename `check_files` to `check_zims`, and `handle_files` to `handle_zims`

worker/src/zimfarm_worker/task/worker.py
elfkuzco: i didn't want to start the zimcheck_uploader from the checker. By adding the filename to the dictionary, it can continue doing any other checks. Comment is misleading now that I think about it but the ideas was for checks to be running independent of if the files were uploaded or not because starting the zimcheck_uploader after each check would block.

backend/src/zimfarm_backend/background_tasks/send_cms_notifications.py
elfkuzco: Any reason we shouldn't the filter in SQL? I mean, if it's going to ultimately removed, then, removing at SQL is better, unless we want to log or do something?

worker/src/zimfarm_worker/task/worker.py
benoit74: I'm totally fine with the variable name, intent, usage... global worker behavior seems ok to me.

My only concern is the comment which is very misleading. AFAIK the map contains only the check files which have completed, and does not allow to run multiple checkers or uploaders at once.

backend/src/zimfarm_backend/background_tasks/send_cms_notifications.py
benoit74: Indeed, let's do this in SQL.

worker/src/zimfarm_worker/task/worker.py
benoit74: (just in terms of position in the codebase)

backend/src/zimfarm_backend/background_tasks/send_cms_notifications.py
elfkuzco: Check updated SQL query and see if it's correct?

benoit74: I'm not that familiar with the "epoch" function so I cannot guarantee this part does the right job, but the rest LGTM

elfkuzco:
- made script executable
- added flag `type=int` to make argparse parse the `--expiration` as int
- reused existing `rebuild_uri` which has type annotations

elfkuzco:
Do you think we should use a minio bucket just like on WP1? It's easier to develop with than the receiver being run in an uploader. I can view the files in my bucket from the browser and for cases like this where there is an update script that works with S3 urls only, it would be seamless

benoit74:
I'm fine with this idea of using Minio for cases where things are pushed to S3 in production (logs, artifacts, zimcheck results). I feel like pushing the ZIMs to the receiver should be the default to represent more exactly the main prod usage + the diversity of upload methods we might have. But it should be just a matter of an env change to upload ZIMs to S3 instead.

elfkuzco:
Added minio to zimfram docker graph. Configured docker compose to upload logs, artifacts and zimcheck results to their respective buckets with the zims being sent to the reciever

backend/src/zimfarm_backend/background_tasks/constants.py
```diff
@@ -66,3 +66,6 @@
 CMS_NOTIFICATIONS_INTERVAL = datetime.timedelta(
     seconds=parse_timespan(getenv("CMS_NOTIFICATIONS_INTERVAL", default="10m"))
 )
+ZIMCHECK_RESULT_WAIT_INTERVAL = parse_timespan(
```
benoit74: `CMS_MAXIMUM_RETRY_INTERVAL` would better convey the intent (retry to send notification to CMS only for a given amount of time after file is ready to send)

backend/src/zimfarm_backend/background_tasks/send_cms_notifications.py
```diff
@@ -23,11 +25,18 @@ def notify_cms_for_checked_files(session: OrmSession):
...
         .where(
-            File.status == "checked",
-            sa.or_(
-                File.cms_notified.is_(None),
-                File.cms_notified.is_(False),
-            ),
+            # We should send notifications for files that meet the following criteria:
+            # - have not been successfully notified
+            #  - have  check_result or check_filename
+            #  - are older than ZIMCHECK_RESULT_WAIT_INTERVAL since check_timestamp
```
benoit74: - `are older` => `are not older`
- explain why we don't want files older than `ZIMCHECK_RESULT_WAIT_INTERVAL` (not retrying forever to send notifications which cannot be sent to the CMS due to another issue)
- fix extra spaces in comment

backend/src/zimfarm_backend/common/constants.py
```diff
@@ -67,6 +67,15 @@ def getenv(key: str, *, mandatory: bool = False, default: Any = None) -> Any:
 # on the bucket and the bucket retention is smaller than value below
 ARTIFACTS_EXPIRATION = int(getenv("ARTIFACTS_EXPIRATION", default="30"))
 
+ZIMCHECK_UPLOAD_URI = getenv(
```
benoit74: `ZIMCHECK_UPLOAD_URI` => `ZIMCHECK_RESULTS_UPLOAD_URI`

backend/src/zimfarm_backend/common/constants.py
```diff
@@ -67,6 +67,15 @@ def getenv(key: str, *, mandatory: bool = False, default: Any = None) -> Any:
...
+    "ZIMCHECK_UPLOAD_URI",
+    default="sftp://uploader@warehouse.farm.openzim.org:1522/zimchecks",
+)
+# zimcheck files expiration, 0 to disable expiration
+# only working in Wasabi S3 for now, works only if retention compliance is activated
+# on the bucket and the bucket retention is smaller than value below
+ZIMCHECK_EXPIRATION = int(getenv("ZIMCHECK_EXPIRATION", default="0"))
```
benoit74: `ZIMCHECK_EXPIRATION` => `ZIMCHECK_RESULTS_EXPIRATION`

PR #1518, opened 2025-11-13
kiwix oauth2 with ory.sh

### LINKED ISSUE #1509: Add support for Oauth2 authentication with ory.sh
This is a sub-issue of https://github.com/openzim/zimfarm/issues/1356

To prepare the transition to ory.sh authentication, we need to first add support for it.

Users will still be able to login through username / password, but also through ory.sh when configured appropriately.

Authentication will be implemented mostly in the frontend part, like backend will not be involved.

Flow to implement is an "Authorization Code flow with PKCE".

API will then accept both the access token it self-issued (for legacy username/password) and the access token issued by ory.sh.

When receiving an access token from ory.sh, it will be an opaque one. It will hence need to call ory.sh userinfo endpoint to retrieve token metadata and match the `sub`.

Each user will have a new `idp_sub` column inside the zimfarm database to match against.

Client ID to use is `d87a31d2-874e-44c4-9dc2-63fad523bf1b`.

Available endpoints:
- OpenID Connect Discovery Endpoint: `https://login.kiwix.org/.well-known/openid-configuration`
- Well-Known JSON Web Keys: `https://login.kiwix.org/.well-known/jwks.json`
- Authorize: `https://login.kiwix.org/oauth2/auth`
- Token: `https://login.kiwix.org/oauth2/token`
- User info: `https://login.kiwix.org/userinfo`

Note: if supporting both "local" username / password and "ory.sh" proves to be too painful, do not hesitate to speak-up, we do not really need a smooth migration, this was more meant as an elegant way to deploy it progressively without too much pain.

elfkuzco:
## Rationale
This PR begins integration of Kiwix oauth2 authentication with ory.sh

## Changes
- implement UI login flow for authentication with Kiwix with support for legacy API authentication. Two modes are supported: oauth with OIDC and oauth with self-service Session.
- fetch user scopes/roles from `/me` endpoint instead of storing the roles in tokens
- add new column (`idp_sub`) to users. This will the same as their `sub` from ory.sh
- seperated the three decoders: `local`, `oauth-oidc`, `oauth-session` into classes with a common interface. This simplifies usage in `get_jwt_claims_or_none`
- added new variables to configure the behaviour of the decoders:
    - `OAUTH_OIDC_CLIENT_ID`: (backend) the client ID while doing oauth with OIDC
    - `OAUTH_OIDC_LOGIN_REQUIRE_2FA ` (backend)  : if oauth with OIDC should require users to sign in with 2FA
    - `OAUTH_SESSION_AUDIENCE_ID ` (backend) : the audience ID if oauth is done with self-service session
    - `OAUTH_SESSION_LOGIN_REQUIRE_2FA ` (backend) : if oauth with self-service session should require 2FA
    - `OAUTH_JWKS_URI ` (backend) : JWKS URI
    - `AUTH_MODES` (backend) : comma-seperated values for detecting which decoder to use. default is `local,oauth-oidc,oauth-session` to specify the 3 decoders should be used. You can omit any and the decoder tool chain will omit a decoder if it's name isn't in the `AUTH_MODES`. Each value corresponds to the name of the decoder
    - `CREATE_NEW_OAUTH_ACCOUNT` (backend): create new user from oauth login if user with `idp_sub` does not not exist. User is created with `viewer` role.
    - `OAUTH_MODE` (frontend): can take one of `session` or `oidc`
    - `LOGIN_MODES` (frontend):  a list of login modes. Default is `["local", "oauth"]`. If `login` is included, user will see `Sign in with Username`. If `oauth` is included, User will see `Sign in with Kiwix`
    - `OAUTH_CLIENT_ID` (frontend): oauth client ID for interacting with oauth
    - `OAUTH_BASE_URL` (frontend): base url of oauth server (should not end in slash `/`)
- refactored auth store in frontend to use interfaces for doing auth-related logic
- remove user data in jwt token (will need to verify with zimit-frontend and wp1) to be sure they aren't using this
- store user data and tokens in localstorage instead of cookies

This closes #1509

### DIFF
--- a/backend/src/zimfarm_backend/api/constants.py
+++ b/backend/src/zimfarm_backend/api/constants.py
@@ -10,8 +10,11 @@ JWT_TOKEN_EXPIRY_DURATION = parse_timespan(
 )
 ZIM_ILLUSTRATION_SIZE = int(getenv("ZIM_ILLUSTRATION_SIZE", default="48"))
 
+# List of authentication modes. Allowed values are "local", "oauth-oidc",
+# "oauth-session"
 AUTH_MODES: list[str] = getenv(
-    "AUTH_MODES", default="local,oauth-oidc,oauth-session"
+    "AUTH_MODES",
+    default="local",
 ).split(",")
 
 # OAuth/OIDC configuration
@@ -21,13 +24,13 @@ OAUTH_JWKS_URI = getenv(
 )
 OAUTH_ISSUER = getenv("OAUTH_ISSUER", default="https://login.kiwix.org")
 OAUTH_OIDC_CLIENT_ID = getenv(
-    "OAUTH_OIDC_CLIENT_ID", default="d87a31d2-874e-44c4-9dc2-63fad523bf1b"
+    "OAUTH_OIDC_CLIENT_ID", default="38302485-7f3d-4e88-b1b3-ec02fb92ec8c"
 )
 OAUTH_OIDC_LOGIN_REQUIRE_2FA = parse_bool(
     getenv("OAUTH_OIDC_LOGIN_REQUIRE_2FA", default="true")
 )
 OAUTH_SESSION_AUDIENCE_ID = getenv(
-    "OAUTH_SESSION_AUDIENCE_ID", default="d87a31d2-874e-44c4-9dc2-63fad523bf1b"
+    "OAUTH_SESSION_AUDIENCE_ID", default="bada4130-5143-4524-a0bb-0d69671beee2"
 )
 OAUTH_SESSION_LOGIN_REQUIRE_2FA = parse_bool(
     getenv("OAUTH_SESSION_LOGIN_REQUIRE_2FA", default="true")
--- a/dev/docker-compose.yml
+++ b/dev/docker-compose.yml
@@ -47,14 +47,14 @@ services:
       BLOB_PUBLIC_STORAGE_URL: http://localhost:8003
       BLOB_STORAGE_USERNAME: zimfarm
       BLOB_STORAGE_PASSWORD: zimpass
-      OAUTH_JWKS_URI: https://tender-wiles-i8ance55ra.projects.oryapis.com/.well-known/jwks.json
-      OAUTH_ISSUER: https://tender-wiles-i8ance55ra.projects.oryapis.com/
+      OAUTH_JWKS_URI: https://login-staging.kiwix.org/.well-known/jwks.json
+      OAUTH_ISSUER: https://login-staging.kiwix.org
       OAUTH_OIDC_CLIENT_ID: 38302485-7f3d-4e88-b1b3-ec02fb92ec8c
       OAUTH_OIDC_LOGIN_REQUIRE_2FA: true
       OAUTH_SESSION_AUDIENCE_ID: 309693e7-ad5e-4379-bf93-ba89314230fd
       OAUTH_SESSION_LOGIN_REQUIRE_2FA: true
       CREATE_NEW_OAUTH_ACCOUNT: true
-      AUTH_MODES: local,oauth-oidc,oauth-session
+      AUTH_MODES: local
 
     command:
       - uvicorn
--- a/dev/frontend-ui-dev/config.json
+++ b/dev/frontend-ui-dev/config.json
@@ -3,7 +3,7 @@
   "MONITORING_URL": "http://localhost:8004/v3",
   "OAUTH_CLIENT_ID": "38302485-7f3d-4e88-b1b3-ec02fb92ec8c",
   "BLOB_MAX_SIZE": 1048576,
-  "OAUTH_BASE_URL": "https://tender-wiles-i8ance55ra.projects.oryapis.com",
+  "OAUTH_BASE_URL": "https://login-staging.kiwix.org",
   "OAUTH_MODE": "session",
-  "LOGIN_MODES": ["local", "oauth"]
+  "LOGIN_MODES": ["local"]
 }


benoit74:
@elfkuzco is kinda blocked in testing this code by the fact that:
- ory.sh doesn't want to set CORS to `localhost` (see https://www.ory.com/docs/guides/cors)
- they recommend to use Ory Tunnel, but Ory Tunnel seems to need an API key ... which gives full access on the project ... not the kind of access we want to give to developers

I've posted a message on Ory Slack to get assistance, but it looks like we do not have any support besides documentation in our ory.sh plan.

benoit74:
We could probably use our own tunneling service in fact.

I've configured "https://*loca.lt" (from https://localtunnel.me/) as redirect URL and CORS, let's me know if it does work (or not, but let's hope for the best)

elfkuzco:
The local tunnel closes unexpectedly after some minutes and says tunnel error. I think ngrok is more reliable than them. I have been using their free tunneling and it works quite fine You could try: https://*ngrok-free.app

benoit74:
@elfkuzco can you try to:
- not ask for a refresh token anymore on ory.sh
- not ask for the IDtoken (we don't use it, so removing 'openid' from scope should be fine)
- implement the silent authentication with silent iframe suggested in https://github.com/kiwix/overview/issues/134#issuecomment-3569767993
- add a configuration flag allowing to automatically create a new DB user with "no rights" (only viewer) whenever we get an authenticated user (from Kiwix) but no matching DB user

elfkuzco:
updated PR description and branch because previous PR created a migration, thus, I had to pull those updates and make sure this PR set the down revision properly.

kelson42:
> I'm however unsure about how to handle the fact that:
> 
>     * `email` is maybe not that a good thing to store in Zimfarm DB (PII)
> 
>     * `name` from Ory.sh is probably a bit different than `username`, especially since it is not guaranteed to be unique at all (and someone could choose `admin` to try funny things)
> 
>     * we use the `username` in many place in the DB / UI to store user actions, but this was a bit risky (username might change) and now it is just deemed to not work anymore (username is not unique anymore at all)
> 
> 
> I wonder if we should:
> 
>     * stop storing user email in DB (it is not used by the app AFAIK and was more a convenience for us to contact worker admins typically)
> 
>     * stop using `username` to track user actions but use user ID
> 
>     * adapt UI to transform user ID into user display name
> 
>     * send ID token to the backend when user is logged in to automatically store / update user display name in Zimfarm DB

AFAP we should not have duplicate information at the Zimfarm because of consistency and security concerns. This is very important. We should better identify user via an id which is not a private data (UUID for example) and ideally this is the only non-zimfarm-specific profile information we store on our side. Means all the rest is transient and dynamically retrieved from the IDP.

benoit74:
I agree about the general idea, but then comes details. If we only store user ID in for instance schedule history, how do we transform it into a display name? The history a given user is viewing has significant chances to reference many other users IDs, and contacting the IdP for each of them is not a native function.

benoit74:
We have confirm my plan with @kelson42:

~~I wonder if~~ we ~~should~~ will:

- stop storing user email in DB (it is not used by the app AFAIK and was more a convenience for us to contact worker admins typically)
- stop using username to track user actions but use user ID (migration needed)
- adapt ~~UI~~ backend to transform user ID into user display name
- send ID token to the backend when user is logged in to automatically store / update user display name in Zimfarm DB

I've already configured Ory.sh to request new users to enter both an email address and a `User Name` (stored as `username`, not unique. Please recreate your personal account on ory.sh so that it gets correctly populated, I've deleted all personal accounts.

benoit74:
@elfkuzco please do not remove the email field for now, we will do it in the aftermath. but do not request it in the scopes (i.e. keep it as today, just request the `profile` to get the `name`, and `openid` to get the ID token. And send both to the backend.

We also need to enforce MFA is used on all users. If I read the doc correctly, you need to pass `acr_values=aal2` in the authentication code with pkce flow so that ory forces users to use MFA. And on the backend, you need to check that `idToken.acr` is `aal2`, and if not, ask the UI to restart auth to elevate privileges.

benoit74:
Edit: you can probably just send the ID token to the backend instead of the access token. Should be fine from what I read, unless I miss something.

elfkuzco:
> stop using username to track user actions but use user ID (migration needed)


what of places where the author is set to `maint-scripts` or canceled by `periodic-scheduler`. Do we need a dedicated account for these?

benoit74:
Good point. Yes, let's create special accounts for these, without password and role so no-one can use them.

elfkuzco:
> We also need to enforce MFA is used on all users. If I read the doc correctly, you need to pass acr_values=aal2 in the authentication code with pkce flow so that ory forces users to use MFA. And on the backend, you need to check that idToken.acr is aal2, and if not, ask the UI to restart auth to elevate privileges.



I don't think the `acr_values=aal2` works as expected for ory. I did a bit of digging and one way you can check the authentication method is via session. given we don't use api key anymore to verify session, this isn't possible on the backend. another way is to check the `amr` which is included in the `id_token`. But so, far, I have only seen `password` and `code` int the `amr` array which conforms to the authentication mechanisms I see when I try to login

From what I got in the community AI bot, it seems this requirement would have to be enforced at ory side
https://www.ory.com/docs/kratos/mfa/step-up-authentication#configuration

backend/src/zimfarm_backend/api/routes/dependencies.py
```diff
@@ -64,7 +88,22 @@ def _get_current_user_or_none(
     ) -> User | None:
         if claims is None:
             return None
-        return get_user_by_id_or_none(session, user_id=claims.subject)
+        user = get_user_by_id_or_none(session, user_id=claims.sub)
+        if user is None:
+            user = get_user_by_idp_sub_or_none(session, idp_sub=claims.sub)
```
benoit74: Why not merging `get_user_by_idp_sub_or_none` into `get_user_by_id_or_none` and simply doing an "OR" on `id` and `idp_sub` fields? since both are very unique, I don't see any risk and it keeps the code a bit cleaner

elfkuzco:
> what if someone arrives with a username which is already taken by someone else in the DB? we should block this situation at API level and display a clear warning in the UI explaining the issue ; it looks like we simply get a DB Integrity error

The warning is already shown in the UI
Screenshot_20251215_055908

elfkuzco:
> what if someone does not consent to (few) scope(s) in the consent screen ?  it looks like we simply do not get expected fields or no idtoken at all ; I'm not sure how we should handle this situation

I currently have some fields as `required` while decoding the token:
https://github.com/openzim/zimfarm/blob/656bfe3df70cd4afb9021aacbbd2ec1a83331407/backend/src/zimfarm_backend/api/token.py#L27-L36

So, if for some reason, the user disallows viewing their profile, the `name` field wouldn't exist and the decoding would raise an error about a required field missing.

backend/src/zimfarm_backend/api/routes/dependencies.py
elfkuzco: done

benoit74:
After discussing with Ory person, he advised us to not use OIDC but prefer session-based login with JWT exchange.

I've setup the necessary on our Ory project, and documented a bit at https://github.com/kiwix/overview/wiki/Ory.sh-SSO

@elfkuzco can you please:
- add support for session-based login (most changes are in the frontend)
- keep supporting OIDC (since it is already implemented, no need to remove it, I'm quite sure we will need it in some deployment)
- add configuration flag (only in frontend I think) to decide which login method we wanna support: local, OIDC and session-based ; we should be able to activate the three methods at the same time, or only one, and sign-in UI should adapt 
- this means we should make the backend `get_jwt_claims_or_none` function more agnostic, accepting a list of valid decoding solutions through environment variables; we can safely assume we will not support more than 3 way of decoding a JWT ; with something like `JWT_1_ISSUER`, `JWT_1_ALGORITHM`, `JWT_1_SIGNING_KEY`, `JWT_1_JWKS_URL`, `JWT_1_ISSUER`, `JWT_1_AUDIENCE`, `JWT_1_REQUIRED_CLAIMS`, `JWT_1_REQUIRE_MFA_METHOD` (none, amr, aal) in terms of environment variables 
- add configuration flag in frontend to customize the signin button labels (e.g. "Sign-in with user", "Sign-in with Kiwix", ...)
- rework (environment) variables, files and functions names to remove the "Kiwix" specificity ; there is not much specific to Kiwix in most of what we've developed here, only OIDC or session-based (which is a bit specific to Ory ATM at least)

elfkuzco:
@benoit74 , check the updated README of the PR to see the different variables for configuring which decoder/ oauth mode to use.

benoit74:
DB is a bit inconsistent, we have `admin` author in `schedule_history` instead of `maint-scripts` for initial schedule_history versions. 

I've updated it with `UPDATE schedule_history SET author = 'maint-scripts' WHERE author = 'admin';` as discussed live, we should add this to the migration `6dcfce58220a`.

elfkuzco:
Done

backend/src/zimfarm_backend/api/constants.py
```diff
@@ -9,3 +9,29 @@
     getenv("JWT_TOKEN_EXPIRY_DURATION", default="1d")
 )
 ZIM_ILLUSTRATION_SIZE = int(getenv("ZIM_ILLUSTRATION_SIZE", default="48"))
+
+AUTH_MODES: list[str] = getenv(
+    "AUTH_MODES", default="local,oauth-oidc,oauth-session"
```
benoit74: I would prefer default to be `local` for smooth local development ; maybe add a comment with all allowed values

backend/src/zimfarm_backend/api/constants.py
```diff
@@ -9,3 +9,29 @@
...
+).split(",")
+
+# OAuth/OIDC configuration
+OAUTH_JWKS_URI = getenv(
+    "OAUTH_JWKS_URI",
+    default="https://login.kiwix.org/.well-known/jwks.json",
+)
+OAUTH_ISSUER = getenv("OAUTH_ISSUER", default="https://login.kiwix.org")
+OAUTH_OIDC_CLIENT_ID = getenv(
+    "OAUTH_OIDC_CLIENT_ID", default="d87a31d2-874e-44c4-9dc2-63fad523bf1b"
```
benoit74: New prod client ID is `38302485-7f3d-4e88-b1b3-ec02fb92ec8c`

backend/src/zimfarm_backend/api/constants.py
```diff
@@ -9,3 +9,29 @@
...
+)
+OAUTH_OIDC_LOGIN_REQUIRE_2FA = parse_bool(
+    getenv("OAUTH_OIDC_LOGIN_REQUIRE_2FA", default="true")
+)
+OAUTH_SESSION_AUDIENCE_ID = getenv(
+    "OAUTH_SESSION_AUDIENCE_ID", default="d87a31d2-874e-44c4-9dc2-63fad523bf1b"
```
benoit74: New prod audience is `bada4130-5143-4524-a0bb-0d69671beee2`

dev/docker-compose.yml
```diff
@@ -47,6 +47,14 @@ services:
       BLOB_PUBLIC_STORAGE_URL: http://localhost:8003
       BLOB_STORAGE_USERNAME: zimfarm
       BLOB_STORAGE_PASSWORD: zimpass
+      OAUTH_JWKS_URI: https://tender-wiles-i8ance55ra.projects.oryapis.com/.well-known/jwks.json
```
benoit74: Update with staging configuration now that it is ready:

```
      OAUTH_JWKS_URI: https://login-staging.kiwix.org/.well-known/jwks.json
      OAUTH_ISSUER: https://login-staging.kiwix.org
```

```

File: dev/docker-compose.yml [hunk 6]
```diff
@@ -47,6 +47,14 @@ services:
...
+      OAUTH_ISSUER: https://tender-wiles-i8ance55ra.projects.oryapis.com/
+      OAUTH_OIDC_CLIENT_ID: 38302485-7f3d-4e88-b1b3-ec02fb92ec8c
+      OAUTH_OIDC_LOGIN_REQUIRE_2FA: true
+      OAUTH_SESSION_AUDIENCE_ID: 309693e7-ad5e-4379-bf93-ba89314230fd
+      OAUTH_SESSION_LOGIN_REQUIRE_2FA: true
+      CREATE_NEW_OAUTH_ACCOUNT: true
+      AUTH_MODES: local,oauth-oidc,oauth-session
```
benoit74: Again, default should be `local` for smoother local development.

dev/frontend-ui-dev/config.json
```diff
@@ -1,5 +1,9 @@
 {
   "ZIMFARM_WEBAPI": "http://localhost:8000/v2",
   "MONITORING_URL": "http://localhost:8004/v3",
-  "BLOB_MAX_SIZE": 1048576
+  "OAUTH_CLIENT_ID": "38302485-7f3d-4e88-b1b3-ec02fb92ec8c",
+  "BLOB_MAX_SIZE": 1048576,
+  "OAUTH_BASE_URL": "https://tender-wiles-i8ance55ra.projects.oryapis.com",
+  "OAUTH_MODE": "session",
+  "LOGIN_MODES": ["local", "oauth"]
```
benoit74: Again, default should be `local`

dev/frontend-ui-dev/config.json
```diff
@@ -1,5 +1,9 @@
 {
   "ZIMFARM_WEBAPI": "http://localhost:8000/v2",
   "MONITORING_URL": "http://localhost:8004/v3",
-  "BLOB_MAX_SIZE": 1048576
+  "OAUTH_CLIENT_ID": "38302485-7f3d-4e88-b1b3-ec02fb92ec8c",
+  "BLOB_MAX_SIZE": 1048576,
+  "OAUTH_BASE_URL": "https://tender-wiles-i8ance55ra.projects.oryapis.com",
```
benoit74: Change to `https://login-staging.kiwix.org`

benoit74 (approved):
LGTM, some very minor tweaks and you can merge. Production is already ready for a dark launch.

PR #1527, opened 2025-11-17
create schedule history while using update_scraper_version script

### LINKED ISSUE #1526: Enhance `backend/maint-scripts/update_scraper_version.py` script to create schedule versions
Currently, the `backend/maint-scripts/update_scraper_version.py` directly updates schedule without creating a proper schedule version in the history. This causes changes made by the script to be invisible in the history + attached to next manual change.

The script should create a proper history version + attach a nice comment to it (short, just saying it is the script)

elfkuzco:
## Rationale
The `update_scraper_version` does not create schedule histories after updating a schedule image tag or offliner definition. By hooking into the existing `update_schedule` function, we can ensure that schedule histories will be created when update is made either via API or CLI script.

## Changes
- call db `update_schedule` from CLI script to update scraper versions
- seperate code to update schedule and requested task
- show secrets while dumping to database in update name mappings function

This fixes #1526

### DIFF
--- a/backend/maint-scripts/update_scraper_version.py
+++ b/backend/maint-scripts/update_scraper_version.py
@@ -49,23 +49,26 @@ database. If you must use the -n flag, here are some examples:
 """
 
 import argparse
+from typing import cast
 
 import sqlalchemy as sa
 from sqlalchemy.orm import Session as OrmSession
 from sqlalchemy.orm.attributes import flag_modified
 
 from zimfarm_backend import logger
-from zimfarm_backend.common.schemas.offliners.builder import generate_similarity_data
+from zimfarm_backend.common.schemas.models import ScheduleConfigSchema
 from zimfarm_backend.common.schemas.orms import OfflinerDefinitionSchema, OfflinerSchema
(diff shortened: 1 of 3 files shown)


backend/maint-scripts/update_scraper_version.py
```diff
@@ -101,23 +103,91 @@ def update_entries(
...
                 f"to {args.image_tag}..."
             )
             obj.config["image"]["tag"] = image_tag
             flag_modified(obj, "config")
 
         if offliner_definition:
-            logger.info(
-                f"setting offliner defintion for {model.__tablename__} {obj_name} ..."
-            )
+            logger.info(f"setting offliner defintion for requested task {obj.id}...")
```
benoit74: defintion => definition

benoit74 (changes requested):
I'm a bit puzzled by the fact that we have a "setting offliner defintion for requested task" message but not a "setting offliner defintion for schedule" ; is this normal ? Can we find a way to still have this message for schedules as well (if not too complicated of course)

benoit74 (approved):
LGTM

Issue #1530, opened 2025-11-17
Locking statements at DB level


benoit74:
I finally achieved to catch a locking condition at DB level which might well be the cause of https://github.com/openzim/zimfarm/issues/1307

| blocked\_pid | blocked\_user | blocking\_pid | blocking\_user | blocked\_statement | current\_statement\_in\_blocking\_process |
| :--- | :--- | :--- | :--- | :--- | :--- |
| 646800 | zimfarm | 646913 | zimfarm | UPDATE task SET updated\_at=$1::TIMESTAMP WITHOUT TIME ZONE, container=$2::JSONB WHERE task.id = $3::UUID | SELECT file.id, file.name, file.status, file.size, file.cms\_on, file.cms\_notified, file.created\_timestamp, file.uploaded\_timestamp, file.failed\_timestamp, file.check\_timestamp, file.check\_result, file.check\_log, file.check\_details, file.check\_filename, file.check\_upload\_timestamp, file.info, file.task\_id <br/>FROM file <br/>WHERE file.task\_id = $1::UUID AND file.name = $2::VARCHAR |

@elfkuzco can you please analyze the codebase and check where we are doing this select statement and why it is taking so long to complete?

Here are logs of the backend at the same moment (remember we have two instances of the API server running at the same time):

[Logs-2025-11-17 17_20_48.txt](https://github.com/user-attachments/files/23587693/Logs-2025-11-17.17_20_48.txt)

elfkuzco:
Found some similar issues on this and the generally suggested approach was to configure the engine pool size and max overflow (https://docs.sqlalchemy.org/en/13/errors.html#queuepool-limit-of-size-x-overflow-y-reached-connection-timed-out-timeout-z). What do you think should be reasonable values for these? So, it's not necessarily those two queries that cause the issue. Any other queries could be shown at the particular time.

Also, it appears there are places where the query could be optimized by eagerly loading objects with `selectinload` or `joinedload` instead of lazy loading anytime we want to access a related attribute, which is the root cause of [N+1](https://docs.sqlalchemy.org/en/20/glossary.html#term-N-plus-one-problem) problem in SQL queries. I will create a PR for this accordingly.

benoit74:
[...]

PR #1535, opened 2025-11-20
skip loading task files in save_event handler

elfkuzco:
## Rationale
In the events handler, the task is loaded along with its files. Also, the call to `create_or_update_file` fetches the updated files. These two scenarios happening concurrently could be the plausible cause of #1530 when file events come in rapidly. 

By skipping the option to load files along with a task and using postgresql's returning clause on `upsert`, we can mitigate this happening as we don't need to load the task files anymore.

## Changes
- avoid loading task files in `save_event` handlers
- use postgresql's returning statement to get the updated file

Will not link the issue to be closed by this PR until we can be sure it doesn't happen again.

### DIFF
--- a/backend/src/zimfarm_backend/common/external.py
+++ b/backend/src/zimfarm_backend/common/external.py
@@ -218,7 +218,7 @@ def advertise_book_to_cms(session: so.Session, task: TaskFullSchema, file_name: 
             )
 
     # record request result
-    task.files[file_name] = create_or_update_task_file(
+    create_or_update_task_file(
         session,
         FileCreateUpdateSchema(
             task_id=task.id,
--- a/backend/src/zimfarm_backend/common/utils.py
+++ b/backend/src/zimfarm_backend/common/utils.py
@@ -122,9 +122,7 @@ def save_event(
 ):
     """save event and its accompanying data to database"""
 
-    task = session.scalars(
-        select(Task).options(so.selectinload(Task.files)).where(Task.id == task_id)
-    ).one_or_none()
+    task = session.scalars(select(Task).where(Task.id == task_id)).one_or_none()
     if task is None:
         raise RecordDoesNotExistError(f"Task {task_id} does not exist")
     schedule = task.schedule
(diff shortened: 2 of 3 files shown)


benoit74:
I don't get how this could solve anything. SELECT statements are not blocking anything AFAIK, so loading less or more data is not going to help. Can you develop your reasoning?

benoit74:
(and here the blocked statement is an UPDATE, not a SELECT, so we probably have two concurrent UPDATE statements)

elfkuzco:
I thought `current_statement_in_blocking_process` meant the statement that is running in the process that is blocking and `blocked_statement` means the statement that is prevented from being executed because there's no session. Seems to be a misunderstanding.

benoit74:
Disclaimer: everything below is my understanding, and this is a complex topic, so do not hesitate to tell if it looks like I'm wrong / misunderstanding / misinterpreting something.

The `blocked statement` is indeed the statement that is prevented from being executed, but this is blocked by what we call a lock, not by a lack of available session (it already has a session ... or a connection in PG terms). Most common locks are held at row level, but it could also be other rarer events.

The `current_statement_in_blocking_process` is the statement currently running in the process which holds the lock preventing other statement to run. It gives no clue on which statement caused the lock to be here, because as far as I've understood PG does not track it.

After some thought, what I would suspect is that we have indeed received two update events for the same task at almost the same time. Since reading files used to take a potentially long time due to size of check results / check logs in some cases, there is a significant probability that the blocking statement was just transferring that data from DB to the API.

Looking at where the condition File.task_id + File.name is used in the code, the blocking statement was indeed very probably running the `get_task_file_or_none` function. So your way to fix the issue is not totally wrong, but I don't think it will really work because it will still fetch the file.

I suspect most of the issue is now resolved since reading the file back is going to be very fast now that we do not have data to transfer anymore.

A more radical question could be "do we really need the whole file data?" or even "do we really need the file back?". Looking at the codebase, it looks like we do not really care about having that file data back (it is simply dropped in the `handle_file_event`, and it is used to update the task in `advertise_book_to_cms` but then we do not use it anymore.

So my suggestion is indeed to stop loading the files in `save_event` (we never use this, even when we want to update files info) and to stop returning anything in `create_or_update_task_file`. This will not bring much gain now that check_results are not slowing down retrieval of files anymore, but it cannot really harm to enhance performance / reduce load on the DB.

Does it makes sense to you?

elfkuzco:
Yes

elfkuzco:
Applied the change to stop loading files from create_or_update_task_file

PR #1537, opened 2025-11-21
set nginx cache rules for files

### LINKED ISSUE #537: Css/js resources should be cached by browsers
Proper HTTP response headers should be set

elfkuzco:
## Rationale
This PR sets up cache control headers for assets and resources served by nginx. 

## Changes
- add cache control headers for css/js/image files
- no cache for html and json files

This fixes #537

### DIFF
--- a/frontend-ui/nginx-default.conf
+++ b/frontend-ui/nginx-default.conf
@@ -15,4 +15,21 @@ server {
     location @rewrites {
         rewrite ^(.+)$ /index.html last;
     }
+
+    location = /index.html {
+	add_header Cache-Control "no-cache";
+    }
+
+
+    location ~* \.json$ {
+	add_header Cache-Control "no-cache";
+    }
+
+    location ~* \.(js|css)$ {
+	add_header Cache-Control "public, max-age=2592000, immutable";
+    }
+
+    location ~* \.(png|jpg|jpeg|gif|webp|ico|svg)$ {
+	add_header Cache-Control "public, max-age=2592000";
+    }
 }


elfkuzco:
> Is this really working? Most our UI URLs are not ending with .html. I did not tested locally tbh.

Yup. Initially added the `*.html` because the server cache rules weren't added with the `location @rewrites`. But I think the `location = /index.html` adds it. Will remove the following one with `*.html`

elfkuzco:
removed the rules for all html files execpt for `index.html` which is needed for the rewrite

elfkuzco:
According to https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Cache-Control, `no-cache` doesn't necessarily mean "don't cache". It allows caches to store a response but requires them to revalidate it before reuse. `no-store` indicates that any caches of any kind should not store this response. Figured we want the `no-store` for these files so there's no caching of any kind.

elfkuzco:
> Why no-cache, no-store and not simply no-cache on HTML and JSON files?

Should I restore it?

benoit74:
What is the issue if we have caching and browser re-validates each time?

Fetching these resources each time looks like a waste to me, but I might miss something.

elfkuzco:
> What is the issue if we have caching and browser re-validates each time?

I think I might have abused it and used it for a case which it was not intended - prevent release of sensitive information
https://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html#sec14.9.2

I would like to remove it.

benoit74:
OK, let's do that.

PR #1540, opened 2025-11-25
add offliner definition version to schedule history entry

### LINKED ISSUE #1539: offliner definition version is not tracked in schedule history
I've just updated both image version and offliner definition, and the history shows only the image version has changed, see https://farm.openzim.org/recipes/gutenberg_mul_all/history?compare=82db4635-f26d-4f37-a895-d74cdb00a3bd...0ec47a47-817c-404e-ac09-3bd7adc6316a

elfkuzco:
## Rationale
By adding the offliner definition version to the schedule history entry, we can see the changes to the version recorded in the history. 

## Changes
- add column to keep track of offliner definition updates in schedule history

This closes #1539


backend/src/zimfarm_backend/migrations/versions/8b0210f9d9ab_store_offliner_definition_version_in_.py
```diff
@@ -0,0 +1,31 @@
...
+down_revision = "b298a5831211"
+branch_labels = None
+depends_on = None
+
+
+def upgrade() -> None:
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.add_column(
+        "schedule_history",
+        sa.Column("offliner_definition_version", sa.String(), nullable=True),
```
benoit74, line 24: Shouldn't we also set `offliner_definition_version` of latest schedule_history of a given schedule to the current schedule `offliner_definition_version`, so that next change does not record it as a modification? For older schedule_history entries, it is impossible to know so keeping `NULL` is probably OK.

elfkuzco: done

elfkuzco:
It didn't work because the subquery re-selects from `schedule_history` and it seems PostgreSQL doesn't allow referencing the update target from inside the `FROM` clause when that table also appears in a subquery inside the `FROM`. 
```
zf_backend     | sqlalchemy.exc.ProgrammingError: (psycopg.errors.UndefinedTable) invalid reference to FROM-clause entry for table "sh"
zf_backend     | LINE 13:                 ON latest.history_id = sh.id
zf_backend     |                                                 ^
zf_backend     | DETAIL:  There is an entry for table "sh", but it cannot be referenced from this part of the query.
zf_backend     | [SQL:
zf_backend     |             UPDATE schedule_history sh
zf_backend     |             SET offliner_definition_version = od.version
zf_backend     |             FROM schedule s
zf_backend     |             JOIN offliner_definition od
zf_backend     |                 ON s.offliner_definition_id = od.id
zf_backend     |             JOIN (
zf_backend     |                 SELECT DISTINCT ON (schedule_id)
zf_backend     |                     id AS history_id, schedule_id
zf_backend     |                 FROM schedule_history
zf_backend     |                 ORDER BY schedule_id, created_at DESC
zf_backend     |             ) latest
zf_backend     |                 ON latest.history_id = sh.id
zf_backend     |             WHERE sh.schedule_id = s.id;
zf_backend     |             ]
```

Fixed it by using a  CTE instead

Issue #1561, opened 2025-12-09
Small issues in new zimfarm monitoring stack


benoit74:
While trying to deploy the `zimfarm-monitor-parent` image on Zimfarm, I experienced some issues.

The core of the issue is that generation of `/etc/netdata/stream.conf` seems to not produce a consistent output. At least quite often the system observes that this `stream.conf` file has changed and decides to restart container.

This is causing regular restarts of the container.

And since `/etc/netdata` was originally not meant to be persisted, it causes forever restarts. Once I've started to persist this folder the stack would achieve to start, but as soon as cronjob triggers to update `stream.conf` it restarts the container. I'm not saying this is the appropriate solution.

I'm not 100% sure what the solution should be, but this needs to be investigated and fixed.

If we decide to persist `/etc/netdata` folder, then we need to fix the generation of `netdata.conf` which get a new row with `hostname=....` each time the pod restarts (two rows actually, a blank one is also added).

I also still struggle to see "the Containers & VMs section on the right sidebar.", see e.g. https://monitoring.openzim.org/host/wikipedia_fr_test_213b6.ondemand#after=1765224510000;before=1765224525000;theme=slate;utc=Africa/Bamako ; not sure if this is my fault or something missing in the image.

And final thing, it looks like we can have intermittent issues retrieving the scraper container IP; see https://farm.openzim.org/pipeline/da3aa2b5-7a2e-48fc-9ed7-299b06df0f48/debug (first task failed, second task succeeded, so I assume it might be that we are looking for IP too fast and container does not yet have an IP, could be a matter of milliseconds probably).

PR #1562, opened 2025-12-09
fix issues with monitor parent restarting multiple times

elfkuzco:
## Changes
- sort users by username to ensure order remains the same. This is one reason why monitor is restarting since the users are returned in any other and the md5 checksum of the `stream.conf` file would be different
- generate `stream.conf` in `entrypoint.sh` first before starting netdata for the first time
- avoid touching `stream.conf` in Dockerfile since it's going to be generated in `entrypoint.sh`. This avoids killing the container and restarting it
- point monitoring url to v3 UI which does not have a  login screen
- add error handling to output of regen-stream.conf file
- set up cronjob in entrypoint.sh instead with env variables

### DIFF
--- a/monitor/parent/entrypoint.sh
+++ b/monitor/parent/entrypoint.sh
@@ -40,7 +40,20 @@ fi
 
 # Run update-stream-whitelist on startup
 echo "Running initial stream whitelist update"
-/usr/local/bin/update-stream-whitelist.sh &
+/usr/local/bin/update-stream-whitelist.sh
+
+# Setup cron job with environment variables
+echo "Setting up cron job with environment variables"
+cat > /etc/cron.d/update-stream-whitelist <<EOF
+DOCKER_USR=${DOCKER_USR}
+NETDATA_LISTENER_PORT=${NETDATA_LISTENER_PORT}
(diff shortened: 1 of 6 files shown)


elfkuzco:
updated the PR description

monitor/parent/entrypoint.sh
```diff
@@ -38,9 +38,22 @@ if mountpoint -q /etc/netdata && [ -z "$(ls -A /etc/netdata)" ]; then
   cp -a /etc/netdata.stock/. /etc/netdata
 fi
 
+# Setup cron job with environment variables
```
benoit74, line 45: Why setting up the cron before running initial stream whitelist update?

To me it looks like it gives a (bad) opportunity to have both cron + initial run in parallel should cron trigger very fast. Very unlikely, but very nasty to debug. If there is no other reasons, I would prefer to run initial update then setup cron.

elfkuzco: done.

elfkuzco:
i'm going to leave the issue #1561 open till we can verify container & vm section shows because i can always see it in local

PR #1566, opened 2025-12-12
enhance ui to upload blobs from schedule editor

### LINKED ISSUE #1558: add support for ui editor to upload blobs
The recipe editor should allow the user to upload local files or input the text content of the files depending on it's kind. On recipe create/update, zimfarm API should upload the files to drive.farm.openzim.org and store it's url

elfkuzco:
## Changes
- remove logic to get_base_type from annotations since this is the same information coming from the schema. The function inspects the type of the class and for cases like `blob`, it would mean defining a custom type. Rather than do this, we just pass the information to the field class and retrieve it there.
- impose size limit on file uploads (default: 1MB)
- allow UI to upload files to the backend

This closes #1558

### DIFF
--- a/frontend-ui/src/components/BlobEditor.vue
+++ b/frontend-ui/src/components/BlobEditor.vue
@@ -1,0 +1,201 @@
+<template>
+  <div>
+    <div class="d-flex align-center">
+      <v-text-field
+        v-model="displayFileName"
+        :label="label"
+        density="compact"
+        variant="outlined"
+        readonly
+        :hint="description ?? undefined"
+        persistent-hint
+        :hide-details="hasError ? false : 'auto'"
+      >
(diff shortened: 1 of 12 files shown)


frontend-ui/src/components/BlobEditor.vue
```diff
@@ -0,0 +1,196 @@
...
+  description?: string | null
+  maxSize?: number // in bytes, default 1MB
+}
+
+const props = withDefaults(defineProps<Props>(), {
+  label: 'File',
+  kind: 'image',
+  required: false,
+  description: null,
+  maxSize: 1 * 1024 * 1024, // 1MB default
```
benoit74: Please allow to configure this with `config.json`

PR #1569, opened 2025-12-12
prevent duplicate blob uploads

elfkuzco:
## Rationale
A blob is considered unique based on it's `checksum`, `flag_name` and `schedule_id` . Thus, if a user reuploads the same file for the same `flag_name` and `schedule_id`, we should not make another upload if we have already uploaded this file.

## Changes
- refactor transformer logic of processing uploads to just prepare the uploads with the all data computed
- process schedule blobs in the db calls to create/update a schedule using the prepared blobs. using the prepared blobs, we can check if a file has been uploaded before for the schedule and flag combination to prevent duplicate uploads
- add functions to retrieve blobs from db using the characteristics that make it unique


benoit74 (approved):
LGTM

PR #1578, opened 2025-12-15
add script to migrate old schedule blobs to table entries

### LINKED ISSUE #1575: Migration existing blobs to the new data structure
This is a subtask of https://github.com/openzim/zimfarm/issues/1362

We need to migrate existing files to the new blob structure. See https://github.com/openzim/zimfarm/issues/1572 for the list of supported flags.

We need a migration script to:
- create blob record in DB
- update schedule accordingly (let's assume we don't need to update tasks and requested tasks, this will be fine, I will take care of having a clean situation when running the migration script)
- move file on drive.farm.openzim.org to proper new location

elfkuzco:
## Rationale
With the introduction of blob tables, old schedules will need to have their files moved to the top-level on WebDAV and information concerning it's checksum and url will need to be added to blob table. This PR adds a migration script to perform the desired functionalities.

## Changes
- fetch schedule blobs from URL, compute it's checksum, move to new name and persist new url to DB
- create entry to track schedule blob

This closes #1575

### DIFF
--- a/backend/maint-scripts/update_schedule_blobs.py
+++ b/backend/maint-scripts/update_schedule_blobs.py
@@ -1,0 +1,203 @@
+#!/usr/bin/env python3
+
+import hashlib
+from typing import Literal
+
+import requests
+from pydantic import AnyUrl
+from sqlalchemy import select
+from sqlalchemy.orm import selectinload
+
+from zimfarm_backend import logger
+from zimfarm_backend.common.constants import (
+    BLOB_PRIVATE_STORAGE_URL,
+    BLOB_PUBLIC_STORAGE_URL,
+    BLOB_STORAGE_PASSWORD,
+    BLOB_STORAGE_USERNAME,
+    REQUESTS_TIMEOUT,
+)
+from zimfarm_backend.common.schemas import BaseModel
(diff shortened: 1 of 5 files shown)


elfkuzco:
Current code moves the files to the new generated name. However, this could be problematic if a file is shared by many schedules. Is that a possible scenario?

benoit74:
> Is that a possible scenario?

I'm quite sure it does exist indeed. Please rework logic to rather first copy the files to new URL, keeping a list of files to delete, and then delete them in the aftermath. I don't think we have that many blobs, so it should be fine.

backend/maint-scripts/update_schedule_blobs.py
```diff
@@ -0,0 +1,180 @@
...
+                            data=schedule.config["offliner"],
+                            skip_validation=True,
+                        ),
+                    },
+                    context={"skip_validation": True},
+                )
+                for flag in flag_mappings:
+                    old_url = getattr(schedule_config.offliner, flag.flag_name)
+                    if not old_url:
+                        logger.warning(
```
benoit74: Why a warning? all flags are optional AFAIK, so most recipes will not have the flag, this is normal

PR #1582, opened 2025-12-16
prevent registering changes if blobs flags are the same

elfkuzco:
## Rationale

On the UI, when a blob is added, the flag is set to the base64 string of the blob. However, sometimes, the user might be re-uploading a blob they already added or a previous version and the we detect them as changes. To prevent this, the backend exposes an endpoint to retrieve a schedules' list of blobs. This allows the UI to compute the checksum of a blob and then check if we have that checksum in our blob list to better detect if this was indeed a new blob

## Changes
- add endpoint to return a schedules' list of blobs
- compute checksum of each blob after edit on the ui
- verify if blob checksum changed and enable/disable button to update changes
- use the diff object on all other flags aside from blobs as it would always find a difference between a base64 string and a URL (which the backend sets). To mitigate this, we explicitly check the checksum of blobs.

### DIFF
--- a/backend/src/zimfarm_backend/api/entrypoint.py
+++ b/backend/src/zimfarm_backend/api/entrypoint.py
@@ -10,6 +10,7 @@ from fastapi.responses import JSONResponse
 from pydantic import ValidationError
 
 from zimfarm_backend.api.routes.auth.logic import router as auth_router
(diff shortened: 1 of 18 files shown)


benoit74 (approved):
LGTM

Issue #1589, opened 2025-12-16
Better track provenance of blobs


benoit74:
In https://github.com/openzim/zimfarm/issues/1362#issuecomment-3588407141

> If data comes from pasting/importing an URL, we should keep track of it

In https://github.com/openzim/zimfarm/issues/1362#issuecomment-3588486329

> This all only focuses on the storing/input of the image. What about the making of it? Whenever I have to make an icon or illustration, I always need to save next to it the URL of the original icon I used. That's important information. Maybe putting it in a comment is enough. If it goes beyond resizing, one may want to keep the source file used. If I spend 30mn on tweaking an icon, I would want to keep it… where ? In the drive and we're back to its limitations.

We need to clarify what / how we want to ask for / store / display all this data

benoit74:
Let's start by adding a free text comment next to each blob, editable / viewable from the UI

This is in fact mandatory before we can apply https://github.com/openzim/zimfarm/issues/1590 / https://github.com/openzim/zimfarm/pull/1599

elfkuzco:
Should this be a column in the `blob` table? And since it won't be sent as part of the config, I'm guessing it would be sent as a top-level entry in the JSON payload? Is that okay?

PR #1599, opened 2025-12-18
enhance schedule blob migration script to handle blobs hosted on different storage

### LINKED ISSUE #1590: Migrate image and css flags which were not hosted on the blob storage
We need another maintenance script to migrate image and CSS flags which were not stored in the blob storage as well.

Two possibility:
- in most cases, the image or CSS is still available so we just need to upload it to blob storage, create blob and fix schedule
- in few cases, the image or CSS are not available anymore, we need to report them for manual fixing by content team

The maintenance script should handle both situations, commit after every schedule handled (it could take long to download all assets) and display a markdown report of assets not available anymore with schedule, flag, url of asset columns, ready to be put into a Github issue

elfkuzco:
## Rationale
This PR enhances the script to migrate schedule blobs to handle situations where assets are not hosted on blob storage

## Changes
- check if blob exists on db before migrating
- add function to pull files from external URLs and upload to blob storage
- print statistical reports in markdown format

This closes #1590

### DIFF
--- a/backend/maint-scripts/update_schedule_blobs.py
+++ b/backend/maint-scripts/update_schedule_blobs.py
@@ -1,6 +1,10 @@
 #!/usr/bin/env python3
 
+# ruff: noqa: T201
+
+import argparse
 import hashlib
+import re
 from typing import Literal
 
 import requests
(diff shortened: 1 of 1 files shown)


elfkuzco:
> have a solution to store where the blob originally comes from ; if we do not do it right now, information is going to be lost / impossible to recover (see https://github.com/openzim/zimfarm/issues/1589)


Would a new db column `comments` be sufficient for this? Or should we reuse the schedule editor comments field and allow the user to add comments there.

benoit74:
> Would a new db column `comments` be sufficient for this? Or should we reuse the schedule editor comments field and allow the user to add comments there.

We should have one comment per blob, so this is indeed a different comment than the one of the schedule history. Maybe naming it `note` is in fact less misleading ... not sure

PR #1601, opened 2025-12-18
suppress query api exceptions

elfkuzco:
## Rationale
Move request call in `query_api` to really suppress exceptions regarding requests. See https://github.com/openzim/zimfarm/issues/1307#issuecomment-3670032092


benoit74 (approved):
LGTM

PR #1605, opened 2025-12-21
show selfish status of worker

### LINKED ISSUE #1603: Display in API and UI when worker is `selfish`
In DB we have a `selfish` column on workers which is very important in terms of scheduling.

This information must be:
- returned in the API (AFAIK it is not)
- displayed in the UI (just like the cordoned and disabled by admin badges, in both the worker list and worker details).

elfkuzco:
## Rationale
This PR returns information regarding selfish status of worker and displays them on the UI

Screenshot_20251221_165608

Screenshot_20251221_165542

## Changes

- return `selfish` status of worker from API
- display `selfish` status in UI

This closes #1603


benoit74:
Oh, we have a pyright issue to fix

PR #1611, opened 2025-12-30
Resolved the Issue by restoring recipe listing page and filters on browser back navigation

### LINKED ISSUE #1610: Recipe listing full black after moving back in browsing history
## Problem

When I move back in the navigation history, under certain cinditions, the full listing can be black (on my iPad)

## Reproducing steps

1. Filter recipes by category "stack_exchane"
2. 100 items per page selected
3. Move to page two
4. open stackoverflow recipe
5. Click on browser back button

## I get

A black listimg

Image

# I expect

Page 2 of the listing

skyy-07:
Fixes issue where recipe listing appeared black after navigating back in browser history. The page number and filters are now stored in URL query parameters and properly restored when navigating back.

Changes:
- Store page number in URL when pagination changes
- Restore page number and filters from URL on component mount
- Reload data when route query changes (back navigation)
- Sync v-data-table-server page state with paginator

Fixes #1610

## Rationale

## Changes

### DIFF
--- a/frontend-ui/src/components/BlobViewer.vue
+++ b/frontend-ui/src/components/BlobViewer.vue
@@ -1,0 +1,170 @@
+<template>
+  <v-btn
+    icon="mdi-eye"
+    size="x-small"
+    variant="text"
+    color="primary"
+    @click="handleView"
+    :title="`View ${kind === 'image' ? 'Image' : kind.toUpperCase()}`"
+    :loading="loadingBlobContent"
+  />
+
+  <!-- Image Viewer Dialog -->
+  <v-dialog v-if="kind === 'image'" v-model="showViewer" max-width="600">
+    <v-card>
+      <v-card-title class="d-flex justify-space-between align-center">
+        <span>View Image</span>
+        <v-btn icon="mdi-close" variant="text" @click="handleClose" />
+      </v-card-title>
+      <v-card-text>
+        <v-progress-circular v-if="loadingBlobContent" indeterminate />
+        <v-img
+          v-else-if="blobContent"
+          :src="blobContent"
+          max-height="70vh"
+          contain
+          alt="Blob image"
+        />
+        <div v-else class="text-error">Failed to load image</div>
+      </v-card-text>
+      <v-card-actions>
+        <v-spacer />
+        <v-btn color="primary" variant="text" @click="handleClose">Close</v-btn>
+      </v-card-actions>
+    </v-card>
+  </v-dialog>
+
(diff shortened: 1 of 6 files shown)


elfkuzco:
@skyy-07 , thanks for the PR and sorry for the delay in response. There's a couple of quirks with the fix even though it works as expected by updating the query parameters. You should try them out and understand better. The first is that when i go to a certain page,  select a recipe and hit the back button, there are duplicate requests to fetch the list of recipes. You should open your network tools to see this.

The second and more troubling one is that when i copy a URL with query parameters and paste it in a new private tab, I get repeated requests and the page becomes extremely flaky. 

The first has to do with Vuetify server side tables making the request internally and you duplicating it again. You probably want to hook into the update options to properly update the routes: https://vuetifyjs.com/en/api/v-data-table-server/#events-update:options

skyy-07:
Thanks for the feedback, Sir.
Here are the Key Changes that I have made:-
1. Decoupled Data Fetching from Table Events
I removed the direct loadData trigger from the @update:options handler.
The handler now only updates the route query via router.replace, and does not initiate any network requests itself.

This prevents double-fetching caused by v-data-table-server emitting internal updates while the app was also fetching manually.
2. Route-Driven Pagination and Fetching
I introduced a watcher on route.query in each component, and data fetching (loadData) is now triggered exclusively from this watcher.

This ensures:
Only one request is made per navigation state.
Back/forward navigation works predictably.
Reloading or opening a deep link in a new/private tab fetches data exactly once based on the URL.

3. Initial State Synchronization from the URL
currentPage and selectedLimit are now initialized directly from the route query when the component mounts.
This avoids the initial “flash” of default values (e.g. page 1), which previously caused unnecessary route updates and request loops when loading a URL with query parameters.
4. Infinite Loop Prevention
I added a guard in onUpdateOptions to compare the incoming table options with the existing route query.
The route is only updated when values actually change, which prevents circular updates and repeated request cycles.

Please review the code and tell me any further errors that I need to check.

elfkuzco:
Fix issues identified by the linter

elfkuzco:
@skyy-07 , sorry but I think you should ignore the PR. The rationale is that I have a solution that's a lot simpler than your current PR and it also extends to other views that use Vuetify's server side tables. I have take a look at #1612 and it seems to incorporate changes from this PR. Please, remove those and keep the changes constrained to the youtube links only.

Also, regarding the CI linter errors, you could simply install pre-commit and ensure the checks pass locally before raising a PR.

skyy-07:
Ok sir.

PR #1618, opened 2026-01-07
use custom color for navbar in dark mode

### LINKED ISSUE #1452: Implement dark theme
Which should be automatically chosen by the browser. No need to any configuration settings.

elfkuzco:
## Changes
- customize color of navbar in dark theme mode

Screenshot_20260107_143130

This closes #1452


elfkuzco:
@rgaudin , I only customized the navbar color because the color is too dull for text on black background. See the following image when I set it to be the application primary color. The name of the worker is almost invisible because the color is dull.

Screenshot_20260107_142402

PR #1622, opened 2026-01-09
Persist page number in route history

### LINKED ISSUE #1609: (recipe) pagination is not kept by navigating in the history
## Problem

When I move back in the navigation history, then the pagination of recipes is lost (I always come back on page 1)

## Reproducing steps

1. Filter recipes by category "stack_exchane"
2. Move to page two
3. open any recipe
4. Click on browser back button

## I get

Page 1 of the recipe listing displayed

# I expect

Page 2 of the listing

elfkuzco:
## Rationale
This PR persists the page number in history allowing the user to return to a previous page number

## Changes
- return current page number in pagination metadata (used by frontend `Paginator` model)
- decouple schedulesview paginator from the schedule store paginator
- modify vuetify's server-side tables to mutate query parameters instead of emitting event to load data
- adapt function to load filters from url to set correct paginator value
- adapt filter change events to reset paginator skip value to 0. This ensures that anytime we change a filter, we always get the entries starting from page 0
- load data in route watcher

This closes #1609

### DIFF
--- a/backend/src/zimfarm_backend/api/routes/platforms/logic.py
+++ b/backend/src/zimfarm_backend/api/routes/platforms/logic.py
@@ -2,7 +2,7 @@ from fastapi import APIRouter
 
 from zimfarm_backend.api.routes.models import ListResponse
 from zimfarm_backend.common.enums import Platform
-from zimfarm_backend.common.schemas.models import Paginator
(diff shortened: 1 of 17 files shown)


elfkuzco:
will extend this to other tables too - worker list, pipeline table and user table

elfkuzco:
seems selected limits aren't updated anymore. will fix.

Issue #1597, opened 2025-12-18
Continue to support external URLs, do not force blobs


benoit74:
I'm just realizing that there are some edge cases where we do not want to use a blob, but rather continue using an (external) URL. This is only for expert cases, so beginners should be biased towards using blobs rather than using an online URL, but we should have a mechanism in place (UI mostly) to still allow using any random URL.

Typical use cases:
- devdocs illustrations: these illustrations are populated by a script, and having to use blob which URLs are unknown is a pain
- yet-to-come mwoffliner article lists (see https://github.com/openzim/zimfarm/issues/1598) which are dynamic in most wikimedia recipes, but we still want to use blobs for "manually configured" recipes

API + UI should allow to still input and view these URLs which are not blobs

benoit74:
Edit: added link to https://github.com/openzim/zimfarm/issues/1598 regarding mwoffliner article lists

elfkuzco:
Does this not directly conflict with #1585 ? I mean, how do we discern when to create from URLs and when not to? Should we have a button on the input that the user would have to click to be able to edit a URL input? That is, by default, we don't edit URL blobs and let the backend deal with it as it is.

benoit74:
Yes it does conflict somehow, we need to find a compromise.

Maybe current approach of uploading the base64 value is not really the good one, and we should rather:
- have a new API to create / delete blobs
- let UI handle blob creation when the user wants to do so, with more advanced UI where user can input a URL to download, input a URL to use as-is - maybe just a toggle between "keep external URL as-is" or "store a copy on Zimfarm", directly drag-and-drop or input data, etc... ; maybe worth to have a popup in all situations 
- continue uploading blob URL just like any URL (as before) in the schedule endpoints

This would simplify logic for the backend, move a bit more responsibility to the frontend, but I feel like it is more the correct semantic we want. Blobs are not mandatory at all from a technical perspective. They are a net added value for usual users in usual use-cases. And the disymetry between base64 at upload and URL when downloading somehow proves there is something odd. The fact that I now realize there are use-cases where we want to only manipulate URLs shows it is probably wrong.

PR #1623, opened 2026-01-13
add support for zim illustration blobs

### LINKED ISSUE #1616: Support IllustrationBlob?
AFAIK, we have some fields in offliner definition that are `url` and those are rendered in the editor as Blob-accepting, meaning that if it's an image, there's the image-editor showing, allowing crop and all.

Most recipes expect an illustration which would be an URL with an image blob. In the case of Illustration, we have needs that go beyond it just being a blob: we want a **PNG** that is 48x48px.

Would it make sense to have a new field type that's maybe `illustration-url` that shows an improved version of the image editor that makes sure it's converted to PNG if not the source format and resized to 48x48 after the crop process?

Current situation assumes every scraper will either fail with incompatible input or properly convert/resize from input. I'm not sure what is preferred. That leads adding a policy about blobs: should those refer to Online source, Copies on our drive, should it be Source image or should it host resized versions instead?

I hope that has not been already answered in previous tickets

elfkuzco:
## Rationale
This PR adds support for blobs which are to be treated as illustrations

## Changes
- Add new kind `illustration` for blob types
- Use `zimscraperlib` functions to convert `image` kinds to `PNG` files and `illustrations` to width and height acceptable to scrapers
- Update Dockerfiles to install packages required for zimscraperlib
- Adapt UI logic to treat illustrations as images and let backend distinguish between them

This closes #1616


elfkuzco:
@rgaudin , I know we discussed forcing external URLs on illustrations to be converted to the specific size but I forgot to mention #1597 which mentions to not force blobs. As such, conversion is only done if a user makes an upload that is not from an external URL. When external URLs are converted are for cases when a user tries to update the comment or maybe modify the file pointed to by the external URL.

elfkuzco:
If this PR is satisfactory, I would need a list of the blobs that are illustrations. For now, I only added https://github.com/openzim/zimfarm/blob/133630aad68c65c7b6edef00cf3a53cdcc22ffa5/backend/maint-scripts/update_offliner_definition_blobs.py#L22

This would enable me to update the offliner definition json files across the different scraper repositories

rgaudin:
> @rgaudin , I know we discussed forcing external URLs on illustrations to be converted to the specific size but I forgot to mention #1597 which mentions to not force blobs. As such, conversion is only done if a user makes an upload that is not from an external URL. When external URLs are converted are for cases when a user tries to update the comment or maybe modify the file pointed to by the external URL.

What I suggested is that it stays external as long as it passes the requirement. So in my mind, when someone adds a URL:
- backend downloads URL
- checks wether it matches requirements (format, width, height)
- If it matches, it stays external
- if it doesn't, it's converted and a blob is created.

Given what it implies, I suggest we keep it as it is now and you open a ticket so that it can be discussed with @benoit74 when he comes back. That additional effor might not be necessary.

backend/src/zimfarm_backend/api/routes/blobs/logic.py
```diff
@@ -41,15 +46,28 @@
...
     "Create a blob for schedule"
 
+    if request.data.startswith("data:"):
+        _, encoded_data = request.data.split(",", 1)
+    else:
+        encoded_data = request.data
+    value = base64.b64decode(encoded_data)
+
+    if request.kind == "image":
+        blob_data = convert_image_to_png(value)
```
rgaudin, line 63: Is this wanted? I don't think it's been requested and should probably be backed by a ticket or something. TBH I don't know what's the use of those image blob beside illustrations…

elfkuzco: currently, all images are sent as base64 and there's no logic for detecting the extension. image files are just uploaded with an extension of `.png` even though they might not be actual PNG files. I figured it would be a good point to actually convert them to PNG files

elfkuzco: Created #1625 for it

PR #1628, opened 2026-01-14
adapt offliner definition blobs from image to illustration

elfkuzco:
## Rationale
This PR updates the script responsible for changing the blob type of offliners from `image` to `illustration`. The following offliner definition flags are changed from image to illustration.
| offliner     | flag             |
|--------------|------------------|
| freecodecamp | illustration     |
| ifixit       | icon             |
| kolibri      | favicon          |
| mindtouch    | illustration_url |
| nautilus     | favicon          |
| openedx      | favicon_url      |
| sotoki       | favicon          |
| zimit        | favicon          |

### DIFF
--- a/backend/maint-scripts/update_offliner_definition_blobs.py
+++ b/backend/maint-scripts/update_offliner_definition_blobs.py
@@ -15,21 +15,21 @@ FLAG_MAPPINGS: dict[
     str, list[dict[str, Literal["image", "css", "html", "illustration"]]]
 ] = {
     "devdocs": [{"logo_format": "image"}],
-    "freecodecamp": [{"illustration": "image"}],
-    "ifixit": [{"icon": "image"}],
-    "kolibri": [{"favicon": "image"}, {"css": "css"}, {"about": "html"}],
-    "mindtouch": [{"illustration_url": "image"}],
+    "freecodecamp": [{"illustration": "illustration"}],
+    "ifixit": [{"icon": "illustration"}],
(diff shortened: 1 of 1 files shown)


elfkuzco:
@rgaudin , i noticed there are some images that seem to require width and height constraints. For example, here is the youtube offliner definition file.
Screenshot_20260114_132742

Should we enforce the width and height constraints for normal images?

rgaudin:
Youtube profile and banner are not ZIM Metadata illustration. Maybe the profile is used to generate it though.

elfkuzco:
Okay. I meant, if we should extend the offliner definition spec to actually have width and height because the description implies it will be resized to these dimensions.

rgaudin:
> Okay. I meant, if we should extend the offliner definition spec to actually have width and height because the description implies it will be resized to these dimensions.

You can open a ticket ; we should probably list all those use cases first.

benoit74:
Waiting for https://github.com/openzim/zimfarm/issues/1680 to settle before merging, but it LGTM if we continue in this direction

benoit74 (approved):
Now that #1680 has settled in my mind, we should move this forward and merge. 

LGTM

Youtube scraper is indeed using profile image both in the UI and then resized to 48x48 for the ZIM illustration. Profile flag must hence still be `image` and not `illustration` to allow arbitrary image. I'm not sure current offliner flag description is 100% correct, but this is a scraper issue.

PR #1630, opened 2026-01-16
edit schedule notifications in ui

### LINKED ISSUE #575: Add notification config in UI
Zimfarm now has built-in notifications support but this is only available through the API. We should allow editors to edit the notifications from the UI.

How it works:

* notifications use 2 routes: `mailgun` (email), `webhook` (HTTP POST of the task's payload) and `slack`.
* notifications are triggered on events: `requested`, `started` and the virtual `ended` (for both `succeeded`, `failed` and `canceled`)
* there are global (for all tasks) notifications configurable via environment variables. Those don't need anything.
* notifications can be set on `schedules` and would thus be propagated to `requested_tasks` and `tasks`. It thus would be possible to add a notification to a running task and it would be triggered at the end. Not sure we need that for now.
* editing a schedule should thus allow setting the notifications.

Notifications are represented by a `notification` dict on schedules which can be validated there. It looks like this:

```json
{
    "notification": {
        "ended": {
            "mailgun": ["some@email.com", "another@one.fr"],
            "webhook": ["http://toto.com/api/hook", "https://kiwix.com/zimfarm"],
            "slack": ["#zimfarm-events", "@kelson"]
        }
    }
}
```

Please note that the `/task/{task_id}` endpoint only exposes the `notification` dict for authenticated requests from a user with `schedules.update` permission for privacy concerns.

elfkuzco:
## Rationale
This PR enhances the schedule editor to modify notification settings for schedules from the frontend UI
Screenshot_20260116_141922

## Changes
- create empty notification object for notifications instead of returning None when fetching schedule
- add notification to schedule history table and generate migration
- add component to edit notification fields for different events.

This closes #575


elfkuzco:
forgot to re-install precommit locally, hence the linter fix commits.

PR #1636, opened 2026-01-20
mark task as failed if zim or zimcheck upload fails

### LINKED ISSUE #1519: Better handling of issues around uploads / checks
Currently, we do not take into account ZIM upload status or check status to set the final task status.

Should we consider that a task with a ZIM which failed to upload or to check (and soon to upload the check result to S3) is failed (even if scraper completed)?

Should we also take into account other uploads (logs and artifacts) status?

From my perspective, any failure should make the task been marked as failed, but we should let the task continue to its end with other things to do to not loose data just because one thing failed.

@rgaudin @kelson42 WDYT?

elfkuzco:
## Rationale
This PR improves the shutdown logic of a scraper to mark a task as succeeded only if the zim and zimcheck results uploads are successful. Also, it enhances the healthcheck service to provide status on logs/artifacts upload for the most recently successful task.

## Changes
- enhance check for zim_files_actions to check if all the zims and zimchecks are uploaded/skipped
- add check to healthcheck service to determine if logs/artificats are uploaded. A log is considered uploaded if there is a `log` filename in the `container` key of the payload. For artifacts, an artifact is considered uploaded if the task has `artifcats_globs` set, an upload_uri and `artifacts` key in the `container` key of the payload.

This closes #1519

### DIFF
--- a/healthcheck/pyproject.toml
+++ b/healthcheck/pyproject.toml
@@ -211,6 +211,8 @@ minversion = "8.3.5"
 testpaths = ["tests"]
 pythonpath = [".", "src"]
 addopts = "--strict-markers"
+asyncio_mode = "auto"
+asyncio_default_fixture_loop_scope = "function"
 
 [tool.coverage.paths]
 healthcheck = ["src/healthcheck"]
--- a/healthcheck/src/healthcheck/router.py
+++ b/healthcheck/src/healthcheck/router.py
@@ -12,6 +12,7 @@ from healthcheck.status.database import (
     check_database_connection,
 )
 from healthcheck.status.frontend import check_frontend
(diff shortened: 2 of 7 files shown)


rgaudin (approved):
LGTM ; please confirm that `actions[ZIM_CHECK]` is how zimit handles the no-zimcheck in relation to:

> All this while keeping in mind that some of these things are optional based on (already existing) config vars (e.g. zimcheck is not running on all Zimfarms since we do not want to run it on zimit.kiwix.org).

elfkuzco:
I believe zimit should be configured with `ZIMCHECK_OPTION` to empty string. If it's an empty string, it will be set to SKIPPED by the worker.

PR #1637, opened 2026-01-20
improve support page view

### LINKED ISSUE #1632: Improve "Support" page
Image

Here my suggestions:
* ZIM files link should not point to the specification, but to library.kiwix.org
* The last link is broken
* I would recommend to present both ways to support in an equal manner (I don't understand why the financial support only is framed
* I would recommend to remov the hr node before last section... not so clear why this is here either.

elfkuzco:
## Rationale

Screenshot_20260120_123636

## Changes
- point `ZIM files` link to library.kiwix.org
- fix broken link to openZIM's repositories

This closes #1632


rgaudin (approved):
LGTM

PR #1639, opened 2026-01-22
prevent emitting save event when image is not edited

### LINKED ISSUE #1638: image editor creating new blobs despite no changes to the image
When a image that is already uploaded to the blob storage is viewed and the user clicks the "Done" button, it appears the request is sent to the backend to create a new blob. While the backend has a filter to prevent duplication of blobs if they have the same checksum, flag and kind type, it appears that the `Cropper` component is modifying the bytes in some way leading to a different checksum.

elfkuzco:
## Rationale
When a user clicks the `Done` button, the image editor originally emitted the image data using `canvas.toDataURL` but it seems this causes different arrangements of the bytes leading to #1638. This PR maintains a computed property to keep track of when a user actually made edits to the image.

## Changes
- remove warning about saving image. see  #1635 
- avoid emitting save event when the user hasn't modified the image as the `canvas.toDataURL` is creating different bytes arrangement of the file fetched from URL. see https://stackoverflow.com/questions/9773154/canvas-todataurl-increases-file-size-of-image
- ask for confirmation before saving edits.

Screenshot_20260122_135654

This fixes #1638

### DIFF
--- a/frontend-ui/src/components/ImageEditorDialog.vue
+++ b/frontend-ui/src/components/ImageEditorDialog.vue
@@ -1,13 +1,18 @@
 <template>
-  <v-dialog v-model="isOpen" max-width="900" persistent scrollable>
+  <v-dialog v-model="isOpen" max-width="600" persistent scrollable>
     <v-card>
-      <v-card-title class="text-h6 bg-primary d-flex align-center">
-        <v-icon class="mr-2">mdi-image-edit</v-icon>
-        Edit Image
+      <v-card-title class="text-h6 bg-primary">
+        <div class="d-flex align-center justify-space-between w-100">
+          <div class="d-flex align-center">
+            <v-icon class="mr-2">mdi-image-edit</v-icon>
+            Edit Image
+          </div>
+          <v-btn icon="mdi-close" variant="text" @click="handleCancel" size="small" />
+        </div>
+      </v-card-title>
 
-        <template v-if="imageLoaded">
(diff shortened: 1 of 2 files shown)


frontend-ui/src/components/ImageEditorDialog.vue
```diff
@@ -138,12 +131,30 @@
...
+    confirm-text="Save"
+    cancel-text="Cancel"
+    confirm-color="primary"
+    icon="mdi-content-save"
+    icon-color="primary"
+    @confirm="handleConfirmSave"
+    @cancel="handleCancelSave"
+  >
+    <template #content>
+      <p>This will save your changes and may create a new blob.</p>
```
rgaudin: I think @kelson42 reported in a different ticket that those information about blobs are superfluous so this part can be removed from the message.

rgaudin: Also, I don't understand the _may_ here. What does this mean? At this point, if we're asking whether to save, that's because we're confident there has been changes and thus blob would be created/updated. Or am I misinterpreting the behavior?
Is this for like if I flip twice, I have tainted `hasEdits` but I have the same image?

elfkuzco: Yes

rgaudin: OK. got it. please remove ` and may create a new blob` and merge

PR #1640, opened 2026-01-23
remove unused binaries for zimscraperlib

### LINKED ISSUE #1627: reduce the image size of the backend API
Adding `zimscraperlib` as a dependency of the backend API in #1623  meant installing OS-level packages with the following command.
```
sudo apt install libmagic1 wget ffmpeg \
    libtiff5-dev libjpeg8-dev libopenjp2-7-dev zlib1g-dev \
    libfreetype6-dev liblcms2-dev libwebp-dev tcl8.6-dev tk8.6-dev python3-tk \
    libharfbuzz-dev libfribidi-dev libxcb1-dev gifsicle
```

This has caused a significant increase in the size of the image from 318Mb to 1.53Gb. How can we optimize this to avoid such size blowup?

elfkuzco:
## Rationale
This PR replaces zimscraperlib library by copying over only the functionalities that have to deal with PNG image processing. As a consequence, api docker image has dropped to 601Mb from 1.53Gb. See #1627

## Changes
- use only functions from zimscraperlib that have to deal with image manipulation
- remove external binaries from dockerfile that aren't required to optimize/process images

This closes #1627

### DIFF
--- a/backend/api.Dockerfile
+++ b/backend/api.Dockerfile
@@ -21,11 +21,16 @@ ENV MAILGUN_FROM="Zimfarm <info@farm.openzim.org>"
 # multiple notification for same event separated with pipe `|`
 # ENV suffixed with event: `GLOBAL_NOTIFICATION_started` or `GLOBAL_NOTIFICATION_ended`
 # ENV GLOBAL_NOTIFICATION_ended slack,#zimfarm-events,@rgaudin|mailgun,reg@kiwix.org
-# curl needed for healthcheck
-RUN apt-get update && apt-get install -y curl libmagic1 wget ffmpeg \
+
+# - curl needed for healthcheck
+# - zimscraperlib: We only install the binaries needed to make zimscraperlib image manipulation work, not
+# the whole list as recommended on the installation page. This is because the installing all
+# the prerequisites binaries blow up the image size by over 1Gb and we don't need all those
+# functionalities.
+RUN apt-get update && apt-get install -y curl libmagic1 wget \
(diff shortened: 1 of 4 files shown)


elfkuzco:
oh, forgot to copy the tests from zimscraperlib too...

elfkuzco:
@rgaudin , do you think it's necessary to copy those tests too?

elfkuzco:
> If that's satisfactory, a comment on pyproject.toml would be mandatory as we would have a dependency listed that not 100% function (although we don't need it).

Shouldn't the comment be on the `api.Dockerfile` instead given that's where the binaries needed for zimscraperlib are installed? As far as pip knows, it has installed `zimscraperlib`.

elfkuzco:
Image size is now at 765Mb. Seems like an additional 165Mb from leaving zimscraperlib as dependency compared to not using it.

rgaudin:
> > If that's satisfactory, a comment on pyproject.toml would be mandatory as we would have a dependency listed that not 100% function (although we don't need it).
> 
> Shouldn't the comment be on the `api.Dockerfile` instead given that's where the binaries needed for zimscraperlib are installed? As far as pip knows, it has installed `zimscraperlib`.

Sure

rgaudin:
> Image size is now at 765Mb. Seems like an additional 165Mb from leaving zimscraperlib as dependency compared to not using it.

That seems enough as a fix for #1627. I'm more concerned that the rest is 600MB. Can be addressed later.

PR #1662, opened 2026-01-27
make worker table stat boxes have the same height

### LINKED ISSUE #1659: Not all worker boxes have the same size
The one on the left has a bigger size

Image

They should be all nicely aligned with the same size

elfkuzco:
Screenshot_20260127_064723

# Changes
- set v-row align prop to stretch so columns align to the tallest column
- set each v-sheet to flex-grow-1 so it expands to fill its column height

This fixes #1659


elfkuzco:
Never mind the skewed stats. My DB is filled with stale tasks and no workers and I don't run the background tasks to expire them.

Issue #1092, opened 2025-04-10
Sortable Headers for Data Tables


end-9214:
Current Pipeline Table doesn't have sortable header. 

![Image](https://github.com/user-attachments/assets/72d16a76-8b10-4260-9490-017cc01db9de)

![Image](https://github.com/user-attachments/assets/3438324f-e9d7-4daf-9086-65382f9a1a30)

Sortable headers - when user clicks on any column's name -> that column will get sorted in `asc` or `desc` order and the order will be indicated using arrow icons (arrow up `asc` arrow down `desc`).

I think sortable headers for this table will look cool, what do you think??

benoit74:
I agree sortable headers would be cool mostly everywhere. This is however quite a nice to have from my experience, rarely needed and significant work since besides UI changes, API also needs to be adapted.

kelson42:
The only input my side is that it has to be done in the backend, otherwise it will conflict with the pagination.

benoit74:
I still don't see much added value for the end-user and I still see lots of complication for the developers.

Still not in favor of implementing any of this

benoit74:
I propose to consider we will not have sortable headers for the time being, too little added value for too much pain.

PR #1664, opened 2026-01-27
sort workers by name only in ascending order

### LINKED ISSUE #1648: Difficult to find a worker in the worker list
We have quite a bit of workers and this is a bit difficult to find a specific one in the list.

Image

I don't know if or how they are storted now, but having them sorte alphabatically would be helpful IMHO.

In addition, having a filter working like the recipe name filter would be great.

elfkuzco:
## Rationale

This PR sorts workers list by name only and in ascending order. Previous sort was done in combination with last_seen which changes frequently and isn't particularly useful.

This closes #1648

### DIFF
--- a/backend/src/zimfarm_backend/db/worker.py
+++ b/backend/src/zimfarm_backend/db/worker.py
@@ -3,7 +3,7 @@ from ipaddress import IPv4Address, IPv6Address
 from typing import Any
 from uuid import UUID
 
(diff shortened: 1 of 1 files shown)


elfkuzco:
I want to create sub issues for sorting by other columns on #1092. This would extend to the other tables across the entire UI and would require quite some changes both in the routes update and backend. Is that okay?

PR #1671, opened 2026-01-29
use highlight.js to add syntax highlighting in text editor

### LINKED ISSUE #1656: add syntax highlighting for texts in blob editor

elfkuzco:
## Rationale
Screenshot_20260129_135731
Screenshot_20260129_135714
Screenshot_20260129_135653
Screenshot_20260129_135632

## Changes
- create CodeEditor component for reading/editing files (used while viewing config in BlobViewer and editing in BlobEditor respectively)
- use highlight.js to add syntax highlighting for text files
- use github theme for syntax highlighting

This closes #1656

### DIFF
--- a/frontend-ui/src/components/CodeEditor.vue
+++ b/frontend-ui/src/components/CodeEditor.vue
@@ -1,0 +1,176 @@
+<template>
+  <div class="code-editor-wrapper" :style="{ height: height }">
+    <highlightjs
+      :language="highlightLanguage"
+      :code="highlightedCode"
+      class="code-highlight-backdrop"
+    />
+    <textarea
+      v-model="internalValue"
+      class="code-textarea"
+      :class="{ 'readonly-textarea': readonly }"
+      :placeholder="placeholder"
+      :readonly="readonly"
+      spellcheck="false"
+      @scroll="handleScroll"
+      @input="handleInput"
+    ></textarea>
+  </div>
+</template>
+
+<script setup lang="ts">
+import type { TextKind } from '@/utils/blob'
+import { ref, computed, watch } from 'vue'
+
+interface Props {
+  modelValue: string
+  fileType?: TextKind
+  readonly?: boolean
+  height?: string
+  placeholder?: string
+}
+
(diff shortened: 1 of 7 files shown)


elfkuzco:
need to refine some things regarding the cursor.

frontend-ui/src/components/CodeEditor.vue
```diff
@@ -0,0 +1,201 @@
...
+    link.rel = 'stylesheet'
+    document.head.appendChild(link)
+  }
+  if (link.href !== href) {
+    link.href = href
+  }
+}
+
+watchEffect(() => {
+  const href = isDark.value
```
benoit74: Shouldn't we package this CSS in the app itself (via an NPM dependency)? I don't know if it's possible or not, but strategy so far has been to avoid external dependency on the Zimfarm if I'm not mistaken, so that we can work "in isolation". I'm also kinda worried we will not update version here when highlight.js will be updated

frontend-ui/src/main.ts
```diff
@@ -22,6 +29,13 @@ import router from '@/router'
 import App from '@/App.vue'
 import constants from '@/constants'
 
+// Register highlight.js languages
+hljs.registerLanguage('css', css)
+hljs.registerLanguage('javascript', javascript)
```
benoit74: Why registering `javascript` and `xml`, do we use it? (even if we could at some point, better to add them only once needed)

elfkuzco: xml is for html. will remove javascript

benoit74: Indeed, thank you

frontend-ui/src/components/CodeEditor.vue
elfkuzco: imported but it seems to be misbehaving for Dark mode with the colors different than the one on the CDN

elfkuzco: fixed following discussion on https://github.com/highlightjs/highlight.js/issues/3652

PR #1675, opened 2026-02-01
Periodic Refresh for task Debug logs

### LINKED ISSUE #1674: Autorefresh of task debug log
If I display the logs of a task

Image

They don't get refreshed, until I refresh the whole page.

We should better:
* Refresh periodically (once per minute) only the debug part
* Show a small beside "scraper stdout" when the refresh call happens
* If there is a problem keep displaying the latest version

rawadhossain:
## Rationale
Task debug logs are currently loaded only once and do not update while a task is running unless the user refreshes the page or switches tabs. Adding a periodic auto refresh to the `debug tab` improves usability while keeping the existing backend and page behavior unchanged.

## Changes
- Added 60s auto refresh for task debug logs when the `debug tab` is active
- Refreshes only the `debug section` without triggering the global loading overlay.
- Shows a small loading indicator beside `Scraper stdout` during refresh
- Keeps last fetched logs if a refresh fails.
- Starts polling only while the task is running and clears the interval on tab switch or component unmount.

This pr closes #1674

### DIFF
--- a/frontend-ui/src/views/TaskDetailView.vue
+++ b/frontend-ui/src/views/TaskDetailView.vue
@@ -479,6 +479,12 @@
                   <v-col cols="12" md="3">
                     <div class="text-subtitle-2">
                       Scraper stdout
+                      <v-progress-circular
+                        v-if="debugRefreshInProgress"
+                        indeterminate
+                        size="small"
+                        class="ml-1"
+                      />
                       <v-btn
                         size="small"
                         variant="outlined"
@@ -614,7 +620,7 @@ import {
   logsUrl,
 } from '@/utils/offliner'
 import { getTimestampStringForStatus } from '@/utils/timestamp'
-import { computed, inject, nextTick, onMounted, ref, watch } from 'vue'
+import { computed, inject, nextTick, onBeforeUnmount, onMounted, ref, watch } from 'vue'
 import { useDisplay } from 'vuetify'
 
 // Props
@@ -647,6 +653,8 @@ const task = ref<Task | null>(null)
 const error = ref<string | null>(null)
 const currentTab = ref(props.selectedTab)
 const flagsDefinition = ref<OfflinerDefinition[]>([])
+const debugRefreshInProgress = ref(false)
+const debugRefreshIntervalId = ref<ReturnType<typeof setInterval> | null>(null)
 
 const eventsHeaders = [
   { title: 'Event', value: 'code' },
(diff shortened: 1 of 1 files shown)


frontend-ui/src/views/TaskDetailView.vue
```diff
@@ -888,9 +912,32 @@ watch(
   () => props.selectedTab,
   (newTab) => {
     currentTab.value = newTab
-    refreshData()
+    if (newTab === 'details') {
```
elfkuzco: This has the effect of not refreshing the task when the user switches tabs. Isn't it possible to achieve the periodic refresh without this?

rawadhossain: Yes, you’re right. Changing that wasn’t necessary to achieve the periodic refresh. I’ll revert the tab watcher to its original form and keep the periodic debug refresh in addition to it.

frontend-ui/src/views/TaskDetailView.vue
```diff
@@ -869,6 +877,22 @@ const refreshData = async () => {
   loadingStore.stopLoading()
 }
 
+const refreshDebugOnly = async () => {
+  if (debugRefreshInProgress.value) return
+  debugRefreshInProgress.value = true
+  try {
+    const response = await tasksStore.fetchTask(props.id, !canViewTaskSecrets.value)
```
elfkuzco: This whole logic seems to be the same thing in the `refreshData` function. Could you try to re-use it?

rawadhossain: I initially thought limiting the tab watcher refresh to avoid unnecessary full reloads when switching to debug tab, since logs are now refreshed periodically.

Based on your feedback, I refactored the logic to reuse `refreshData`, by adding an optional `debugOnly` flag so existing behavior remains unchanged, while the periodic debug refresh can skip the loader and error handling and keep the last logs on failure.

```tsx
const refreshData = async (options?: { debugOnly?: boolean }) => {
  if (!options?.debugOnly) {
    loadingStore.startLoading('Fetching task...')
  }
  const response = await tasksStore.fetchTask(props.id, !canViewTaskSecrets.value)
  if (response) {
    task.value = response
    // Use nextTick to ensure the DOM has updated before scrolling
    nextTick(() => {
      scrollLogsToBottom()
    })
  } else {
    if (!options?.debugOnly) {
      error.value = 'Failed to fetch task'
      for (const error of tasksStore.errors) {
        notificationStore.showError(error)
      }
    }
  }
  if (!options?.debugOnly) {
    loadingStore.stopLoading()
  }
}

const refreshDebugOnly = async () => {
  if (debugRefreshInProgress.value) return
  debugRefreshInProgress.value = true
  try {
    await refreshData({ debugOnly: true })
  } finally {
    debugRefreshInProgress.value = false
  }
}
```
Does it seem reasonable to you now?

elfkuzco: Maybe, name the param `hideLoading` since you are moslty hiding notifications

rawadhossain: Sure, I’ll rename it.

elfkuzco:
can you squash the commits into one and re-request review. It looks good to me.

PR #1686, opened 2026-02-05
Do not follow redirects in API queries, URLs should be fine

### LINKED ISSUE #1679: POST request to /requested-tasks/ are failing with 401
Watcher was misconfigured and used `POST /requested-tasks/` instead of `POST /requested-tasks` to request tasks (this is fixed in https://github.com/openzim/zimfarm/pull/1678).

What is really disturbing is that from my tests and logs, the requests to `POST /requested-tasks/` were returning a 401 error which was very misleading.

I did not reproduced the issue in my dev machine yet, so it might be another problem, but worth to investigate.

benoit74:
Fix #1679


elfkuzco:
LGTM

PR #1698, opened 2026-02-16
compute task duration from started status in worker view

### LINKED ISSUE #1697: Task duration is not correct in worker overview
We had two long tasks running on `mwoffliner4` and I just have cancelled it.

Here is what Insee now:

Image

The "Task" column is wrong by two manners:
* Column header should not be "Task", but "Duration"
* The value given is not the whole task duration, but looks like to be the time elapsed since last event has been attached on the task.

elfkuzco:
## Rationale
This PR sets the duration of the task to be computed from the `started` status of a task instead of the current status.

## Changes
- compute timestamp from `started` status instead of current status of task in worker detail and worker list views
- rename column from `Task` to `Duration` in worker detail view

This closes #1697

### DIFF
--- a/frontend-ui/src/components/WorkersTable.vue
+++ b/frontend-ui/src/components/WorkersTable.vue
@@ -155,7 +155,7 @@
             <TaskLink
               :id="item.task.id"
               :updatedAt="item.task.updated_at"
-              :status="item.task.status"
+              :status="'started'"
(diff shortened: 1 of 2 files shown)


benoit74:
Closing because I've realized this is not what we want: https://github.com/openzim/zimfarm/issues/1697#issuecomment-4183174329

PR #1703, opened 2026-02-17
store illustrations sizes directly in metadata

elfkuzco:
## Rationale
Currently, the task worker skips the illustration sizes in teh `zim.metadata_keys` of an `Archive` and computes a formatted one of the form `Illustration_{size}x{size}`. However, the cms expects key `Illustration_48x48@1` (which exists in the zim.metadata_key). This PR restores setting the keys as they are without any formatting of the key name.

## Changes
- detect the size of the illustration from the illustration key and extract the content for that size
- use the illustration key directly in the metadata

### DIFF
--- a/worker/src/zimfarm_worker/task/zim.py
+++ b/worker/src/zimfarm_worker/task/zim.py
@@ -25,7 +25,7 @@ def get_zim_info(fpath: pathlib.Path) -> dict[str, Any]:
     for size in zim.get_illustration_sizes():
         payload["metadata"].update(
             {
-                f"Illustration_{size}x{size}": base64.standard_b64encode(
+                f"Illustration_{size}x{size}@1": base64.standard_b64encode(
                     zim.get_illustration_item(size).content
                 ).decode("ASCII")
             }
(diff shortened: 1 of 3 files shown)


worker/src/zimfarm_worker/task/zim.py
```diff
@@ -22,11 +32,14 @@ def get_zim_info(fpath: pathlib.Path) -> dict[str, Any]:
             if not key.startswith("Illustration_")
         },
     }
-    for size in zim.get_illustration_sizes():
+    illustration_sizes: list[str] = [
```
benoit74: I would prefer to keep old code and simply replace f"Illustration_{size}x{size}" with `f"Illustration_{size}x{size}@1"`, since we know we only have `@1` dpr for now

New code is less readable and as suspect to break should we have `@2` dpr at some point

elfkuzco: restored

elfkuzco:
added maintenance script to update existing illustration metadata for existing files

benoit74:
FYI, I ran the maintenance script on production DB, LGTM

PR #1705, opened 2026-02-17
avoid removing canceling task from list of known keys

elfkuzco:
## Rationale

This PR improves the logic for exiting gracefully when cancellation is requested by stopping the scraper before uploading it's logs and artifacts. It also avoids marking a task that is in CANCEL_REQUESTED from having it's containers and directories removed by the logic to cleanup leftovers

## Changes
- rename function to cancel_and_remove_task to cancel_task with option for removal (default to True)
- stop scraper before uploading its logs and artifacts


benoit74 (approved):
LGTM, nice catch!

PR #1707, opened 2026-02-18
generate and use access token before advertising book to cms

### LINKED ISSUE #1677: Authenticate to Ory.sh before notifying the CMS
In order to notify the CMS, the Zimfarm needs to be authenticated (so that not anybody can send notifications to the CMS).

For this, we will use Ory.sh :
- the Zimfarm will be an OAuth client in Ory, and hence have a `client_id` + a `client_secret`
- the Zimfarm will exchange this `client_id` and `client_secret` for an `access_token` (via the client credentials grant)
- the Zimfarm will send the `access_token` in each query to the CMS
- the Zimfarm will renew the `access_token` (sending again `client_id` and `client_secret`) only when access token is close to expire (5 minutes)

Environment variables allow to pass client id, client secret and renewal time limit.

elfkuzco:
## Rationale
This PR enhances the logic that advertises book to cms to generate and use access tokens as authentication mechanism.
It also removes unwanted fields sent as part of the payload to the CMS endpoint

## Changes
- add env vars for oauth2 client token generation
- add functionality to generate access token and cache expires_time
- move logic for advertising book to cms to background task where it's used.
- remove unnecessary params in payload. previous code added `period` (retrieved from `metadata["Date"]`) and download_url. These fields are now computed/retrieved by the CMS 

This closes #1677

### DIFF
--- a/backend/src/zimfarm_backend/background_tasks/constants.py
+++ b/backend/src/zimfarm_backend/background_tasks/constants.py
@@ -8,7 +8,7 @@ from zimfarm_backend.common.constants import getenv, parse_bool
 HISTORY_TASK_PER_SCHEDULE = int(getenv("HISTORY_TASK_PER_SCHEDULE", default=10))
 
 # Stalled task timeouts
-STALLED_GONE_TIMEOUT = parse_timespan(getenv("STALLED_GONE_TIMEOUT", default="1h"))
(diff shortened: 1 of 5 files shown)


elfkuzco:
added the cms audience id to the data sent while generating token

PR #1713, opened 2026-02-23
translate oauth2 callback error messages

### LINKED ISSUE #1712: Ory session auth: better handle case where user is authenticated but without sufficient factors
This issue is about the situation where we have configured the API to require two factors.

When the user does not have two factors, this is now correctly handled and we get the error advising the user to configure his account with a second factor.

Assuming the user does what he is required to do (configure a second factor), when he come back to the Zimfarm UI and tries to sign-in with Kiwix again, he is presented an authentication error:

Image

The problem seems to be that the call to https://login.kiwix.org/sessions/whoami?tokenize_as=kiwix-tokenizer returns following payload which is not handled correctly by the UI:

```
{
    "error": {
        "id": "session_aal2_required",
        "code": 403,
        "status": "Forbidden",
        "reason": "An active session was found but it does not fulfill the requested Authenticator Assurance Level. Please verify yourself with a second factor to resolve this issue.",
        "details": {
            "redirect_browser_to": "https://login.kiwix.org/self-service/login/browser?aal=aal2"
        },
        "message": "Session does not fulfill the requested Authenticator Assurance Level"
    },
    "redirect_browser_to": "https://login.kiwix.org/self-service/login/browser?aal=aal2"
}
```

Following the `redirect_browser_to` URL worked fine, except that it seems to miss the `return_to=https://farm.openzim.org/oauth/callback` parameter.

elfkuzco:
## Rationale
This PR attempts to unify the error message sent by both the zimfarm and Ory APIs to a list of strings that can be shown on the UI. This way, messages can be used by the views that require them (e,g Callback view)  knowing they are a list of strings and not really caring where they are coming from

## Changes
- enhance logic for translating errors to include oauth2 error responses too

This closes #1712

### DIFF
--- a/frontend-ui/src/types/errors.ts
+++ b/frontend-ui/src/types/errors.ts
@@ -4,8 +4,17 @@ export interface ErrorResponse {
   errors?: Record<string, string>
 }
 
+export interface OAuth2ErrorRecord {
+  id: string
+  code: number
+  status: string
+  reason: string
+  details: Record<string, string>
+  message: string
+}
+
 export interface OAuth2ErrorResponse {
   code?: number
-  error: string
+  error: string | OAuth2ErrorRecord
   error_description?: string
 }
--- a/frontend-ui/src/utils/errors.ts
+++ b/frontend-ui/src/utils/errors.ts
@@ -1,9 +1,39 @@
-import { type ErrorResponse } from '@/types/errors'
+import { type ErrorResponse, type OAuth2ErrorResponse } from '@/types/errors'
 
-export function translateErrors(error: ErrorResponse) {
-  // build up a list of the errors
-  const errors = []
-  if (error.errors) {
+export function translateErrors(error: ErrorResponse | OAuth2ErrorResponse) {
+  // Handle OAuth2ErrorResponse
+  if ('error' in error) {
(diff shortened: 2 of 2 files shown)


elfkuzco:
Toggling 2FA on and off didn't help to reproduce the scenario in #1712. So, I figured handling the error and translating them as messages would be ideal.

benoit74:
I'm not opposed to merge this since it could help for situations we do not correctly handle. But this is not a fix for #1712, we need to handle this situation correctly to better guide the user.

I did not try this on my own, but I think you can reproduce the error in #1712 with:
- remove 2FA from your Ory account
- logout from Zimfarm/Ory
- login again in Zimfarm => Zimfarm will complain you do not have 2FA activated, but you will have an Ory session (if not, login directly to Ory to get the session, but I think you do have the session in Ory, just not in Zimfarm)
- add 2FA to your account
- login again in Zimfarm => you should get the error I got

Is this exactly what you tried? If not, I will try to experiment to get the error again

elfkuzco:
> Is this exactly what you tried? 

Yup. And after enabling the 2FA app, I come back to zimfarm, click on sign in and it works. Didn't get the error payload.

Issue #1704, opened 2026-02-17
Why gwern.net_en_all keeps getting cancelled?


benoit74:
See https://github.com/openzim/zim-requests/issues/1541#issuecomment-3914713322

@elfkuzco can you have a look? At first glance, it looks like we have very big logs written by the crawler, which in turn might cause some timeouts somewhere.

benoit74:
I also don't get why we still don't have log file + task worker logs uploaded. I thought it was supposed to be fixed with https://github.com/openzim/zimfarm/pull/1670 and https://github.com/openzim/zimfarm/pull/1673

elfkuzco:
payload to retrieve task details is around `21Mb`. could be there are API timeouts. Is it possible to get the API logs around this period?

benoit74:
Sure: 

[Logs-2026-02-17 16_40_56.txt](https://github.com/user-attachments/files/25368779/Logs-2026-02-17.16_40_56.txt)

elfkuzco:
issue seems to be that the worker manager is cleaning the task files before the task gracefully shuts down. didn't notice this in dev as logs were too small and got uploaded quickly. fixing...

elfkuzco:
https://github.com/openzim/zimfarm/blob/e544f86fd6bbe67b4e9c369199a7000446b1cd79/worker/src/zimfarm_worker/task/worker.py#L806-L829

@benoit74 , do we need to keep a limit on how many lines to stream because this would stream till the end and given there's a limit to how long the upload takes, it would fail if this takes too long. As per [the docker docs](https://docs.docker.com/reference/cli/docker/container/stop/#timeout), a `SIGKILL` is sent to forcefully kill the container if it doesn't exit gracefully within the timeout.

benoit74:
We need all logs, and so far we never had issues regarding log uploads. I remember gigantic logs like ~ 500M files which worked fine. What changed here?

On another hand, I feel like the logs which are sent "on-the-fly" to the API for immediate display are indeed sometimes way too big and should be trimmed. What is ensuring we do not fill the DB with these "on-the-fly" logs?

elfkuzco:
With #1705 merged, is is it possible to re-run the recipe so we can see the logs and understand why?

benoit74:
Sure, just started

Popolechien:
Issue is still there: https://farm.openzim.org/pipeline/2afe0f6c-4cbb-4b62-b645-60c094e41cee
[...]

PR #1714, opened 2026-02-24
limit the size of scraper logs submitted on the fly

elfkuzco:
## Rationale
This PR aims to restrict the size of scraper logs generated and submitted to the API as the requests appear to be rejected by load balancer and don't reach the API, thereby making it seem like the task is stale and inadvertently getting canceled by the periodic task (see #1704). For now, the size is capped at 1million charactrs both for the scraper stderr and scraper stdout output. Since these logs are eventually uploaded, the log attribute of the task will be more of a mirror of the most recent logs instead of the full logs (which can exceed the API limit)

## Changes
- cap scraper stdout and stderr output sent to the API at 1 milion characteres

### DIFF
--- a/worker/src/zimfarm_worker/common/requests.py
+++ b/worker/src/zimfarm_worker/common/requests.py
@@ -78,7 +78,7 @@ def query_api(
         return Response(
             status_code=resp.status_code,
             success=resp.ok,
-            json=resp.json() if resp.text else {},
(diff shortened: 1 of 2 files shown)


elfkuzco:
I'm not going to attach #1704 to the PR for now until the recipe is re-run to completion.

PR #1715, opened 2026-02-24
Add new maps offliner

benoit74:
- Add new `maps` offliner which we need to support in the Zimfarm
- Since I was here, remove `psiram` warehouse path which does not exists and we don't intend to use


benoit74:
This is probably a good PR to track what still needs to be enhanced to achieve the ultimate goal of not having to touch the codebase everytime we deploy a new scraper 🤣 

Still way better than it used to be

PR #1721, opened 2026-02-26
drop email fields

### LINKED ISSUE #1720: Drop user `email` column
The `email` column is a PII and should hence be avoided. AFAIK, it was meant as a convenience to contact worker owners and as a requirement for a potential future self-reset password flow (never implemented).

@elfkuzco can you check if there any real usage of this data in the codebase? Could WP1 or zimit-frontend use this for anything?

If not we need to drop it from UI, API and DB.

elfkuzco:
## Changes
- drop email fields across the codebase and DB

This closes #1720


benoit74 (approved):
LGTM

PR #1722, opened 2026-02-26
add display_name field to users and use user_id as unique identifier

### LINKED ISSUE #1689: Clean usage of `username` and `displayname`
When user arrives with a JWT and the `name` inside the JWT is different from the one stored in DB (and not empty / missing), we should update the `username` in DB so that users can update their username directly in Ory and so that we ensure a consistent data across all our applications.

elfkuzco:
## Rationale
This PR refactors the API/UI to use both the username and id to perform actions on user views. Since Ory integration, users can have names which are similar to existing users (registered locally) and this would result in a unique constraint violation. To mitigate this, a new column `display_name` is added to the DB and API responses. This field is not unique and serves as a place to store the names coming from Ory and the locally registered users. For local users, the display name is generally the same as the username.

## Changes
- add display_name field to dB and API responses
- allow path parameters to take in either user_id or username (since both are unique)
- refactor UI calls to make requests with user_id instead of username
- refactor logic that prevents unprivileged users from modifying other users into a reusable dependency

This closes #1689


elfkuzco:
Pending elaboration on https://github.com/openzim/zimfarm/issues/1689#issuecomment-3967397290, this seems to align with much of #1689

backend/src/zimfarm_backend/api/routes/dependencies.py
```diff
@@ -61,18 +64,28 @@ def _get_current_user_or_none(
...
                 role=RoleEnum.VIEWER,
                 idp_sub=claims.sub,
             )
-            user = get_user_by_id_or_none(session, user_id=claims.sub)
+            user = get_user_by_id_or_none(
+                session, user_id=claims.sub, fetch_ssh_keys=True
+            )
+
+        # if token contains a "name" attribute and display_name is different, update it
+        if user and claims.name and user.display_name != user.display_name:
```
benoit74: `user.display_name != claims.name`?

elfkuzco:
> Before merging, can you double check we will not break any "running" thing due to API change (besides the frontend which will have to be updated as soon as API is updated of course, I'm more thinking about workers and tasks). I don't think so, but I prefer to ask ^

Yup. Nothing should break. There's no logic to change token decoding or key authentication with SSH (so long as workers have username since that's used in generating the auth message). Tested locally with a new task and everything works fine.

PR #1731, opened 2026-03-05
fix: grey bg not set at the right level

### LINKED ISSUE #1643: Grey Background not set at the right level
If you need one (I guess this might be a subject of discussion) it should be not on the whole block, but only behind each tile:

Image

The liner between the pagination and the tiles is not necessary

Pranavjeet-Naidu:
Fixes #1643

## Changes
The only file modified is PipelineTable.vue, with all the changes being a pure css modification. 
Instead of sticking with the table's default styling, now its overridden by scoped css that removes the grey block and adds it behind each of the rows. One more change is removal of the separator line between the rows and pagination. 

### DIFF
--- a/frontend-ui/src/components/PipelineTable.vue
+++ b/frontend-ui/src/components/PipelineTable.vue
@@ -1,6 +1,6 @@
 <template>
   <div>
-    <v-card v-if="!errors.length" :class="{ loading: loading }" flat>
+    <v-card v-if="!errors.length" :class="{ loading: loading }" flat color="transparent">
       <!-- Load All Last Runs button - only show in failed tab -->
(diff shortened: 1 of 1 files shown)


Pranavjeet-Naidu:
> WDYT about copilot remarks? Not sure they are 100% relevant but I don't feel like they are stupid, waiting for your feedback.

They were actually relevant but have been addressed in the latest commit: 

- Edge gap - Already fixed with negative margins
- Mobile spacing - Already handled in data-table.css, lines 11~16 have rule for the margin of 4px, it also has !important

Pranavjeet-Naidu:
The commits have been squashed into one and pushed as well :)

PR #1741, opened 2026-03-10
Cms healthcheck

### LINKED ISSUE #1459: healthcheck for CMS availability
As discussed in #557 , the healthcheck service should monitor if the CMS is available. This would require hitting an endpoint on the CMS service to determine if everything is healthy. @benoit74 , does the CMS endpoint have a route for viewing all it's endpoints? If so, which endpoint should we hit for this check?

elfkuzco:
## Rationale
This PR adds a healthcheck to the zimfarm to check if CMS is available and if there are files that have not been notified past a given a threshold since they were created.
Screenshot_20260310_090524

## Changes
- share logic to get list of files to notify and expose endpoint to view list of files
- add healthcheck components for cms availabilty and list of pending files for which notifications haven't been sent.

This closes #1459

### DIFF
--- a/healthcheck/src/healthcheck/constants.py
+++ b/healthcheck/src/healthcheck/constants.py
@@ -23,13 +23,15 @@ DEBUG = parse_bool(getenv("DEBUG", default="false"))
 
 REQUESTS_TIMEOUT = parse_timespan(getenv("REQUESTS_TIMEOUT", default="1m"))
 
-ZIMFARM_API_URL = getenv("ZIMFARM_API_URL", mandatory=True)
-ZIMFARM_FRONTEND_URL = getenv("ZIMFARM_FRONTEND_URL", mandatory=True)
+ZIMFARM_API_URL = getenv("ZIMFARM_API_URL", default="https://api.farm.openzim.org/v2")
+ZIMFARM_FRONTEND_URL = getenv(
+    "ZIMFARM_FRONTEND_URL", default="https://farm.openzim.org"
+)
 ZIMFARM_USERNAME = getenv("ZIMFARM_USERNAME", mandatory=True)
 ZIMFARM_PASSWORD = getenv("ZIMFARM_PASSWORD", mandatory=True)
 ZIMFARM_DATABASE_URL = getenv("ZIMFARM_DATABASE_URL", mandatory=True)
 
-CMS_API_URL = getenv("CMS_API_URL", mandatory=True)
+CMS_API_URL = getenv("CMS_API_URL", default="https://api.cms.openzim.org/v1")
 CMS_ENABLED = parse_bool(getenv("CMS_ENABLED", default="false"))
 CMS_PENDING_THRESHOLD = datetime.timedelta(
     seconds=parse_timespan(getenv("CMS_PENDING_THRESHOLD", default="24h"))


.github/workflows/healthcheck-Tests.yml
```diff
@@ -30,18 +30,27 @@ jobs:
       - name: Run healthcheck container
         run: |
           docker run -d --name zimfarm-healthcheck-test \
+            --network host \
             -e ZIMFARM_API_URL=http://localhost:8001 \
+            -e ZIMFARM_FRONTEND_URL=http://localhost:8001 \
```
benoit74, line 35: We should use better dev values. Zimfarm frontend URL is de-facto wrong, and regarding CMS I would prefer we use prod CMS if we don't have a local dev instance, here it is really too misleading.

healthcheck/src/healthcheck/constants.py
```diff
@@ -27,3 +28,9 @@ def parse_bool(value: Any) -> bool:
 ZIMFARM_USERNAME = getenv("ZIMFARM_USERNAME", mandatory=True)
 ZIMFARM_PASSWORD = getenv("ZIMFARM_PASSWORD", mandatory=True)
 ZIMFARM_DATABASE_URL = getenv("ZIMFARM_DATABASE_URL", mandatory=True)
+
+CMS_API_URL = getenv("CMS_API_URL", mandatory=True)
+CMS_ENABLED = parse_bool(getenv("CMS_ENABLED", default="false"))
```
benoit74, line 35: `CMS_ENABLED` is unused.

I think we should:
- either not check for CMS status at all when CMS is not enabled
- or (my preference) not fail the whole healthcheck status when CMS checks fails but CMS is not enabled

In both cases, we need to indicate in the UI that CMS checks status are ignored when `CMS_ENABLED` is not set.

healthcheck/src/healthcheck/status/cms.py
```diff
@@ -0,0 +1,107 @@
...
+    else:
+        data = CmsHealthCheck.model_validate(response.json)
+    return Result(
+        success=response.success,
+        status_code=response.status_code,
+        data=data,
+    )
+
+
+async def check_cms_pending_notifications() -> Result[PendingNotificationsStatus]:
```
benoit74, line 51: This check should ignore the `CMS_MAXIMUM_RETRY_INTERVAL`, we want to check there is no pending notifications at all. The `CMS_MAXIMUM_RETRY_INTERVAL` is meant to avoid retrying "forever" to send a notification which is not working, but it must still be considered a problem in terms of monitoring.

healthcheck/src/healthcheck/status/cms.py
```diff
@@ -0,0 +1,107 @@
...
+    """Check if there are files pending CMS notifications for too long."""
+
+    check_name = "cms-pending-notifications"
+
+    response = await query_api(
+        f"{ZIMFARM_API_URL}/files/pending-cms-notifications",
```
benoit74, line 57: This implementation seems risky, should the API return hundreds of objects (if CMS has a big issue for instance) then we might get a very large response, while we only need to know if we have any pending notification which is older than `CMS_PENDING_THRESHOLD`.

Should we sort results by oldest first (would be quite logical) and implement pagination?

Implementing pagination would be handy for the background task as well (so that we do not overwhelm it with too many objects in memory should we have too many notifications to process).

.github/workflows/healthcheck-Tests.yml
elfkuzco: The test more or less visits the `/doc` endpoint to check the API can really start. Not that the healthchecks pass or can run because if any fails, status is 503.

elfkuzco:
Limited the results and kept API pretty much the same.

.github/workflows/healthcheck-Tests.yml
elfkuzco: This is more of a test to see the API is actually running. Not that the healthchecks are passing. Would need to have env vars for zimfarm username and password too for authentication. Any check that fails means status code will be 503 and the job would fail because it checks status code is 200

elfkuzco:
Screenshot_20260310_152852

.github/workflows/healthcheck-Tests.yml
benoit74: Oh sorry, got it, I assumed we were in dev docker-compose ... my bad.

PR #1743, opened 2026-03-12
Silent pydantic serialization warnings in models build without validation

### LINKED ISSUE #1541: Strange issue around `AnyUrl`
I saw that while using the `update_scraper_version.py` ; looks like it is only a warning, but I don't get why we get this warning, worth to remove IMHO (surprisingly it looks like we do not have same warnings in the API backend logs ...)

```
[2025-11-25 09:54:18,795: INFO] setting freecodecamp image tag for schedule freecodecamp_en_coding-interview-prep to 2.0.3...
/usr/local/lib/python3.13/site-packages/pydantic/main.py:463: UserWarning: Pydantic serializer warnings:
  PydanticSerializationUnexpectedValue(Expected `<class 'pydantic.networks.AnyUrl'>` but got `<class 'str'>` with value `'https://github.com/openzim/freecodecamp/raw/refs/heads/main/zimui/src/assets/icon_coding_interview.svg'` - serialized value may not be as expected.)
  return self.__pydantic_serializer__.to_python(
/usr/local/lib/python3.13/site-packages/pydantic/functional_serializers.py:446: UserWarning: Pydantic serializer warnings:
  PydanticSerializationUnexpectedValue(Expected `<class 'pydantic.networks.AnyUrl'>` but got `<class 'str'>` with value `'https://github.com/openzim/freecodecamp/raw/refs/heads/main/zimui/src/assets/icon_coding_interview.svg'` - serialized value may not be as expected.)
  lambda x, h: h(x), schema=core_schema.any_schema()
```

benoit74:
Fix #1541

When pydantic models are build without validation, we use the `model_construct` method to skip any validation. This is mandatory for data "read" cases of offliner config which might not "yet" match expected schema.

The downside was that when serializing the model built with `model_construct`, pydantic is raising warnings because some fields are not of expected type. This is expected because we've build the model with `model_construct` which uses only bare Python types.

This PR enhance the situation by suppressing pydantic warnings:
- only for models which are known to support building without validation
- and only when these models have been built skipping validation

This is expected to remove useless "expected" logs while keeping any other unexpected serialization warning logged.

### DIFF
--- a/backend/maint-scripts/update_scraper_version.py
+++ b/backend/maint-scripts/update_scraper_version.py
@@ -56,7 +56,10 @@ from sqlalchemy.orm import Session as OrmSession
 from sqlalchemy.orm.attributes import flag_modified
 
 from zimfarm_backend import logger
-from zimfarm_backend.common.schemas.models import ScheduleConfigSchema
+from zimfarm_backend.common.schemas.models import (
+    DockerImageSchema,
+    ScheduleConfigSchema,
+)
 from zimfarm_backend.common.schemas.orms import OfflinerDefinitionSchema, OfflinerSchema
 from zimfarm_backend.db import Session
 from zimfarm_backend.db.models import RequestedTask, Schedule
@@ -122,8 +125,8 @@ def update_schedules(
                 f"setting {offliner.id} image tag for schedule {obj.name} "
                 f"to {args.image_tag}..."
             )
-            schedule_config = schedule_config.model_copy(
-                update={"image": {"name": schedule_config.image.name, "tag": image_tag}}
+            schedule_config.image = DockerImageSchema(
+                name=schedule_config.image.name, tag=image_tag
             )
 
         if offliner_definition.id != schedule.offliner_definition_id:


elfkuzco:
When i exec into the container and run `./update_scraper_version.py -o zimit -t 2.1.0`, the warnings still show but this time it is:
```sh
[2026-03-12 12:42:03,544: INFO] setting zimit image tag for schedule nixos.wiki_en_all to 2.1.0...
/usr/local/lib/python3.13/site-packages/pydantic/main.py:463: UserWarning: Pydantic serializer warnings:
  PydanticSerializationUnexpectedValue(Expected `DockerImageSchema` - serialized value may not be as expected [input_value={'name': <DockerImageName...zimit'>, 'tag': '2.1.0'}, input_type=dict])
  return self.__pydantic_serializer__.to_python(
```

benoit74:
I get it as well. But I wonder if this is not a real warning which should be fixed and which was previously masked by other warnings. Because this PR does not touch the DockerImageSchema or any class using it.

benoit74:
Yes this is a real problem. DB contains `ghcr.io/openzim/zimit` which is not a DockerImageName.

elfkuzco:
Okay. Culprit seems to be from this line: https://github.com/openzim/zimfarm/blob/a5f722185d4384311442e6126530ce648b847be3/backend/maint-scripts/update_scraper_version.py#L120-L127


Changing it to this works and also fixes the DB entry
```py
            schedule_config.image = DockerImageSchema(
                name=schedule_config.image.name, tag=image_tag
            )

```

elfkuzco:
Should I apply this fix onto this PR or new PR?

benoit74:
@elfkuzco you can do this in this same PR, it is ok

benoit74 (commented):
Don't we have the same issue with `new_schedule_config.model_copy(update=update_data)` in schedule update code of the API, where we also update the image? Did you checked that?

elfkuzco:
Yes. But that works as expected because FastAPI transforms those correctly. The `DockerImageSchema` does some pre-processing that removes the `ghcr.io/` prefix and so, it doesn't complain.

PR #1744, opened 2026-03-12
retrieve zim donwload/view urls from cms

### LINKED ISSUE #1439: CMS - retrieve file view and download links from the CMS
Once https://github.com/openzim/zimfarm/issues/1438 is implemented and if the Zimfarm operates with a CMS, the Zimfarm will not know where the ZIM files can be downloaded from and in which library they can be viewed.

The CMS will provide an API endpoint to be able to query for file locations. A list of ZIM UUIDs will be passed, and the CMS will return a list of links with their types (`download` or `view`).

Zimfarm API / UI should be adapted to:
- display multiple links per ZIM in the UI (currently UI shows only the `download` link which is not totally appropriate, see https://github.com/openzim/zimfarm/issues/704)
- query CMS (when plugged) for this links in the API (probably only `GET /tasks/{task_id}` only) ; it is important this still returns proper values when no CMS is used in association with the given Zimfarm

elfkuzco:
## Rationale
This PR enhances the API to retrieve download/view links from the CMS while retrieving a task.
Screenshot_20260312_121013

## Changes
- retrieve cms download/view links if `INFORM_CMS` is enabled while fetching a task
- update UI to show download/view links for zim files. fallback to show the original download url if no urls found

This closes #1439

### DIFF
--- a/backend/src/zimfarm_backend/common/upload.py
+++ b/backend/src/zimfarm_backend/common/upload.py
@@ -1,9 +1,20 @@
 import pathlib
 import urllib.parse
 
+import requests
+
 from zimfarm_backend import logger
-from zimfarm_backend.common.constants import SECRET_STRING_LENGTH
-from zimfarm_backend.common.schemas.orms import RequestedTaskFullSchema, TaskFullSchema
+from zimfarm_backend.common.constants import (
+    CMS_BASE_URL,
+    REQ_TIMEOUT_CMS,
+    SECRET_STRING_LENGTH,
+)
+from zimfarm_backend.common.schemas.orms import (
+    RequestedTaskFullSchema,
+    TaskFileSchema,
+    TaskFullSchema,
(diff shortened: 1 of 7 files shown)


backend/src/zimfarm_backend/common/upload.py
```diff
@@ -126,3 +137,28 @@ def generate_http_upload_url(uri: str, filename: str) -> str:
...
+
+    try:
+        response = requests.get(
+            f"{CMS_BASE_URL}/books/zims",
+            timeout=REQ_TIMEOUT_CMS,
+            params={"zim_ids": list(zim_ids.keys())},
+        )
+        response.raise_for_status()
+    except Exception:
+        logger.exception("An unexepceted error occurred while fetching ZIM URLs")
```
benoit74: typo in `unexepceted` + add `from CMS` at the end so that it is a bit more obvious

PR #1758, opened 2026-03-17
Ignore warehouse path

### LINKED ISSUE #1438: CMS - Transfer ownership of ZIM location to the CMS
Currently, Zimfarm editors have to specify the warehouse path in the recipe.

With this issue, we will transfer this responsibility to the CMS. 

This is a mandatory change because we want much more flexibility on this like easily move ZIM for dev to prod and host multiple copies of same ZIM in multiple warehouses. This will also allow to free Zimfarm editors from the responsibility to decide where the ZIM should be stored.

## warehouse path

The notion of warehouse path will be removed from the Zimfarm. ZIMs will all be stored in a single location on Zimfarm side. The CMS will then handover the ZIM file and move it as required, when there is a CMS plugged after the Zimfarm.

## task manager, uploader and receiver

The task manager, uploader and receiver will stop managing ZIM warehouse paths. 

The task manager will be responsible to rename the ZIM files produced by the scraper to ensure a unique ZIM name and avoid collisions: ZIM files will be renamed to `<zim_uuid.zim>`. It means the task manager will have a dependency on the python-libzim from this point to have access to ZIM UUID.

The task manager, uploader and receiver will then upload / move file at the root of the "location". The single receiver responsibility will now be to authenticate workers uploaers so that we ensure only allowed systems are publishing ZIM files. 

If possible, the receiver should refuse to overwrite a file which already exist. Or the uploader should check for that.

The fact the receiver is simple is important because we wanna avoid to restart it often since it will cause ZIM uploads to have to restart.

## CMS is optional

It is important to keep in mind that CMS is an optional component and we will continue to have Zimfarm(s) (like farm.zimit.kiwix.org) which do not have a CMS running after them. We however need to maintain compatibility between both options.

elfkuzco:
## Rationale
This PR enhances the zimfarm by overwriting the warehouse in the response when  `ZIMCHECK_OPTION` configured. This does not modify the DB entry, only the response. Consequently, if there's a zimcheck option, the worker will rename the file to be based on the zim uuid.

## Changes
- rename files to be of the form `<zim_uuid>.zim` when there's a zimcheck option configured
- overwrite the warehouse path response to be the root folder if there's a zimcheck option configured. 
- adapt UI to generate filename from zim info if one exists otherwise default to the file name. This has the effect that when `ZIMCHECK_OPTION` is set, the filename first shows as a UUID before being updated when the zimcheck is finally uploaded.

This closes #1438


elfkuzco:
> > force warehouse_path to "root" when creating a requested task and option USE_WAREHOUSE_PATH is disabled (ignoring any warehouse path configured in the schedule)

I just realized that leaving it in the requested task only would be the best place since the task is creating from the requested task.

elfkuzco:
I have disabled the receiver cron job based on an env var `ENABLE_ZIM_MOVE`

backend/src/zimfarm_backend/db/tasks.py
```diff
@@ -106,7 +106,7 @@ def get_task_by_id_or_none(session: OrmSession, task_id: UUID) -> TaskFullSchema
             timestamp=task.timestamp,
             config=ExpandedScheduleConfigSchema.model_validate(
                 {
-                    "warehouse_path": task.config["warehouse_path"],
+                    "warehouse_path": (task.config["warehouse_path"]),
```
benoit74: why adding parentheses?

receiver/entrypoint.sh
```diff
@@ -22,8 +22,12 @@ cat $JSON_PATH
 # Start openssh-server
 /etc/init.d/ssh start
 
-# Create cron entry for ZIM quarantine
-echo "* *  * * *  root  /usr/bin/flock -w 0 /dev/shm/cron.lock /usr/local/bin/check_zims.sh $ZIMCHECK_PARALLEL_JOBS $ZIM_SRC_DIR $ZIM_DST_DIR $ZIM_QUAR_DIR >> /dev/shm/check_zims.log 2>&1" >> /etc/cron.d/check_zims
-chmod +x /etc/cron.d/check_zims
+ENABLE_ZIM_MOVE=${ENABLE_ZIM_MOVE:-1}
```
benoit74, line 25: ENABLE_ZIM_MOVE should be enabled by default

receiver/entrypoint.sh
```diff
@@ -22,8 +22,12 @@ cat $JSON_PATH
...
+
+# Create cron entry for ZIM quarantine if ENABLE_MOVE is set
```
benoit74: I know it was here before your came, but comment is a bit misleading:

`Create cron entry for ZIM management in receiver`

receiver/entrypoint.sh
```diff
@@ -22,8 +22,12 @@ cat $JSON_PATH
...
+if [ "$ENABLE_ZIM_MOVE" = "1" ]; then
+    echo "* *  * * *  root  /usr/bin/flock -w 0 /dev/shm/cron.lock /usr/local/bin/check_zims.sh $ZIMCHECK_PARALLEL_JOBS $ZIM_SRC_DIR $ZIM_DST_DIR $ZIM_QUAR_DIR >> /dev/shm/check_zims.log 2>&1" >> /etc/cron.d/check_zims
```
benoit74: I know it was here before your came, but now I can't bare them anymore ^^

Can you please rename all these from `check_zims` to `move_zims`? This would clearly help maintenance.

receiver/README.md
```diff
@@ -34,6 +34,8 @@ A file events watcher that moves incoming ZIM files to final location after a su
 Previously, it used `zimcheck` from libzim tools to check that ZIMs are valid and move them to quarantine folder if check fails. Now, it just checks that they are in appropriate
 subfolder and moves them to final destination.
 
+To disable the movement of ZIM files ane leave them where they are, set `ENABLE_ZIM_MOVE` to any value other than `1`.
```
benoit74: `ane` => `and`

`leave them where they are` => `leave them in the upload jail`

receiver/README.md
benoit74: `ZIM quarantaine` => `ZIM move`

worker/src/zimfarm_worker/task/worker.py
```diff
@@ -1056,10 +1057,16 @@ def refresh_files_list(self):
             return
         for fpath in self.task_workdir.glob("*.zim"):
             if fpath.name not in self.zim_files_actions_status.keys():
+                # Rename file to <zim_uuid>.zim if we are ignoring warehouse path
+                if not bool(self.task["upload"]["zim"].get("use_warehouse_path")):
```
benoit74: This is not correct, if `use_warehouse_path` is None (because it is an old task which did not had this flag) then we should use former logic. Maybe switching to `DISABLE_WAREHOUSE_PATH` (to have the reverse logic) would make things easier

backend/src/zimfarm_backend/api/routes/tasks/logic.py
```diff
@@ -17,7 +17,10 @@
 )
 from zimfarm_backend.api.routes.models import ListResponse
 from zimfarm_backend.api.routes.tasks.models import TaskCreateSchema, TaskUpdateSchema
-from zimfarm_backend.common.constants import ENABLED_SCHEDULER, INFORM_CMS
+from zimfarm_backend.common.constants import (
```
benoit74: Remove these irrelevant changes

backend/src/zimfarm_backend/api/routes/tasks/logic.py
```diff
@@ -118,6 +121,7 @@ def get_task(
     )
     if INFORM_CMS:
         populate_zim_urls(cast(TaskFullSchema, task))
+
```
benoit74: Remove these irrelevant changes

backend/src/zimfarm_backend/common/schemas/orms.py
```diff
@@ -333,6 +334,7 @@ class ScheduleFullSchema(BaseModel):
     similarity_data: list[str] = Field(exclude=True)
     offliner: str
     version: str
+    use_warehouse_path: bool
```
benoit74: Is this used anywhere?

backend/src/zimfarm_backend/db/schedule.py
```diff
@@ -489,6 +490,7 @@ def create_schedule_full_schema(
         version=schedule.offliner_definition.version,
         offliner=schedule.offliner_definition.offliner,
         archived=schedule.archived,
+        use_warehouse_path=USE_WAREHOUSE_PATH,
```
benoit74: Is this used anywhere?

receiver/entrypoint.sh
elfkuzco: It is enabled by default. That is the purpose of the `:-1` syntax.

backend/src/zimfarm_backend/common/schemas/orms.py
elfkuzco: It's how I intend to communicate the UI to hide the dropdown for warehouse paths

backend/src/zimfarm_backend/db/schedule.py
elfkuzco: It's how I intend to communicate to the UI to hide the dropdown for warehouse paths

backend/src/zimfarm_backend/db/tasks.py
elfkuzco: seems to be from ruff

backend/src/zimfarm_backend/common/schemas/orms.py
benoit74: A simple option in `config.json` is probably sufficient. Should API and UI have discrepancy, it is not going to break much things. Or do I miss something?

receiver/entrypoint.sh
benoit74: Indeed, sorry

backend/src/zimfarm_backend/common/schemas/orms.py
elfkuzco: Nope, nothing will break.

elfkuzco:
What should be done when the receiver `ENABLE_ZIM_MOVE` is toggled on/off? Should we change the way it moves files to ignore files that are of the form `<zim_uuid>.zim`. Currently, if I toggle it on and off, it will attempt to move the zim files at the root to the quarantine folder because it uses `find -iname '*.zim'` to find ZIM files.

benoit74:
> What should be done when the receiver ENABLE_ZIM_MOVE is toggled on/off? 

It is fine if we have files left in the root of the jail **and** we turn `ENABLE_ZIM_MOVE` back on. We will take care of that situation manually (and it is just files moves, not a big deal).

frontend-ui/src/views/TaskDetailView.vue
```diff
@@ -762,6 +762,28 @@ const sortedFiles = computed(() => {
...
+    const date = metadata.Date
+    const flavour = metadata.Flavour
+
+    if (name && date) {
+      const parts = [name]
+      if (flavour) {
+        parts.push(flavour)
+      }
+      parts.push(date)
+      return parts.join('_') + '.zim'
```
benoit74: `parts.join('_') + '.zim'` is going to be too misleading we could believe it is the real filename while it is not ; I propose to rather use `parts.join(' / ')` so that we clearly know we have different data.

Ultimately, I would prefer to have 3 columns, but it means that should we produce multiple ZIM, we might end-up with a table with some files on 1 column and some on 3 which is not possible. 

We can live with this "hack" until I open and then we implement https://github.com/openzim/zimfarm/issues/1765 afterwards

PR #1767, opened 2026-03-22
show warnings about worker manager being out of date

elfkuzco:
## Rationale
This PR enhances the worker to report it's docker image information while checking in with the API. Accordingly, it warns the user of the outdated image and informs them to perform updates.

## Changes
- enhance worker to submit docker image information during checkin
- fetch and cache latest worker-manager image information on API
- show outdated worker information on UI
Screenshot_20260320_145748
Screenshot_20260320_145724


backend/src/zimfarm_backend/common/constants.py
```diff
@@ -149,6 +150,16 @@ def getenv(key: str, *, mandatory: bool = False, default: Any = None) -> Any:
 REQ_TIMEOUT_CMS = int(getenv("REQ_TIMEOUT_CMS", default="10"))
 REQ_TIMEOUT_GHCR = int(getenv("REQ_TIMEOUT_GHCR", default="10"))
 
+# Credentials for fetching zimfarm-worker-manager package versions
+GITHUB_TOKEN = getenv("GITHUB_TOKEN", default="")
```
benoit74, line 155: Can we use fine-grained PAT? Which permissions are required?

worker/src/zimfarm_worker/manager/worker.py
```diff
@@ -241,28 +253,77 @@ def check_in_at(self, webapi_uri: str):
...
+            and self.local_worker_manager_version.created_at
+            < self.api_worker_manager_version.created_at
+        ):
+            logger.warning(
+                "Local worker manager image is out of date: "
+                f"\n\tLatest Image Hash={self.api_worker_manager_version.hash}"
+                f"\n\t\tCreated={self.api_worker_manager_version.created_at.isoformat()}"
+                f"\n\tLocal Image Hash={self.local_worker_manager_version.hash}"
+                f"\n\t\tCreated={self.local_worker_manager_version.created_at.isoformat()}"
+                "\nPlease, update the worker manager to the latest "
```
benoit74, line 322: No need to run `zimfarm update do`, simply running `zimfarm restart` (or `zimfarm start`, which are doing the exact same thing) will pull latest worker-manager image.

benoit74:
Please also update `worker/README.md` to recommend worker owner setup two crontab entries (provide sample crontab entries), one to run `zimfarm restart` once a day and one to run `zimfarm update` once a week, and indicate that day / time of the day should be chosen wisely so that they can be available should it break something.

backend/src/zimfarm_backend/common/constants.py
elfkuzco: Yes, Only the `read:packages` permission is required

worker/src/zimfarm_worker/manager/worker.py
elfkuzco: `zimfarm update do` has the side effect of pulling the latest version of the shell script

backend/src/zimfarm_backend/common/constants.py
benoit74: Please add a comment for that

worker/src/zimfarm_worker/manager/worker.py
benoit74: Updating the shell script is really more rare, and I would not recommend to do it everytime worker manager needs to be updated. Do you see a reason to do so?

backend/src/zimfarm_backend/common/constants.py
elfkuzco: Done

worker/src/zimfarm_worker/manager/worker.py
elfkuzco: No. I just thought when a new env variable is added and worker behaviour changes, user might want to have the latest version of the script that has the variable. But given the worker sets reasonable defaults, I'm guessing it's not that important.

PR #1783, opened 2026-03-24
rename schedule to recipe

### LINKED ISSUE #1768: Clean confusion between `schedule` and `recipe` terms
Quite late in the inception of Zimfarm, it has been decided that schedules should in fact be named recipes (correct me if I'm wrong, I wasn't there).

Due to limited resources available at that time and other priorities, only UI has been fixed to use the `recipe` term, but everything code-related (API endpoint, DB object, code base ...) continues up to now to use the `schedule` term

Now that we are in 2026, we know that `recipe` seems to be quite a good term since this is what users have used for years without complaining.

It however always causes confusions in discussions between devs and the rest of the world, including new devs.

I propose that we complete the move and migrate everything to the `recipe` term.

@kelson42 please speak-up in the coming days if this move does not suits you anymore or we will proceed

Migration will have impact on ongoing Zimfarm components (worker manager, task manager, ...) but also WP1 and zimit-frontend. We will hence offer a backward compatible API for few weeks / months and probably implemented ourselves PRs for WP1 and zimit-frontend.

elfkuzco:
## Rationale
This PR renames `schedule` to `recipe` across the codebase. In order to maintain compatibility with wp1 and worker manager, it keeps a schedule router too on top of the new recipes router and accepts query params that mention `schedule`. Additionally, it also returns `schedule_` fields in the JSON response so integrations like wp1 and zimit-frontend can continue to function properly

## Changes
- rename schedules to recipes across the codebase
- create migrations for db to rename table, constraints and recreate constraints
- submit file info when file is marked as created from the task worker. This ensures file info will always be sent regardless of whether checker is enabled or not. Additionally, it improves the UI as the file info will be used to compute the name instead of cryptic UUID when warehouse path is disabled (see #1758). Also, the upload status of files is now dynamic and shows various states of the uploading state with the download icons only shown when  a file has indeed being uploaded
Screenshot_20260325_110455
Screenshot_20260325_105121

This closes #1768


audiodude:
Obviously I'm not familiar with the Zimfarm codebase, so can you point out where the `/schedules/*` endpoints are preserved?

elfkuzco:
> Obviously I'm not familiar with the Zimfarm codebase, so can you point out where the /schedules/* endpoints are preserved?

Here it is defined in the same file as the `recipes` router and used across the routes. There are other places where the JSON response returned `schedule_name` but those are kept too even though they now return an additional `recipe_name`. All mentions of `schedule` in JSON responses are kept intact.  https://github.com/openzim/zimfarm/blob/df3698476342dba90ad785ec655d4ebc9bd9fa54/backend/src/zimfarm_backend/api/routes/recipes/logic.py#L100-L106

benoit74:
Just wanted to mention I did not forget about this but wanted to review this in a quiet moment. Not yet found time for that ^^

PR #1786, opened 2026-03-25
upgrade watcher

### LINKED ISSUE #1785: improve watcher code with type hints and explicitness
While most of the zimfarm components have been updated, the watcher has been largely untouched. All of the functions have no type annotations and it is unclear how initialization of the instance attributes are done (it uses `setattr`). This leads to strange logic like https://github.com/openzim/zimfarm/blob/96fc33887d873085dfd16b8e0218d71f100c2df2/watcher/watcher.py#L462-L465 where the `schedule_in_zimfarm` (supposedly a method) is used both as a check (for truthiness) and invoked as a call

elfkuzco:
## Rationale
This PR improves the watcher code by providing type hints, setting up CI rules to enforce quality and adapt code for dev environments.

## Changes
- set up CI for enforcing watcher code quality
- add new env `S3_SUMMARY_FILE_URL` to allow user/dev to provide custom URL for summary file as default doesn't work in minio
- pass parameters in init instead of using `setattr`
- use parameter `schedule_on_update` to determine if schedule should be updated instead of using the function name for the check https://github.com/openzim/zimfarm/blob/96fc33887d873085dfd16b8e0218d71f100c2df2/watcher/watcher.py#L462-L465

This closes #1785

### DIFF
--- a/.github/workflows/watcher-QA.yml
+++ b/.github/workflows/watcher-QA.yml
@@ -1,0 +1,42 @@
+name: Watcher QA
+
+on:
+  pull_request:
+    paths:
+      - "watcher/**"
+  push:
+    paths:
+      - "watcher/**"
+    branches:
+      - main
+
+jobs:
+  check-qa:
+    runs-on: ubuntu-24.04
+    steps:
+      - uses: actions/checkout@v6
+
+      - name: Set up Python
+        uses: actions/setup-python@v6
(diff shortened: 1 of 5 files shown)


elfkuzco:
marking as draft until decision is made on #1783 as API calls are made to the `recipes` enpdoint

PR #1788, opened 2026-03-26
improve contributors setup flow

elfkuzco:
## Rationale
This PR aims to improve the contributor setup flow by directing them towards the `dev/README.md` where there are now scripts to create offliners, create workers and import recipes from production database. This way, everything is seamless and new devs can get to work on the components they want to without issues.

## Changes
- add new script to import recipes from production APIs (secrets are greyed out since new user has no permissions to view secrets)
- move sections from the contributing guide to the `dev/README.md`
- remove test SSH keys as script to create worker now generates keys and moves them to the correct location
- provide example commands on how to perform every beginner action
- update logic to create initial user by checking if new user exists instead of user count as alembic migration creates some maintenance accounts and this made the initial user to not be created


elfkuzco:
marking as draft till #1786 is merged

PR #1792, opened 2026-03-27
diagnose why a requested task is not starting

### LINKED ISSUE #908: Report why a task is not starting
We regularly get questions about why a given requested task is not starting.

While this is easy for the machine to know, it consumes significant developer resources to check all existing reasons, and it will probably only get worse in the future where we will probably add more complex scheduling constraints.

As a user, it would be nice to know why a given requested task is not starting.

I don't know yet how feasible it is to compute this information for all requested tasks, refresh it every 5 minutes or so, and hence display this information on all requested tasks.

Or maybe this could be an "on-demand" feature where the user will click a button and get a reason.

Probably the situation is so complex that we cannot give a single reason, but we will have to display one reason per worker in many situations. Meaning the reason is not just a phrase but a list of constraints blocking the start of the task.

elfkuzco:
## Rationale
This PR enhances the API to diagnose why a requested task isn't starting on a particular worker.
Screenshot_20260327_130640
Screenshot_20260327_130128
Screenshot_20260327_124756

## Changes
- add endpoint to diagnose why a requested task isn't starting
- refactor logic to find a requested task for worker so code can be reused
- add button to diagnose a requested task (requires permission of users who can create requested tasks). UI runs this diagnostic for the first 20 workers as notifications can be noisy

This closes #908

### DIFF
--- a/backend/src/zimfarm_backend/db/requested_task.py
+++ b/backend/src/zimfarm_backend/db/requested_task.py
@@ -2,14 +2,13 @@ import datetime
 from typing import cast
 from uuid import UUID
 
-from humanfriendly import format_size
-from pydantic import Field
+from humanfriendly import format_size, format_timespan
 from sqlalchemy import BigInteger, delete, func, or_, select, update
 from sqlalchemy.orm import Session as OrmSession
 from sqlalchemy.orm import selectinload
 
 from zimfarm_backend import logger
-from zimfarm_backend.common import getnow
+from zimfarm_backend.common import getnow, to_naive_utc
 from zimfarm_backend.common.constants import (
     ARTIFACTS_EXPIRATION,
     ARTIFACTS_UPLOAD_URI,
@@ -31,6 +30,7 @@ from zimfarm_backend.common.schemas.models import (
     ScheduleNotificationSchema,
 )
 from zimfarm_backend.common.schemas.orms import (
+    BaseRequestedTaskSchema,
     ConfigResourcesSchema,
     ConfigWithOnlyOfflinerAndResourcesSchema,
     OfflinerDefinitionSchema,
@@ -39,6 +39,7 @@ from zimfarm_backend.common.schemas.orms import (
     RequestedTaskLightSchema,
     ScheduleDurationSchema,
     TaskUploadSchema,
+    WorkerLightSchema,
 )
 from zimfarm_backend.db import count_from_stmt
 from zimfarm_backend.db.exceptions import RecordDoesNotExistError
@@ -50,7 +51,7 @@ from zimfarm_backend.db.offliner_definition import (
 )
 from zimfarm_backend.db.schedule import get_schedule_duration, get_schedule_or_none
 from zimfarm_backend.db.tasks import RunningTask, get_currently_running_tasks
-from zimfarm_backend.db.worker import get_worker_or_none
+from zimfarm_backend.db.worker import create_worker_schema, get_worker_or_none
 from zimfarm_backend.utils.offliners import expanded_config
 from zimfarm_backend.utils.timestamp import (
     get_status_timestamp_expr,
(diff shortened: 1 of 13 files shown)


frontend-ui/src/components/PipelineTable.vue
```diff
@@ -332,6 +353,43 @@ async function handleLoadAllLastRuns() {
...
 }
+
+async function diagnoseTask(task: RequestedTaskLight) {
+  diagnosingTasks.value[task.id] = true
+  try {
+    let workersToDiagnose: string[] = []
+    if (task.worker_name) {
+      workersToDiagnose = [task.worker_name]
+    } else {
+      const workersList = await workersStore.fetchWorkers({ limit: 20 })
```
benoit74: This is a tool which will be rarely used, but needs to show a thorough explanation, so we should:
- run on the whole list of workers, not only the first 20, this is going to be misleading
- implement a "real" UI to show results (10 or 20 notifications is a bit of a mess to read)
- run diagnostics only one worker at a time, not all in parallel, and add a 1 sec pause between each diagnostic to reduce strain on the API / DB, and also to make the thing a bit slow so that people do not abuse it "just for fun"
- display diagnostic results as they arrive ; I imagine something where the UI piles up results with something like "workerxx: Cheking ..." ; then "Workerxxx: cannot start the task because ...", then "worker xyz: cheking ...", etc ...
- corolarry: report all cases where a task should be able to start on a given worker from current perspective, worker by worker (i.e. not consider the whole situation but report it worker by worker)
- inverse the logic between errors and success ; this diagnostic button will typically be used to check why a task is stuck in the pipe, so the fact that a given requested task cannot start on a given worker is more something normal, and the fact that we have a no clue why a task is not starting on a given worker is abnormal
- consider the case where the task is not starting only because workers are starting only one task at a time, and worker has started a task few secs / minutes ago ; while we don't know how often the worker polls, if we find no reason a worker is not starting a given task, we should check how long ago a task has been started in the worker and indicate it as a potential explanation ; some along the lines of "we have no reason for this task to not start on this worker but workers do start only one task at a time and last task started xx minutes ago"

frontend-ui/src/components/PipelineTable.vue
```diff
@@ -153,6 +153,19 @@
           <span v-else>n/a</span>
         </template>
 
+        <template #[`item.diagnose`]="{ item }">
```
benoit74, line 156: This button is very visible and I'm a bit worried users won't get what this is about and just click it just to see or just too often.

I suggest we keep it where it is but we add a confirm popup before running the diagnostic which explain something along the lines of what the diagnose will do, that it will consume significant resources and should be run only when something is really unexplained / seems abnormal

backend/src/zimfarm_backend/db/requested_task.py
```diff
@@ -614,120 +652,232 @@ def get_possible_task_with_resources(
...
-        )
+    if reason := does_platform_allow_worker_to_run(
+        worker=worker,
+        all_running_tasks=all_running_tasks,
+        running_tasks=running_tasks,
+        task=task,
+    )[1]:
+        return reason
+
+    avail_memory = worker.resources.memory - sum(
```
benoit74, line 738: As discussed on Slack, we should not compute worker available resources. Problem is that this is not the exact same logic which will run when worker will ask for a next task to start. 

The reason we introduced current worker available resources and not expected ones from zimfarm perspective is that some workers might poll tasks from multiple zimfarm instances. And what the zimfarm knows about the worker is hence not the worker reality. 

I feel like it would be better to store real worker available resources in the DB (whenever worker polls for a next task to start) ; it will provide more accurate results for this check and be helpful info anyway.

frontend-ui/src/components/PipelineTable.vue
elfkuzco: It's not visible unless user has permission to create a task

benoit74: Yep, but among users with this permission not everyone is knowledgeable of technical limitations and details

backend/src/zimfarm_backend/db/requested_task.py
elfkuzco: okay. will implement storing the stats on poll after discussion on #1795 as current "available" resources submitted during poll aren't accurate

frontend-ui/src/components/PipelineTable.vue
elfkuzco: updated PR description with images to address these

frontend-ui/src/components/PipelineTable.vue
elfkuzco: added  a confirmation dialog

PR #1801, opened 2026-03-31
Move from login.kiwix.org to ory.login.kiwix.org

benoit74:
For https://github.com/kiwix/overview/issues/157


benoit74:
@elfkuzco FYI

PR #1808, opened 2026-03-31
always return diagnostic reason

### LINKED ISSUE #1806: Enhance diagnose reasons when we have no reasons
In task diagnostics, when we have no reason to not start a given task, we display something like `last task started xxx ago`.

The value is wrong, it is computed based on all timestamps while we wanna know when last task started on this worker, not when it got an update.

elfkuzco:
## Changes
- always return diagnostic reason for why a task isn't starting
- use task started timestamps to build up list of running task timestamps instead of their `updated_at`. default to `reserved` since all tasks must be reserved
- rename default function that returns diagnostic reason to `_get_worker_unavailable_reason`

This closes #1806

### DIFF
--- a/backend/src/zimfarm_backend/db/requested_task.py
+++ b/backend/src/zimfarm_backend/db/requested_task.py
@@ -55,6 +55,7 @@ from zimfarm_backend.db.worker import create_worker_schema, get_worker_or_none
 from zimfarm_backend.utils.offliners import expanded_config
 from zimfarm_backend.utils.timestamp import (
     get_status_timestamp_expr,
+    get_timestamp_for_status,
 )
 
 # Maximum positive integer value for PostgreSQL BigInteger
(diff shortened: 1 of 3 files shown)


backend/src/zimfarm_backend/db/requested_task.py
```diff
@@ -664,10 +665,18 @@ def _can_run(task: RequestedTaskWithDuration, resource: ResourcesSchema) -> bool
     return True
 
 
-def _check_worker_unavailable_reason(
+def _get_worker_unavailable_reason(
     worker: WorkerLightSchema, running_tasks: list[RunningTask]
-) -> str | None:
-    running_task_timestamps = [t.updated_at for t in running_tasks]
+) -> str:
+    running_task_timestamps = [
```
benoit74, line 671: Shouldn't we only use `reserved` here, since the fact the task really started is not relevant for worker to reserve another task to start at next poll interval? Or do I miss something?

elfkuzco: No, I will use `reserved`

elfkuzco:
Thought you had approved and I squashed the commit

PR #1821, opened 2026-04-03
poll api with remaining resource statistics

### LINKED ISSUE #1795: manager requesting for tasks with system stats instead of provisioned stats
Currently, the manager tries to find a task that it can do by sending a query with the available CPU, memory and disk. However, these resource units are the ones available on the worker machine and not the remainder based on what owner says is the total. For example, I have a worker set up to report `5GiB` even though I have `30Gib` free space on `/var` (docker volume). When worker checks in, it correctly reports it's total disk usage of `5Gib`. However, when it starts polling, it uses `avail_disk` of `28.5GiB` to find a task which is incorrect. Even though this appears to be initially harmless because the algorithm to find a task uses the worker-level  stats to filter tasks it can do at the DB level, it can become problematic when it tries to pick the next candidate that can be run with the "available" resources. Even more so is that we now want to diagnose why a requested task isn't starting (#908) and so we want to always update the worker stats with the "available" resources.

https://github.com/openzim/zimfarm/blob/dbaf57eae7d52d98a65af0654f23e7db801f5c5e/workers/app/common/docker.py#L164-L190

https://github.com/openzim/zimfarm/blob/dbaf57eae7d52d98a65af0654f23e7db801f5c5e/workers/app/manager/worker.py#L144-L169

https://github.com/openzim/zimfarm/blob/dbaf57eae7d52d98a65af0654f23e7db801f5c5e/dispatcher/backend/src/utils/scheduling.py#L328-L343

@rgaudin, is there a reason, the worker is configured to find tasks using the "available" stats instead of the "remaining" one?

elfkuzco:
## Rationale
This PR updates worker manager to correctly poll the API with remaining resources instead of the available resources (see #1795)

## Changes
- poll API with remaining resources
- update worker resources with those sent in poll requests
- adapt UI to compute total as the sum of the worker resource and the tasks it is currently executing
- adapt logic to diagnose requested task to use the worker's reported stats for diagnosis`

This closes #1795

### DIFF
--- a/backend/src/zimfarm_backend/api/routes/requested_tasks/logic.py
+++ b/backend/src/zimfarm_backend/api/routes/requested_tasks/logic.py
@@ -25,6 +25,7 @@ from zimfarm_backend.api.routes.requested_tasks.models import (
     RequestedTaskSchema,
     UpdateRequestedTaskSchema,
 )
+from zimfarm_backend.api.routes.workers.models import GetRequestedTaskSchema
 from zimfarm_backend.common import WorkersIpChangesCounts, getnow
 from zimfarm_backend.common.constants import (
     ENABLED_SCHEDULER,
@@ -32,13 +33,7 @@ from zimfarm_backend.common.constants import (
     USES_WORKERS_IPS_WHITELIST,
 )
 from zimfarm_backend.common.external import update_workers_whitelist
-from zimfarm_backend.common.schemas.fields import (
-    ZIMCPU,
-    NotEmptyString,
-    WorkerField,
-    ZIMDisk,
-    ZIMMemory,
-)
+from zimfarm_backend.common.schemas.fields import NotEmptyString
 from zimfarm_backend.common.schemas.models import calculate_pagination_metadata
 from zimfarm_backend.common.schemas.orms import (
     ConfigResourcesSchema,
@@ -183,10 +178,7 @@ def get_requested_tasks(
 @router.get("/worker")
 def get_requested_tasks_for_worker(
     request: Request,
-    worker_name: Annotated[WorkerField, Query()],
-    avail_cpu: Annotated[ZIMCPU, Query()],
-    avail_memory: Annotated[ZIMMemory, Query()],
-    avail_disk: Annotated[ZIMDisk, Query()],
+    query: Annotated[GetRequestedTaskSchema, Query()],
     session: Annotated[OrmSession, Depends(gen_manual_dbsession)],
     current_account: Annotated[
         Account, Depends(get_current_account_with_session(session_type="manual"))
@@ -194,7 +186,7 @@ def get_requested_tasks_for_worker(
 ) -> ListResponse[RequestedTaskLightSchema]:
     """Get list of requested tasks for a worker."""
 
-    worker = get_worker(session, worker_name=worker_name)
+    worker = get_worker(session, worker_name=query.worker_name)
 
     fallback_ip = request.client.host if request.client else None
     x_forwarded_for = request.headers.get("X-Forwarded-For", fallback_ip)
@@ -203,23 +195,30 @@ def get_requested_tasks_for_worker(
 
         if ip_changed:
             logger.info(
-                f"Worker {worker_name} IP changed from {worker.last_ip} to "
+                f"Worker {query.worker_name} IP changed from {worker.last_ip} to "
                 f"{x_forwarded_for}"
             )
-            worker = update_worker(
-                session, worker_name=worker_name, ip_address=x_forwarded_for
-            )
-        else:
-            worker = update_worker(session, worker_name=worker_name)
 
-        # commit explicitly last_ip and last_seen changes, since we are not
-        # using an explicit transaction, and do it before calling Wasabi so
-        # that changes are propagated quickly and transaction is not blocking
+        worker = update_worker(
+            session,
+            worker_name=query.worker_name,
+            ip_address=x_forwarded_for if ip_changed else None,
+            avail_disk=query.avail_disk,
+            avail_memory=query.avail_memory,
+            avail_cpu=query.avail_cpu,
+            total_disk=query.total_disk,
+            total_memory=query.total_memory,
+            total_cpu=query.total_cpu,
+        )
+
+        # commit explicitly since we are not using an automatic transaction,
+        # and do it before calling Wasabi so that changes are propagated quickly
+        # and transaction is not blocking
         session.commit()
 
         if ip_changed and USES_WORKERS_IPS_WHITELIST:
             try:
-                record_ip_change(session, worker_name)
+                record_ip_change(session, query.worker_name)
             except Exception as exc:
                 logger.exception("Pushing IP changes to Wasabi failed")
                 raise ServiceUnavailableError(
@@ -240,9 +239,9 @@ def get_requested_tasks_for_worker(
     task = find_requested_task_for_worker(
         session=session,
         worker=create_worker_schema(worker),
-        avail_cpu=avail_cpu,
-        avail_memory=avail_memory,
-        avail_disk=avail_disk,
+        avail_cpu=query.avail_cpu,
+        avail_memory=query.avail_memory,
+        avail_disk=query.avail_disk,
     ).requested_task
 
     return ListResponse(
--- a/backend/src/zimfarm_backend/db/models.py
+++ b/backend/src/zimfarm_backend/db/models.py
@@ -126,13 +126,16 @@ class Worker(Base):
     )
     name: Mapped[str] = mapped_column(unique=True, index=True)
     selfish: Mapped[bool]
-    cpu: Mapped[int]
-    memory: Mapped[int] = mapped_column(type_=BigInteger)
-    disk: Mapped[int] = mapped_column(type_=BigInteger)
+    total_cpu: Mapped[int]
+    total_memory: Mapped[int] = mapped_column(type_=BigInteger)
+    total_disk: Mapped[int] = mapped_column(type_=BigInteger)
     offliners: Mapped[list[str]]
     platforms: Mapped[dict[str, Any]]
     last_seen: Mapped[datetime | None]
     last_ip: Mapped[IPv4Address | None]
+    available_cpu: Mapped[int]
+    available_memory: Mapped[int] = mapped_column(type_=BigInteger)
+    available_disk: Mapped[int] = mapped_column(type_=BigInteger)
     deleted: Mapped[bool] = mapped_column(default=False, server_default=false())
     contexts: Mapped[dict[str, Any]] = mapped_column(
         default_factory=dict, server_default="{}"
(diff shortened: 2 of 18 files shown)


elfkuzco:
@benoit74 , because I adapted the logic in the UI to compute the total resources of the worker as the sum of it's resources and it's currently running tasks, there is a bit of a misinformation when the scraper exits and the worker polls the API with updated stats. This discrepancy appears because the list of running tasks in the UI includes `scraper_completed` and when the worker polls, it might exclude the scraper which just exited (and it's resources were claimed). This might be due to the fact that my poll interval is set to `10s` in development. Do you think it's necessary to add an icon to the UI saying "stats might be a little bit skewed"?

benoit74:
I agree there is something wrong here. I will dive into the code and come back to you.

worker/src/zimfarm_worker/common/docker.py
```diff
@@ -247,7 +247,12 @@ def query_host_stats(client: DockerClient, workdir: Path):
     mem_avail = as_pos_int(ZIMFARM_MEMORY - mem_used)
 
     return HostStats(
```
benoit74, line 234: I don't get why we are making such inconsistent things. We have three metrics, and each one has a distinct logic, this makes no sense to me.

I do get we have different ways to get "used" data, but then computation of available and remaining should always follow the same pattern IMHO.

Can you propose something more "aligned" or explain the discrepancies?

What is important to me is that worker owner can assign only some of its machine resources to the Zimfarm and the worker manager acts accordingly (and we probably never know about real total machine resources).

worker/src/zimfarm_worker/manager/worker.py
```diff
@@ -209,9 +209,9 @@ def poll_api(self, webapi_uri: str) -> bool:
             path="/requested-tasks/worker",
             params={
                 "worker_name": self.worker_name,
-                "avail_cpu": host_stats.cpu.available,
-                "avail_memory": host_stats.memory.available,
-                "avail_disk": host_stats.disk.available,
+                "avail_cpu": host_stats.cpu.remaining,
```
benoit74: We should probably report everything needed by the UI here, so that we do not make any assumptions like summing all running tasks, this is going to be too risky / misleading. 

Probably reporting total + remaining is sufficient. But then I would recommend to change field name for clarity.

Might means we need to make assumptions in the UI for the transitory phase, but we need to clean this.

worker/src/zimfarm_worker/common/docker.py
benoit74: And I'm not really convinced we need to report the 4 metrics here once they are consistent

elfkuzco: > Can you propose something more "aligned" or explain the discrepancies?

With the exception of the disk's `available`, the stats all are computed the same way. Given it's this `available` info that's the cause of the issue and it's not really used anywhere, I think it's simpler to remove it completely and only have the stats fields contain `total`, `used` and `remaining`.

worker/src/zimfarm_worker/manager/worker.py
elfkuzco: The DB `worker` table only contains rows for storing one stat for the CPU. If we continue storing the total (as it was before this PR), it would mean I have to compute the remaining when diagnosing a requested task. It is the diagnosis logic that really requires the remaining.

benoit74: Why don't we store both total and remaining in DB?

worker/src/zimfarm_worker/common/docker.py
benoit74: I agree.

elfkuzco:
Working on the host resource logic and i'm now convinced that the stats reported weren't the host stats (only the disk free is correct maybe because python recurses into the host directory to get actual disk metrics). the rest of the other stats are purely the sum of the running containers. according to this stack overflow link (https://stackoverflow.com/questions/76737334/how-to-get-hosts-processes-info-and-disk-space-usage-info-when-in-a-docker#76737864), we need to bind /host/proc filesystem to get the actual host metrics.

since we can get the accurate disk space on host without mount, teh only check at checkin/polling is done to check that what is remaining is available on the disk. there are no corresponding checks for available CPU/memory

EDIT: Because of the new migrations, I had to rebase on top of the latest changes

backend/src/zimfarm_backend/api/routes/requested_tasks/logic.py
```diff
@@ -203,14 +195,21 @@ def get_requested_tasks_for_worker(
...
+            ip_address=x_forwarded_for if ip_changed else None,
+            avail_disk=query.avail_disk,
+            avail_memory=query.avail_memory,
+            avail_cpu=query.avail_cpu,
+            total_disk=query.total_disk,
+            total_memory=query.total_memory,
+            total_cpu=query.total_cpu,
+        )
 
         # commit explicitly last_ip and last_seen changes, since we are not
```
benoit74: Comment probably deserves to be a bit more generic now that we are updating more than `last_ip` and `last_seen`.

worker/src/zimfarm_worker/manager/worker.py
```diff
@@ -124,26 +125,31 @@ def __init__(
...
             "(avail)"
             f"\n\tRAM : {format_size(host_stats.memory.total)} (total) ;  "
             f"{format_size(host_stats.memory.available)} (avail)"
             f"\n\tDisk: {format_size(host_stats.disk.total)} (configured) ; "
             f"{format_size(host_stats.disk.available)} (avail) ; "
             f"{format_size(host_stats.disk.used)} (reserved) ; "
-            f"{format_size(host_stats.disk.remaining)} (remaining)"
         )
 
-        if host_stats.disk.available < host_stats.disk.remaining:
```
benoit74, line 142: Sorry, now I recall a bit why we had this specificity around disk size. The problem of disk is that the total configured it is a "promise" which can easily be unfulfilled. By that I mean that the worker owner can easily promise to dedicate a given amount of disk to Zimfarm but this space might not be available anymore for any reason (including Zimfarm itself doing bad cleanup). This is true as well for CPU and mem but with way fewer consequences (and way less capability to check that).

This check you are removing "saved" us quite a handful times by ensuring that we quickly detect bad situations where the disk space promised in config is in fact not available anymore (and not because it is consumed by running tasks but because "something else" consumed it).

 I'm OK to remove this check with this PR only if we agree to put it back quickly in the aftermath. Or we should fix this now, with something clearer so we do not get confused again.

backend/src/zimfarm_backend/db/models.py
```diff
@@ -126,13 +126,16 @@ class Worker(Base):
...
-    memory: Mapped[int] = mapped_column(type_=BigInteger)
-    disk: Mapped[int] = mapped_column(type_=BigInteger)
+    total_cpu: Mapped[int]
+    total_memory: Mapped[int] = mapped_column(type_=BigInteger)
+    total_disk: Mapped[int] = mapped_column(type_=BigInteger)
     offliners: Mapped[list[str]]
     platforms: Mapped[dict[str, Any]]
     last_seen: Mapped[datetime | None]
     last_ip: Mapped[IPv4Address | None]
+    remaining_cpu: Mapped[int]
```
benoit74: Shouldn't we prefer to name this `available_xxx` so that we use the same name everywhere? I agree `remaining_xxx` is a bit better but I'm concerned we have different names in the codebase for the same thing. Or maybe even better, we use `remaining_xxx` but then we use it everywhere (which means we need a transition at API to support both params, but it is not that a big deal).

benoit74 (changes requested):
Sorry, I now realized what we are "breaking" with this change. A bit late, but still better than too late ^^

worker/src/zimfarm_worker/manager/worker.py
elfkuzco: Didn't actually remove it. Just refactored the logic to compute the actual disk space into a different function and then use the container stats to compare against it. 


https://github.com/openzim/zimfarm/blob/d47f1bf49f61fccc85c14799085e1a31c889b9b5/worker/src/zimfarm_worker/manager/worker.py#L201-L210

Previous implementation encapsulated the actual disk space within the stats object which was the root of the issue for the PR. `available` is computed as `ZIMFARM_DISK_SPACE - disk_used`. 

Think maybe you missed it because I did a rebase and the PR got force-pushed because. Had to rebase against the other PRs because there were already migrations introduced by the user-related PRs and it wasn't linear.

backend/src/zimfarm_backend/db/models.py
elfkuzco: Okay. I will include the `available` in the query params for backwards compatibility

worker/src/zimfarm_worker/manager/worker.py
elfkuzco: Will fix typo in `lowlower` :)

benoit74: I simply oversight this block of code ... sorry ... glad that at least you found a typo 🤣

backend/src/zimfarm_backend/db/models.py
elfkuzco: ended up renaming `remaining` to available as it introduces less changes and without hacks for supporting old query param from older workers

worker/src/zimfarm_worker/manager/worker.py
```diff
@@ -124,26 +125,31 @@ def __init__(
         # ensure we have access to docker API
         self.check_docker()
 
-        # display resources
-        host_stats = query_host_stats(self.docker, self.workdir)
+        # display resources. These aren't the actual "host" statistics but an
+        # aggreagation of the total stats used by our own containers.
```
benoit74: `aggreagation` => `aggregation`

worker/src/zimfarm_worker/manager/worker.py
```diff
@@ -124,26 +125,31 @@ def __init__(
...
+        host_stats = query_host_stats(self.docker)
+        disk_free = self.free_disk_space()
+        if host_stats.disk.available > disk_free:
+            self.should_stop = True
+            logger.critical(
+                "Configured disk space is not available. "
+                f"{format_size(disk_free)} (avail), "
```
benoit74: I would prefer we log a bit more things to have something clearer to read and act upon. It is not like we should have this message tons of times in the logs and something clear would help worker owner to know how to act. 

WDYT about this:

`Configured disk space is not available. Only {format_size(disk_free)} currently available while we expect at least {format_size(host_stats.disk.available)} to be available. This computation is based on running tasks expected disk consumption and a configured total of {format_size(host_stats.disk.total)} available for Zimfarm on the machine. Potential causes might be a task which consumes more disk than configured in its schedule, dangling objects (containers, volumes, images) or something else on the host consuming more disk than anticipated.`

benoit74 (approved):
LGTM, two minor remarks to consider and you can merge, no need for a new review.

worker/src/zimfarm_worker/manager/worker.py
elfkuzco: Better IMO

PR #1823, opened 2026-04-03
rename user to account

### LINKED ISSUE #1811: Rename "user" to "account"
The "user" term is incorrect in the DB, in the API and in the UI. What we are dealing with are more "account".

We should hence migrate, with a transition phase where old API is still supported (just like "recipe" vs "schedule").

elfkuzco:
## Rationale
This PR renames the user model to account across the DB and API

## Changes
- generate migrations to rename table and columns
- rename `user` to `account` across the codebase
- add route to view accounts via `/accounts` 

This closes #1811


backend/src/zimfarm_backend/api/routes/accounts/logic.py
```diff
@@ -0,0 +1,376 @@
...
+)
+from zimfarm_backend.exceptions import PublicKeyLoadError
+from zimfarm_backend.utils.cryptography import (
+    get_public_key_fingerprint,
+    get_public_key_type,
+    load_public_key,
+)
+
+router = APIRouter(prefix="/accounts", tags=["accounts"])
+users_router = APIRouter(prefix="/accounts", tags=["accounts"])
```
benoit74: `/users`?

benoit74 (changes requested):
One important bug to fix, and then we are good to go.

Note that I've check custom roles in the DB and we had one account which used a custom role with permissions on "users", which I already duplicated to "accounts".

Maybe a bit late to merge just before a long weekend, probably better to wait a bit before merging this.

benoit74:
Looks like you've spotted two bugs where I found only one 🤣 

Would you prefer to modify API endpoints the UI uses in this same PR to be on the safe side? I don't mind tbh.

elfkuzco:
>> Would you prefer to modify API endpoints the UI uses in this same PR to be on the safe side? I don't mind tbh.

Yeah

benoit74 (commented):
Sorry if what I said was confusing, but on frontend side I think the only thing to change is to change the name of the API service to call + probably use `account_id` instead of `user_id` where relevant + change permission check logic. The rest needs to continue using the "user" term since we will have "user" components and "worker" component. Maybe we need to rename store as well, I'm not sure. At least having duplicate Vue.JS routes is not needed, and moving them to "accounts" is probably not the right move since we will have only "users" displayed there.

benoit74 (approved):
LGTM, thank you!

Issue #1813, opened 2026-04-02
Allow to hide viewers from the "Users" list


benoit74:
UI to display users should allow to hide accounts with viewer role (checkbox or something like this): we have too many and they have no more permission than unauthenticated users.

Still showing all users by default allows to not make things too hidden and avoid confusions with search.

Should we hide viewers by default, when someone type something to search for, does it search in all users or only non-viewers? unclear situation. While if by default everything is shown, we just have to respect user decision. 

Hence if user selected to hide viewers, search should be done only on non-viewers.

Issue #1812, opened 2026-04-02
Do not show workers in the UI "Users" tab


benoit74:
By default, UI to display users (or users in "old words") should not show accounts with worker roles (they will be managed solely from "Workers" tab).

Issue #1819, opened 2026-04-02
Allow users to have finer permissions than today


benoit74:
For now, we have defined roles and permissions as illustrated in https://github.com/openzim/zimfarm/issues/1546. This issue seems to concentrate on the (re)definition of the roles and permissions, and where/how to store, define, enforce them.

We also have https://github.com/openzim/zimfarm/issues/1655 which concentrate on the fact that we want some users to be allowed to create (or edit) recipes only with mwoffliner scraper.

Both issues seems to clearly miss the big picture of what we want and need.

We also need to ensure users will not be able to modify any recipes but only the ones they've been granted access too. In a first approximation, we can probably say they will touch only the ones they've created. But we can imagine quite easily we will have to transfer some recipes to another person because the creator did not completed the job. And ultimately we probably strive to create small teams who collaborates on a set of recipes.

And we are probably not speaking only about recipes but also about workers where they can request the task to run. And they should probably not be allowed to make the recipe scheduled on a regular basis (ensuring not any worker will grab the job).

We also have the same kind of problem for worker owners we would like to allow to manage their workers from the UI (#1480). Most users will be allowed to interact with only one or few workers.

I consider that approaches like Google Zanzibar (implemented in Ory Keto) are far beyond what we need, but that a middle ground like OPA or AWS Cedar are probably the direction we need to move into. Note that I clearly miss the big picture on these solutions, just wanted to mention some names to clarify my points.

In summary: I feel like we need something more powerful than roles and permissions and broader than single attributes on DB / business objects to correctly implement the authorizations we need to have in Zimfarm in particular, but also at Kiwix in general (it is just that Zimfarm is the first tool we will significantly open to the public without a siloed approach like we have on imager-related services).

benoit74:
We've agree with @kelson42 that for now we will KISS and just implement something very similar to what we have on the CMS.

We will introduce a new concept in Zimfarm which is `team` and  move from `editor` and `editor-requester` roles, to `global-editor` and `global-editor-requester`, and add `team-editor` and `team-editor-requester` roles. When on a `team-xxx` role, users will be attached to zero or more teams. They will have editor (or editor-requester) access only to the recipes which are attached to their team(s), keeping simple viewer role on other recipes.

Recipes will all be attached to one or more team. One team is mandatory and the default one is `Kiwix` (used for all recipes we currently have in ZF).

These permissions will apply to recipes, tasks and requested tasks.

[...]

PR #1831, opened 2026-04-07
migrate ssh keys to from accounts to workers

### LINKED ISSUE #1815: Move SSH Keys management solely to "worker detail" screen
Managing worker SSH keys (view, add, remove) should be possible from worker details screen, and only from there (no more possible from user screens).

elfkuzco:
## Rationale
This PR migrates Ssh keys from the `Account` model to the `Worker` model. As a consequence, routes for performing CRUD actions on ssh keys have been moved from the account logic to the worker logic with the worker name specified in the routes. In addition, the UI is updated to move SSH key management from User detail view to Worker detail view. This lays the foundation for #1812 and #1813
Screenshot_20260407_124318
Screenshot_20260407_124136

## Changes
- migrate ssh keys from account to worker model. the DB migration script will create the same ssh key for all the workers belonging to an account.
- return ForbiddenError instead of Unauthorized error if user doesn't have permissions
- move logic for dealing with ssh keys to worker module
- move ssh key management from account detail to worker detail views
- add dependency to prevent users from performing actions on workers they do not own if they don't have desired permissions. worker owners do anything they want on their worker
- update monitor logic to fetch ssh keys for regenerating streaming keys
- update receiver code to fetch ssh keys

This closes #1815


elfkuzco:
@benoit74 , previously, permissions for `ssh_keys` were stored on the `account` Role. It will need to be moved to the `worker` entity assuming there are any custom scopes

elfkuzco:
Also, I've tried to keep the code so that workers can still function by fetching the account, then making a list of it's ssh keys from each one of it's workers. Should the zimfarm instead accept the worker's username during SSH authentication instead of the owner's username?

benoit74:
> Also, I've tried to keep the code so that workers can still function by fetching the account, then making a list of it's ssh keys from each one of it's workers. Should the zimfarm instead accept the worker's username during SSH authentication instead of the owner's username?

I think there is a bit of confusing here. 

First, the account linked to the worker is not the worker owner, it is the worker account. The worker owner will be a different account, linked to Ory, with different permissions. The worker account is necessary to describe permissions of the worker.

Let's focus on `badger1` worker for instance. It has a `badger1` account, and this worker owner will have a `jim` account which will allow him to control `badger1`, `badger2` and `badger3`.  Is that clearer? 

Second, each worker will be linked to only one account, and each account will have at most one worker (if it is a worker account). So there is no situation where we need the iterate over workers from account to fetch all fingerprints. So I do agree we should aim to simplify things and have workers send their worker_name. Which means we probably can get rid of `username` for worker accounts (just like for Ory accounts).

elfkuzco:
> Let's focus on badger1 worker for instance. It has a badger1 account, and this worker owner will have a jim account which will allow him to control badger1, badger2 and badger3. Is that clearer?

Yup. There will equally be `badger2` and `badger3` accounts. But I'm not sure how `jim` would be able to manage exclusively the `badger` workers given there's no association between `jim` and the `badger` workers to say that `jim` owns these workers. If it's based on permissions only, then, `jim` would be able to modify/delete keys belonging to different worker, say `mwoffliner1` because even though it has permissions to modify workers, it shouldn't have for everyone.

benoit74:
>  But I'm not sure how jim would be able to manage exclusively the badger workers given there's no association between jim and the badger workers to say that jim owns these workers. If it's based on permissions only, then, jim would be able to modify/delete keys belonging to different worker, say mwoffliner1 because even though it has permissions to modify workers, it shouldn't have for everyone.

I don't know either, this is the exact scope of what has to be analyzed and then implemented in https://github.com/openzim/zimfarm/issues/1819

elfkuzco:
> Second, each worker will be linked to only one account, and each account will have at most one worker (if it is a worker account). So there is no situation where we need the iterate over workers from account to fetch all fingerprints. So I do agree we should aim to simplify things and have workers send their worker_name. Which means we probably can get rid of username for worker accounts (just like for Ory accounts).

Added commit for workers to auth with their worker name (instead of account username) and updated worker README. We can already get rid of username for worker accounts. Only requirement for having a username is if user has a password and these worker accounts can opt to not have passwords depending on mode of auth

backend/src/zimfarm_backend/api/routes/workers/logic.py
```diff
@@ -122,18 +169,18 @@ def get_worker(
 
 
 @router.put(
-    "/{name}/check-in",
+    "/{worker_name}/check-in",
     dependencies=[Depends(require_permission(namespace="workers", name="create"))],
 )
 def check_in_worker(
```
benoit74, line 172: I think there is multiply things wrong here:
- the `check-in` should rather be done only by worker itself, hence needing a special permission "accessible only by worker itself"
- the fact that it can create a worker should probably be removed, we will add special API to create workers in follow-up PR + worker cannot create "itself" like it did in the past since it has now way to be authenticated since its ssh key does not yet exists in DB. I would propose to simply "update" worker characteristics at check-in now

backend/src/zimfarm_backend/api/routes/workers/logic.py
```diff
@@ -45,6 +72,26 @@
 router = APIRouter(prefix="/workers", tags=["workers"])
 
 
+def require_permission_if_not_worker_owner(namespace: str, name: str):
```
benoit74: `require_permission_if_not_worker_owner` => `require_permission_if_not_worker_itself`

backend/src/zimfarm_backend/api/routes/workers/logic.py
```diff
@@ -45,6 +72,26 @@
...
+    """
+    Ensure an account has permission to perform action on a worker that isn't theirs.
```
benoit74: `isn't theirs` => `isn't the worker itself`

backend/src/zimfarm_backend/api/routes/workers/logic.py
```diff
@@ -45,6 +72,26 @@
...
+    """
+
+    def _require_permission_if_not_worker_owner(
+        worker_name: Annotated[NotEmptyString, Path()],
+        db_session: Annotated[OrmSession, Depends(gen_dbsession)],
+        current_account: Annotated[Account, Depends(get_current_account)],
+    ):
+        worker = db_get_worker(db_session, worker_name=worker_name)
+        if worker.account_id != current_account.id:
```
benoit74: I would rather rewrite it like this for readability:

```python

if worker.account_id == current_account.id:
  return

if not check_account_permission(
  current_account, namespace=namespace, name=name
):
  raise ForbiddenError("You are not allowed to access this resource")
```

backend/src/zimfarm_backend/migrations/versions/fddc3fbad602_move_ssh_keys_from_account_to_worker.py
```diff
@@ -0,0 +1,106 @@
...
+    op.drop_constraint("fk_sshkey_account_id_account", "sshkey", type_="foreignkey")
+    op.alter_column("sshkey", "account_id", nullable=True)
+
+    # copy SSH keys from accounts to workers. For each SSH key associated with
+    # an account, create a copy for each worker that belongs to that account
+    connection = op.get_bind()
+    connection.execute(
+        sa.text(
+            """
+            INSERT INTO sshkey (id, name, fingerprint, type, key, added, worker_id)
```
benoit74: Why don't we simply set `worker_id` column and blank `account_id` when ssh key is associated with a worker? I fell like a single `UPDATE` statement would be enough.

benoit74 (changes requested):
See inline remarks.

I also feel like this is a breaking PR, i.e. workers will stop being able to communicate with the API if their worker name is not equal to their account name. It is OK with current DB (we only have an issue with the very old `homelet` worker which could even have been removed from DB), but worth to mention and double check.

No problem regarding permission change on custom roles, we do not have such a case in DB.

Finally, why do we keep both `ZIMFARM_USERNAME` and `ZIMFARM_WORKER_NAME` in `worker` side, since we only have one property to use from this PR?

backend/src/zimfarm_backend/api/routes/workers/logic.py
elfkuzco: > the check-in should rather be done only by worker itself, hence needing a special permission "accessible only by worker itself"

Should a new permission be created on the worker level or do you mean the worker account?

benoit74: I mean the worker account indeed.

backend/src/zimfarm_backend/migrations/versions/fddc3fbad602_move_ssh_keys_from_account_to_worker.py
elfkuzco: done

elfkuzco:
Gotten rid of username across the worker side too. auth now happens with worker name instead of account username. only uploader code mentions username but that has been modified to use the worker name as value. It isn't really an issue because the receiver only uses it when there is no username in it's URL. I suppose this was set many years ago to adapt to a worker uploading to a host machine with the same name as them

backend/src/zimfarm_backend/api/routes/workers/logic.py
benoit74: Sorry, looks like I was not clear at all.

Instead of `require_permission(namespace="workers", name="checkin")` I think we need `require_permission_if_not_worker_itself()` (without any args, not sure it is possible ATM) to force this API to be available only to the worker itself.

worker/contrib/zimfarm.config.example
```diff
@@ -2,18 +2,15 @@
...
 # directly at its root. Will be used as well for other Zimfarm
 # temporary data.
 ZIMFARM_ROOT=~/zimfarm
 
 ### OPTIONAL
 
-# Worker name (your choice, can be different from the username)
-ZIMFARM_WORKER_NAME="$ZIMFARM_USERNAME"
+# Worker name
+ZIMFARM_WORKER_NAME="$ZIMFARM_WORKER_NAME"
```
benoit74: This makes no sense 😉 We should rather say `myworkername`.

benoit74 (changes requested):
Two more things and I thing we are good

elfkuzco:
> Instead of require_permission(namespace="workers", name="checkin") I think we need require_permission_if_not_worker_itself() (without any args, not sure it is possible ATM) to force this API to be available only to the worker itself.

Checkin route doesn't need the route level permissions check after all as it does this
https://github.com/openzim/zimfarm/blob/7286ba015f18c4325b33a0a283469ce80676f50e/backend/src/zimfarm_backend/api/routes/workers/logic.py#L185-L187

benoit74 (approved):
LGTM, thank you

PR #1833, opened 2026-04-07
set file-task relationship cascade on delete

elfkuzco:
## Rationale
Background task to remove old tasks uses SQLAlchemy's delete construct instead of loading the task and calling `session.delete` on it. As a consequence, this throws foreign key violation errors when there are files pointing to the task. This PR updates the foreign key constraint to cascade on delete allowing tasks and their corresponding files to be deleted with/without being loaded in session

https://github.com/openzim/zimfarm/blob/b09e59b1643fdd7d6a938cc70f2e5a62b66cbd76/backend/src/zimfarm_backend/background_tasks/cancel_tasks.py#L206-L229


benoit74 (approved):
LGTM ; don't we have the same issue with other "parent-child" relationships?

elfkuzco:
> LGTM ; don't we have the same issue with other "parent-child" relationships?

It's not particularly obvious given `session.delete` is mostly used to delete objects. but I suppose adding this for other parent-child relationships is the robust and better way

elfkuzco:
> don't we have the same issue with other "parent-child" relationships?

Experimented with setting these on other table entries but it's not particularly clear what should be done when some entries are deleted. For example, if a worker is deleted, should all the SSH keys belonging to it be deleted? Should all the tasks the worker has done be deleted?  Same cascading set of questions apply to `Account` too.

I opted to leave it as it is since there's nowhere a `delete` is actually used except for the one this PR attempts to solve.

PR #1834, opened 2026-04-07
Hide workers from users list

### LINKED ISSUE #1812: Do not show workers in the UI "Users" tab

### LINKED ISSUE #1813: Allow to hide viewers from the "Users" list

elfkuzco:
## Rationale
This PR enhances the API to show/hide worker accounts while fetching accounts. In addition, it also adds the ability to hide users with viewer role

## Changes
- hide accounts with `worker` role unless `show_workers` is set to True
- add query param to hide users with viewer role from the UI

This closes #1812
This closes #1813

### DIFF
--- a/backend/src/zimfarm_backend/api/routes/accounts/logic.py
+++ b/backend/src/zimfarm_backend/api/routes/accounts/logic.py
@@ -80,7 +80,12 @@ def get_accounts(
 ) -> ListResponse[AccountSchema]:
     """Get a list of accounts"""
     results = db_get_accounts(
-        db_session, skip=params.skip, limit=params.limit, username=params.username
+        db_session,
+        skip=params.skip,
+        limit=params.limit,
+        username=params.username,
+        show_workers=params.show_workers,
+        show_viewers=params.show_viewers,
     )
     return ListResponse(
         meta=calculate_pagination_metadata(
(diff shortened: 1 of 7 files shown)


elfkuzco:
Built on top of it to hide users with viewer role instead of a different PR

benoit74 (approved):
LGTM. I wonder if the fact that by default API does filter (workers and viewers) is not a bit counter-intuitive. I'm really not sure, maybe it is ok, I let you decide, just wanted to mention that it surprised me. But I do agree that it also matches UI needs so we could argue that API should prefer to show useful data by default just like the UI.

elfkuzco:
marking as draft till https://github.com/openzim/zimfarm/pull/1831 is closed

PR #1836, opened 2026-04-08
associate worker to one account

### LINKED ISSUE #1814: Revisit worker to user association
Currently, multiple workers can be associated with a single user (account in #1811 perspective). 

This association was probably meant so that the user account could be handed over to a real human, but I feel like this is creating confusion because should the human be an editor, I don't think we should let the worker has these elevated privileges as well.

I propose that we make current `user_id` column unique, so that a worker is associated with a unique account which represent its privileges and the ssh key **this** worker is allowed to use).

I will reconfigure prod to fix WMF workers situation (they all share a single user ATM).

elfkuzco:
## Rationale
This PR adds a unique constraint to the `account_id` of the `Worker` table, thereby, restricting an account to belong to one and only one worker.

This closes #1814


benoit74 (approved):
LGTM

PR #1840, opened 2026-04-09
create worker from API/UI

### LINKED ISSUE #1816: Create worker from workers screen
The list of workers screen should allow to create a new worker.

New worker creation will require input of a unique worker name (only alphanum + dash allowed, always lowercase ; nvm about existing workers not matching this constraint) and a SSH key.

It will automatically create both the account and the worker DB entries, allowing simplification of the worker checkin logic (no need to account for missing worker entry in DB at first checkin).

Downside it that it means that `last_seen` and `last_ip` of the worker must be nullable, but it looks like it is already like this.

### LINKED ISSUE #1817: Remove worker role from user screens
Users should not be allowed to have a worker role from the UI, we need to not show this value in the UI.

elfkuzco:
## Rationale

This PR enhances the API to be able to create new workers (and their associated accounts). Accordingly, the UI has an interface to invoke this endpoint

Screenshot_20260409_061320

## Changes
- add API to create workers
- refactor logic to create ssh key by moving validation to the DB code
- refactor SSH input logic into a component and share between CreateWorkerDialog and view for editing worker
- remove `worker` role from constants (used to create role options in user update and creation views)

This closes #1816
This closes #1817

### DIFF
--- a/backend/src/zimfarm_backend/db/worker.py
+++ b/backend/src/zimfarm_backend/db/worker.py
@@ -11,16 +11,21 @@ from sqlalchemy.orm.strategy_options import selectinload
 from zimfarm_backend.common import getnow, to_naive_utc
 from zimfarm_backend.common.constants import WORKER_OFFLINE_DELAY_DURATION
 from zimfarm_backend.common.enums import TaskStatus
+from zimfarm_backend.common.roles import RoleEnum
 from zimfarm_backend.common.schemas import BaseModel
-from zimfarm_backend.common.schemas.models import DockerImageVersionSchema
+from zimfarm_backend.common.schemas.models import DockerImageVersionSchema, KeySchema
 from zimfarm_backend.common.schemas.orms import (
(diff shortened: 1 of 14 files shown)


backend/src/zimfarm_backend/db/worker.py
```diff
@@ -275,3 +280,34 @@ def check_in_worker(
...
+    ssh_key: KeySchema,
+):
+    """Create a new worker.
+
+    A new account bearing the same name as the worker will be created. This new
+    account will be created with the "worker" role.
+    """
+
+    account = create_account(session, display_name=worker_name, role=RoleEnum.WORKER)
+    check_in_worker(
```
benoit74: To be reworked following previous PR rework

benoit74 (changes requested):
LGTM, besides one rework needed due to previous PR change.

benoit74 (approved):
LGTM

PR #1843, opened 2026-04-10
Reorganize user details screen

### LINKED ISSUE #1818: Reorganize user details screen
User edit screen can be reorganized, especially once we have get rid of worker ssh keys in #1815 

We first need to display a single button for all edits. The fact we have a distinct API for passwords is a technical detail which must not be exposed in the UI.

Then we need to display things in a sensible order:
- first display name + role (these are mandatory)
- then username + password (explaining this is for local authentication)
- then IDP sub (explaining this is the ID in external auth provider)
- then custom role details is role is "custom"

All these (display name, role, username, the fact that a password is set or not and the IDP sub) must be shown as well in the "profile" tab (maybe to rename "View" btw).

elfkuzco:
## Rationale
This PR re-organizes the user details screen by including additional information in the view tab
Screenshot_20260414_100014

Screenshot_20260410_085935

## Changes
- show display name, role and authentication settings in user details view
- use single update button to update both user basic information and password
- require current password when it's the same that's changing the password

This closes #1818

### DIFF
--- a/frontend-ui/src/views/UserView.vue
+++ b/frontend-ui/src/views/UserView.vue
@@ -5,14 +5,19 @@
 
 <template>
   <v-container>
+    <ErrorMessage
+      v-if="!canReadAccounts"
+      message="You don't have permission to view user accounts."
+    />
+
     <!-- Loading state when data hasn't been loaded yet -->
-    <div v-if="!dataLoaded && loadingStore.isLoading" class="text-center pa-8">
+    <div v-if="canReadAccounts && !dataLoaded && loadingStore.isLoading" class="text-center pa-8">
       <v-progress-circular indeterminate size="64" />
       <div class="mt-4 text-body-1">{{ loadingStore.loadingText }}</div>
     </div>
 
     <!-- Content only shown when data is loaded -->
-    <div v-if="dataLoaded">
+    <div v-if="canReadAccounts && dataLoaded">
       <v-row>
         <v-col cols="12">
           <h2 class="text-h4 mb-4">
@@ -27,9 +32,10 @@
             <!-- Tabs -->
             <v-tabs v-model="selectedTab" color="primary" class="mb-4">
               <v-tab value="details" :to="{ name: 'user-detail', params: { userId: userId } }">
-                Profile
+                View
               </v-tab>
               <v-tab
+                v-if="canUpdateAccounts"
                 value="edit"
                 :to="{
                   name: 'user-detail-tab',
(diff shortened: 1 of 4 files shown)


frontend-ui/src/views/UserView.vue
```diff
@@ -57,10 +57,73 @@
...
+
+                      <!-- Authentication Card -->
+                      <v-col cols="12">
+                        <v-card variant="outlined" class="h-100">
+                          <v-card-title class="text-subtitle-1">
+                            <v-icon class="mr-2">mdi-shield-key-outline</v-icon>
+                            Authentication
+                          </v-card-title>
+                          <v-list density="compact">
+                            <v-list-item>
```
benoit74: Display this block only if `user.username` is set

frontend-ui/src/views/UserView.vue
```diff
@@ -57,10 +57,73 @@
...
+                                  <code>{{ user.role }}</code>
+                                </div>
+                              </v-list-item-title>
+                            </v-list-item>
+                          </v-list>
+                        </v-card>
+                      </v-col>
+
+                      <!-- Authentication Card -->
+                      <v-col cols="12">
```
benoit74: Display this block only if `user.username` or `user.idp_sub` is set

frontend-ui/src/views/UserView.vue
```diff
@@ -57,10 +57,73 @@
...
+                                  <span class="mr-2">Username:</span>
+                                  <code v-if="user.username">{{ user.username }}</code>
+                                  <span v-else class="text-medium-emphasis text-body-2"
+                                    >Not set</span
+                                  >
+                                </div>
+                                <div class="d-flex align-center mt-1">
+                                  <span class="mr-2">Password:</span>
+                                  <span class="text-body-2">{{
+                                    user.has_password ? 'Set' : 'Not set'
```
benoit74: Replace `Set` with `******` ; logic hangling the `Not set` is fine since this is never supposed to happen.

benoit74 (changes requested):
Thank you! 

Requiring current password was not asked for and is no convenient since changing password through this screen is usually done by an admin who has no clue about current user password. What was the logic behind this change? Does it means that I've overlooked the fact that the same component is also used in the flow where user want to modify his own password? If this is the reason, then we need a different component. At API level, normal userx are not able to modify anything beyond their password (or did I missed something?).

I also think current UI does not allow to delete a user password while it should. We can for instance display a constant string `******` by default and whenever it is changed (blanked or set to something else) then we set the user password (or blank it).

Layout of edit form is also not really organized since it mixes local auth with idp (contrary to original proposition). I would prefer to keep original proposition and have clearer separation between local auth and idp.

In edit side, we should display local auth and idp_sub only when it is already set (so that admin can blank them) or auth mode matches. This will avoid to display local auth settings when we accept only oauth, avoiding confusions for the admins. On "read" side, I'm fine with displaying whatever we have since we should not hide potential bad / unexpected configurations.

elfkuzco:
> What was the logic behind this change? Does it means that I've overlooked the fact that the same component is also used in the flow where user want to modify his own password? If this is the reason, then we need a different component. At API level, normal userx are not able to modify anything beyond their password (or did I missed something?).

Was trying to replicate API logic for updating a password but API already has restrictions preventing user from modifying their own details.

elfkuzco:
Forgot there's already a view for changing password. simplified edit logic now.

benoit74 (approved):
LGTM, nice catch regarding profile operation permissions, indeed even "self" should be restricted from most operations

PR #1844, opened 2026-04-10
require specific fields during account creation based on auth mode

### LINKED ISSUE #1820: Revisit user creation screen
When creating a user, we should ask for his display name and role and based on current UI kind of auth either:
- prompt for idp-sub
- prompt for username (by default identical to display name, indicate this is what the user will use to login) and password (with same behavior of default password generated, and button to regenerate)

elfkuzco:
## Rationale
This PR enhances the user creation screen to require speicifc fields depending on auth mode. It also enhances the API request schema to restrict the range of input values.
Screenshot_20260410_124340

## Changes
- require idp_sub when only oauth mode is configured and make optional when both auth modes configured
- require username and password when only local mode is configured and make optional when both auth modes configured 
- make username mandatory when password is set irrespective of configured mode
- set display name from username when it is missing

This closes #1820

### DIFF
--- a/backend/src/zimfarm_backend/api/routes/accounts/logic.py
+++ b/backend/src/zimfarm_backend/api/routes/accounts/logic.py
@@ -1,5 +1,5 @@
 from http import HTTPStatus
-from typing import Annotated
+from typing import Annotated, cast
 
 from fastapi import APIRouter, Depends, Path, Query, Response
 from sqlalchemy.orm import Session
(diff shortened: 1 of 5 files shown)


benoit74 (approved):
LGTM.

For readability, we should require `username` to contain only a limited set of characters: lowercase letters (a-z) + digits (0-9) + dash. Consequence is that the logic to transform display name into username needs to be updated to make this automatic transformation, replacing accented and special characters with their nearest equivalent. It should also automatically cope with collisions (automatically adding a digit suffix, and incrementing it as needed until there is no collision). I'm ok if you implement this in a follow-up issue / PR.

PR #1845, opened 2026-04-10
authenticate workers with ssh bearer tokens

### LINKED ISSUE #1826: Completely stop to rely on JWT for workers
Despite authenticating with SSH keys, workers also finally get an access token and a refresh token. 

I would like to change this in light of Ory move where we would like to block any access with locally crafted JWT (for security reasons).

We need to:
- have workers always send their `X-SSHAuth-Message` and `X-SSHAuth-Signature` in all API calls
- accept this **directly** in all routes (in place of a JWT)

I would like that we consider the possibility to move from `X-SSHAuth-Message` and `X-SSHAuth-Signature` to a regular bearer where we encoding message (which is already encoded as username.timestamp) + signature in the bearer: `Authorization: Bearer {username}.{timestamp}.{signature}`. I feel like it would make the situation way easier to understand and simplify the codebase.

I propose to not consider caching / reusing message and signature since this is fast enough to do that we should not care.

@elfkuzco can you study the feasibility of this and come back?

elfkuzco:
## Rationale
This PR introduces a new form of bearer tokens. These tokens are generated with a workers' private key and is a drop-in replacement for calls to the `/auth/ssh-authorize` endpoint. The tokens are of the format `{worker_name}.{timestamp_str}.{signature}` where `timestamp_str` is a naive datetime string in isoformat (without milliseconds and other smaller units). 

## Changes
- mark `/auth/ssh-authorize` endpoint as deprecated (new issue to be created to remove it once all worker managers have been updated if PR is approved)
- add new decoder for decoding these SSH tokens. Usually the first decoder to be run as it fails fast and worker calls are likely more important
- refactor dependency `get_jwt_claims_or_none` to incorporate new decoder signature

This closes #1826

### DIFF
--- a/backend/src/zimfarm_backend/api/routes/auth/logic.py
+++ b/backend/src/zimfarm_backend/api/routes/auth/logic.py
@@ -116,8 +116,8 @@ def refresh_access_token(
     return _refresh_access_token(db_session, request.refresh_token, response)
 
 
-@router.post("/ssh-authorize")
-def authenticate_worker_with_ssh_keys(
+@router.post("/ssh-authorize", deprecated=True)
+def authenticate_account_with_ssh_keys(
     db_session: Annotated[OrmSession, Depends(gen_dbsession)],
     x_sshauth_message: Annotated[
         str,
--- a/backend/src/zimfarm_backend/api/routes/dependencies.py
+++ b/backend/src/zimfarm_backend/api/routes/dependencies.py
@@ -27,38 +27,48 @@ AuthorizationCredentials = Annotated[
 ]
 
 
-def get_jwt_claims_or_none(
-    authorization: AuthorizationCredentials,
-) -> JWTClaims | None:
-    """
-    Get the JWT claims or None if the account is not authenticated.
+def get_jwt_claims_or_none_with_session(
+    session_type: Literal["auto", "manual"] = "auto",
+):
+    def _get_jwt_claims_or_none(
+        authorization: AuthorizationCredentials,
+        session: Annotated[
+            OrmSession,
+            Depends(gen_dbsession if session_type == "auto" else gen_manual_dbsession),
+        ],
+    ) -> JWTClaims | None:
+        """
+        Get the JWT claims or None if the account is not authenticated.
+        """
 
-    Tries to decode as Zimfarm token first, then as Kiwix token.
-    """
-    if authorization is None:
-        return None
-    token = authorization.credentials
+        if authorization is None:
+            return None
+        token = authorization.credentials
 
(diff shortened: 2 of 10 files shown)


benoit74 (commented):
Isn't this breaking the logic which allows to have a clear error message? At least I don't get how it works anymore. For instance I feel like if a worker sends an expired SSH token, then it will only get a generic message of the OAuthSessionTokenDecoder saying something about the fact it is not a valid JWT. Or do I miss something?

I think the problem might already have been there before with the various JWT decoders but a bit hidden by the fact we usually have only one configured + they all accept JWT.

elfkuzco:
> Isn't this breaking the logic which allows to have a clear error message? At least I don't get how it works anymore. For instance I feel like if a worker sends an expired SSH token, then it will only get a generic message of the OAuthSessionTokenDecoder saying something about the fact it is not a valid JWT. Or do I miss something?

Do you mean the issue with the timestamp being out of date? The token decoders raise a ValueError and the functiont that decodes a token watches for a ValueError. Also, we have app-level handlers for ValueError messages which extract the `args[0]` which is the original message from the exception and translate it into an HTTP error

benoit74:
> Do you mean the issue with the timestamp being out of date? The token decoders raise a ValueError and the functiont that decodes a token watches for a ValueError. Also, we have app-level handlers for ValueError messages which extract the args[0] which is the original message from the exception and translate it into an HTTP error

Yes, I'm referring to the sample case of the timestamp being out of date. 

I got the logic around ValueError raising and later handling, but don't we have two decoders which raise a ValueError, and only last one wins with `exc_cls` logic? (having only first one to win would not really solve this issue, only move the problem).

elfkuzco:
Indeed, the exception class that is raised is the most recent one and not the first one and given the nature of the decoder chains, the original exception would be lost. Solution would either be to place the SShTokenDecoder last or only update exception class if none has been set (that is we only save the first)

benoit74:
Saving only the first does not help either because it moves the problem from SSH tokens not having proper message to OAuth tokens not having the proper message.

As discussed live, I think we need a finer logic in decoders to better handle the situation by implementing more "checks" in `can_decode` so that we end up with only one decode matching the `can_decode` criteria. Probably to be based on issuer for JWT, and I'm fine if this forces us to have either oidc or session decoders active (doesn't make much sense to have both, and we have only one atm).

elfkuzco:
Implemented the `can_decode` methods to determine if a token can be decoded by a specific decoder

benoit74 (approved):
LGTM, thank you

PR #1846, opened 2026-04-13
Fix permission used to display 'Archive' button in recipe details

benoit74:
Users with "only" archive permission (and not delete) should be able to access the archive (or restore) button (kinda obvious).


elfkuzco (approved):
LGTM

PR #1849, opened 2026-04-15
remove schedule routes and field names

### LINKED ISSUE #1824: Cleanup: remove `/schedules` routes and old field names
Once every users of the API has moved to new `/recipes` route and new field names, we need to remove old routes and fields.

elfkuzco:
## Changes
- remove schedule route and field names starting with schedule_

This closes #1824


benoit74 (approved):
LGTM, to be deployed only after careful review of API logs to be sure we do not use old API anymore (I will do that when you ask me for)

benoit74:
We are now good to deploy, I do not see any usage of these old routes. Cannot say for field names for sure, but I assume that any system using the new route is also using the new field name. We will fix should this prove to be wrong.

@elfkuzco please proceed with the merge

PR #1850, opened 2026-04-16
check if audience id is in list

elfkuzco:
## Changes
- check if configured audience id is in token audience id instead of direct comparison


benoit74 (approved):
LGTM

PR #1854, opened 2026-04-16
harmonize status display and duration timestamps

### LINKED ISSUE #1697: Task duration is not correct in worker overview

elfkuzco:
## Rationale
This PR aims to harmonize the way status and duration timestamps are displayed. It introduces a new component `StatusDisplay` that sets the status of the task item along with the timestamp of the status (or default to the time the status was updated at)
Screenshot_20260416_161342
Screenshot_20260416_161330
Screenshot_20260416_161320
Screenshot_20260416_154145
Screenshot_20260416_154136
Screenshot_20260416_154126
Screenshot_20260416_154122

This closes #1697

### DIFF
--- a/frontend-ui/src/components/PipelineTable.vue
+++ b/frontend-ui/src/components/PipelineTable.vue
@@ -45,54 +45,13 @@
           </router-link>
         </template>
 
-        <template #[`item.requested`]="{ item }">
-          <v-tooltip location="bottom">
-            <template #activator="{ props }">
-              <span v-bind="props">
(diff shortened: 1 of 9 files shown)


benoit74 (approved):
LGTM, thank you!

PR #1855, opened 2026-04-17
request for confirmation on action to be taken on recipes

### LINKED ISSUE #1669: Actions applying to all selected recipes/task/items
This is a follow-up of #1649 and #1650 

I understand that we have, at least for archives, actions which apply to all select items.

For the archives:
* clear selection
* restore recipes

For recipes:
* clear selection
* request recipes

The problems with current approach is:
* big buttons which are "rarely" used as usual behaviour is to deal with recipes one by one
* big buttons for actions which are relatively "dangerous"
* positioned on the right, although the select checkboxes are on the left
* unusual UI design

The usual way for this is to propose a select box, on the left (still at the top of the list) with the list of actions to apply to all select items. List items should be first "Actions" (so nothing to do)... and then the list of possible actions.

elfkuzco:
## Rationale
This PR moves buttons to perform actions on recipes to the left and requires a confirmation before proceeding with each action

Screenshot_20260417_104729
Screenshot_20260417_104719
Screenshot_20260417_104710

This closes #1669

### DIFF
--- a/frontend-ui/src/components/RecipesTable.vue
+++ b/frontend-ui/src/components/RecipesTable.vue
@@ -3,7 +3,7 @@
     <v-card v-if="!errors.length" :class="{ loading: loading }" flat>
       <v-card-title
         v-if="showSelection || $slots.actions"
-        class="d-flex flex-sm-wrap flex-grow-1 flex-column-reverse flex-sm-row align-sm-center justify-sm-end ga-2"
+        class="d-flex flex-sm-wrap flex-grow-1 flex-column-reverse flex-sm-row align-sm-center justify-sm-start ga-2"
       >
         <slot name="actions" />
         <v-btn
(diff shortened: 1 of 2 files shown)


benoit74:
Very naive question: can you remind me why we need the "Clear selections" button while we already have the "header checkbox" (the one to the left of "Name" header) which could do the same?

elfkuzco:
> an you remind me why we need the "Clear selections" button while we already have the "header checkbox" (the one to the left of "Name" header) which could do the same?

The checkbox mostly selects everything and unselects everything. The "Clear Selections" is more granular in that it only unselects what it is selected

benoit74:
> The checkbox mostly selects everything and unselects everything. The "Clear Selections" is more granular in that it only unselects what it is selected

I'm sorry but I don't get it. How "unselects what is selected" is different than "unselects everything"? Is the "Clear Selections" button really bringing more added value than confusion?

elfkuzco:
I can remove it if you want. It's not really adding much value.

benoit74 (approved):
LGTM, I feel like we will discuss about these buttons again but at least the PR is doing what has been discussed and agreed about in the issue

Issue #1362, opened 2025-09-19
Uploading files (ZIM illustrations but also any image, text file, ZIP, ...) is too cumbersome


kelson42:
Current solution:
* Choose online an illustration
* Access our drive
* Upload the file with the right name/directory
* Come back to the Zimfarm
* Link the image

This has to be done for each illustration. This is too complicated.

We have to be able to manage this issue via file upload or upload via URL directly in the recipe of the Zimfarm

kelson42:
@benoit74 Remarked that this was tracked nowhere, but this is a really impairing "stupid" issue. Assigning to you for assessment.

benoit74:
Current process for setting an illustration in a recipe is more complex than what you describe, at least this is what I usually do:

- Find an illustration online
- Download on your computer
- Rename file to have a nice looking / correct name (needing to locate the file you've just downloaded on your computer)
- Access our drive UI with admin credentials
- Upload the file with the right name/directory (needing to again locate the file you've just downloaded with a different UI)
- Find / build the URL which has to be used for the uploaded file (huge source of mistakes, because URL is different than the one use to upload files)
- Come back to the Zimfarm
- Put URL in proper recipe parameter

The problem is also wider than illustrations. It is hard to create / update a custom CSS for zimit. It is hard to create / update a JSON definition and a ZIP file for nautilus. And I probably forgot other cases.

The problem is also that we do not have any authorization system on the drive, anyone with access can modify all Zimfarm files. When we will start to define permissions on who can modify a given recipe, I'm sure we will want as well to be able to restrict access to only the files of this recipe.

Corollary is also that users of zimit.kiwix.org are "on their own" when it comes to manipulating these files. Skilled persons can find a solution to upload their files somewhere, but non technical persons will probably never achieve to have a good 48x48 illustration on their ZIM (unless it exists by chance somewhere online).

The problem is also that it is hard to define and maintain a convention around file naming, location, ...

The problem is also we have no "visible" link between files on the drive and recipes, it is mostly impossible for a user to guess in which recipe a given drive file is used. This makes it impossible to clean the drive.

The problem is also that we do not force the user to use a drive URL, which causes situations where the URL used was "somewhere else" and suddenly disappear. We are not capable to run the recipe anymore. It is possible to recover by retrieving last ZIM illustration and putting in the drive, but this is causing extra errors and maintenance for "nothing". 

[...]

PR #1856, opened 2026-04-21
move blob content to database

### LINKED ISSUE #1634: Get rid of Webdav dependency
The Webdav backend has been used because it was easy to deal with for the Zimfarm users, developers and sysadmins, But it is not an appropriate storage for blobs for a machine. Either use S3 or the DB.

elfkuzco:
## Rationale
This PR moves recipe blob content to the database, effectively getting rid of webdav as a dependency (see https://github.com/openzim/zimfarm/issues/1634#issuecomment-4281286403)

## Changes
- add `content` column to Blob table. Will need to get rid of URL in a later PR after migrations have been done via the maint-scripts
- add maint-script to fetch blobs from their URLs and store the content in the DB
- update maint-script to update recipe blobs to remove webdav-related code and update blob URLs (works for existing recipes too because they do not have content in the DB)
- add endpoint to download blob

This closes #1634


elfkuzco:
`BLOB_PRIVATE_STORAGE_URL` and `BLOB_PUBLIC_URL` should generally be the same in prod or to two URLs that resolve to the same resource. It's kept different in dev as I need(ed) to see the blobs from the editor and I can't access the API on `http://backend:80`. 

If PR is approved and the maint-scripts have been run to migrate content, new migration will be added to remove the `url` column on the database.

benoit74 (commented):
I'm a bit surprised we still need both blob private and public URLs. This was necessary for WebDav because file operations could only be done on the private URL. Now there is only one URL per blob to use in recipe flags, the one of the API. And since it is the one of the API, I don't get why we still need `BLOB_PUBLIC_STORAGE_URL` and `BLOB_PRIVATE_STORAGE_URL` env variables. I feel like only `API_ENDPOINT` is sufficient, the rest can be computed (and reduces risk of missconfiguration IMHO).

Can you also develop how we will handle the transition to this new codebase? Which maint-script needs to be run, how existing DB configuration will still be fine until maint-script completed, etc ...

elfkuzco:
> I feel like only API_ENDPOINT is sufficient, the rest can be computed (and reduces risk of missconfiguration IMHO).

Okay. Will implement this

> Can you also develop how we will handle the transition to this new codebase? Which maint-script needs to be run, how existing DB configuration will still be fine until maint-script completed, etc ...

The steps will be as follows:
- run migrations to create new column
- run `maint-scripts/fetch_blob_raw_content.py` : This will update the blobs in the DB by setting their content to the data pointed to by the `url` of the corresponding row
- run `maint-scripts/update_recipe_blobs.py`: This will modify flags of existing recipes in the DB that are blobs. It has filter rules for omitting specific blobs from being updated. It first fetches all the blob in the DB for that recipe flag and checks if it's URL matches the newer one. If it doesn't, it fetches the blob, creates it in the DB and updates the flag to point to the new URL.

benoit74 (commented):
I'm sorry but I still don't get how `fetch_blob_raw_content` and `update_recipe_blobs` will cooperate.

As far as I've understood, `fetch_blob_raw_content` will only update `content` column, but keep recipe flags untouched.

And in `update_recipe_blobs`, we compare recipe flag URL with blob URL which will always be different because flag URL is from webdav while blob URL is now on API. We hence always download the flag content from webdav (or anything else if not yet migrated appropriately) and create a new blob (or update existing one if checksum matches) with freshly downloaded content.

I'm mostly fine with `update_recipe_blobs` logic because it is probably the most conservative one, but then do we really need `fetch_blob_raw_content` logic?

I'm also curious about what will happen should blob checksum not match. We will create a new blob, hence having two blobs for a single recipe ID + flag Name. But only one blob is really used. What will delete the useless blob? Do we keep it because it "might" be used for historical reasons? I'm a bit worried we end-up filling the DB with unused duplicates.

elfkuzco:
> I'm mostly fine with update_recipe_blobs logic because it is probably the most conservative one, but then do we really need fetch_blob_raw_content logic?

> I'm also curious about what will happen should blob checksum not match. We will create a new blob, hence having two blobs for a single recipe ID + flag Name. But only one blob is really used. What will delete the useless blob? Do we keep it because it "might" be used for historical reasons? I'm a bit worried we end-up filling the DB with unused duplicates.

I decided to keep the older blobs because of https://github.com/openzim/zimfarm/issues/1362#issuecomment-3588407141:

"UI should allow to see older versions (via a dedicated tooling for the pictures, or via a link to Zimfarm diff feature for text content)"

For now, we don't have the UI tool to view the older versions (other than in recipe history) but I imagine this comment means a dropdown which will be used to scroll through the versions of the blob. Perhaps, user wants to go back to a version.

So, essentially, it's kept for historical reasons. Only when a recipe is deleted does the `recipe_id` get set to NULL at the database level and then, the background job that deletes orphaned blobs removes them

elfkuzco:
> I'm also curious about what will happen should blob checksum not match.

If the blob exists on webdav and it's in the DB, then, there is no reason for the checksums to differ. Unless the file has been modified or the checksum in the DB isn't correct

benoit74:
OK thank you, got it

dev/docker-compose.yml
```diff
@@ -45,10 +45,7 @@ services:
       # username and password of admin user
       INIT_USERNAME: admin
       INIT_PASSWORD: admin
-      BLOB_PRIVATE_STORAGE_URL: http://blob-storage
-      BLOB_PUBLIC_STORAGE_URL: http://localhost:8003
-      BLOB_STORAGE_USERNAME: zimfarm
-      BLOB_STORAGE_PASSWORD: zimpass
+      API_ENDPOINT: http://backend:80/v2 # change this to an ngrok URL if you want to view the files from browser
```
benoit74: Why not using `http://localhost:8000/v2`? This would allow developers to see all blobs, run Zimfarm tasks, ... What do I miss?

elfkuzco: It won't be able to run zimfarm tasks because the scraper container's `localhost:8000` is pointing to it's own network and blob won't exist

elfkuzco: it's only a URL that can be accessed by both task containers and browser that can be used. Given task execution is of higher precedence than viewing blobs in the browser, I opted to set it to the backend container name instead

benoit74: OK, got it.

I'm however not sure I agree about the precedence, since it means devs can't test many parts of the UI without setting an ngrok URL **and** running DB migration. 

In general devs will work on a prod DB backup and recipe flags will anyway be set to prod URL. So tasks should run fine (fetching the blob content from prod, but in general it will work). 

Being able for a dev to test all blobs edition, diff, uploading them again and seeing them in the UI and so on in a convenient way seems quite important to me.

Do I miss something again?

elfkuzco: No.

benoit74: OK, then please prefer to use localhost URL if we agree it will make UI work correctly and is quite unlikely to block tasks execution for devs using a prod DB backup.

benoit74 (approved):
LGTM, thank you

PR #1857, opened 2026-04-21
add and edit images directly in recipe editor

elfkuzco:
## Rationale
This PR enhances the blob editor to allow the user to add/edit images directly in the recipe editor instead of a dialog box.

Screenshot_20260421_143650
Screenshot_20260421_143636
Screenshot_20260421_143622


elfkuzco:
@benoit74 , couldn't easily implement the logic to store original URL (https://github.com/openzim/zimfarm/issues/1362#issuecomment-4281261975)

> This field can be freely edited and is automatically populated when the user copy-paste a URL (append if any text is already present in the field).

The user can input the original URL in the comments but automatically populating it isn't particularly straightforward. For example, if I delete a blob and paste in another multiple times, the URL will keep getting appended those same number of times. Feel the user should be allowed to manually add whatever they want in the comments

benoit74 (approved):
LGTM, I feel like we are definitely moving in the right direction, thank you

PR #1858, opened 2026-04-22
add and edit text blobs directly in recipe editor

elfkuzco:
## Rationale
This PR enhances the recipe editor to be able to add/edit text blobs directly without a dialog.

Screenshot_20260422_103742
Screenshot_20260422_103354
Screenshot_20260422_103342

## Changes
- refactor the inline image editor and text editor to track their different checksums internally as there can be mutliple blob editors on a recipe page
- move most of the core logic to the various editor components instead of the generic blob editor


benoit74 (approved):
LGTM, well done!

PR #1863, opened 2026-04-23
allow users to input external blob URLs

### LINKED ISSUE #1859: mwoffliner `articleList` should not always be a blob
For mwoffliner, the `articleList` argument was mostly always pointing to a URL which content is dynamic (like either dynamically generated at each call or updated on a regular schedule like once a month

With recent migration to blobs, we've migrated this dynamic content to a blob, and this was hence a mistake on my side since it means content is now "static"

I'm not sure how we should solve this, since it is a rare case where we can benefit from the inline text editor for situations where a user has a static list to input, but we might also want to input a dynamic URL. 

An extremely lean UX would be that when user input a URL in the text area (supposed to contain the blob content), then the UI detects the situation and just use this URL instead of the blob URL. But I'm not sure this would be intelligible.

Otherwise we can also add a radio button to choose between "manual" and "dynamic" content and adapt UI accordingly.

It obviously means we need to dig into recipe history or an archive backup in order to restore dynamic URLs. Basically every `articleList` parameter which was not pointing to the drive is probably a dynamic URL. Probably worth to double-check the list manually should we have static lists at other places.

@elfkuzco WDYT?

elfkuzco:
## Rationale
This PR enhances the blob editor to allow editors to input external blob URLs if the flag in the offliner definition supports it. For such flags, the blob editor can be viewed in local/remote mode with remote allowing the user to input an external URL and `local` for the other modes of input already supported by the blob editor.
Screenshot_20260512_165838

## Changes
- add property `allow_remote_url` to the flag schema. should only be set for flags that are blob types
- update the maintenance script to modify mwoffliner's `articleListToIgnore` and `articleList` to be txt blobs
- add script to modify any property belonging to any offliner flag. will be used to update both flags by setting their `allow_remote_url` to True.
- force remote mode when we allow remote mode but blob editor fails to load content because default mode is local

This closes #1859

### DIFF
--- a/backend/maint-scripts/update_offliner_definition_blobs.py
+++ b/backend/maint-scripts/update_offliner_definition_blobs.py
@@ -12,14 +12,18 @@ from zimfarm_backend.db.models import OfflinerDefinition
 from zimfarm_backend.db.offliner_definition import create_offliner_definition_schema
 
 FLAG_MAPPINGS: dict[
-    str, list[dict[str, Literal["image", "css", "html", "illustration"]]]
+    str, list[dict[str, Literal["image", "css", "html", "illustration", "txt"]]]
 ] = {
     "devdocs": [{"logo_format": "image"}],
     "freecodecamp": [{"illustration": "illustration"}],
     "ifixit": [{"icon": "illustration"}],
     "kolibri": [{"favicon": "illustration"}, {"css": "css"}, {"about": "html"}],
     "mindtouch": [{"illustration_url": "illustration"}],
-    "mwoffliner": [{"customZimFavicon": "illustration"}],
+    "mwoffliner": [
+        {"customZimFavicon": "illustration"},
+        {"articleList": "txt"},
+        {"articleListToIgnore": "txt"},
+    ],
     "nautilus": [
         {"main_logo": "image"},
         {"secondary_logo": "image"},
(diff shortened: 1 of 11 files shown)


kelson42:
I don't understand why we need an input! AFAIK I have written amny time that this was not requested/necessary.

benoit74 (commented):
I don't think we want or need this for images / illustrations. We should probably rather have this option only on parameters where it makes sense. And mwoffliner `--articleList` is maybe the only one for now, other text fields probably only wants a blob locally managed.

We should discuss it live with @kelson42 before we move this forward, do not make code changes reacting to this comment until we have an agreement.

kelson42:
After discussion with @benoit74, here would be my recommendations:
* UI with radioboxes look good
* Please upgrade the PR description with proper screenshots (for article list)
* Radioboxes label should be "Local" and "Remote"

benoit74:
> Radioboxes label should be "Local" and "Remote"

Please also update "technical" values (not only label) to `local` and `remote` for consistency.

And properly link #1859 and fully solve this issue in this PR, since it is the reason we are implementing this change (probably means we need a maintenance script to update mwoffliner offliner definition)

elfkuzco:
I have updated the PR. Had to rebase and force push because of outdated migrations making it difficult to test locally.

@benoit74, you will need to run the scripts in the following order:
```sh
# Update the existing definitions to migrate articleList and articleListToIgnore as txt blobs
python3 update_offliner_definition_blobs.py

# Set the allow_remote_url property for the articleList flag
python3 update_offliner_definition_flag.py -o mwoffliner -f articleList -p allow_remote_url -v true --type boolean

# Set the allow_remote_url property for the articleListToIgnore flag
python3 update_offliner_definition_flag.py -o mwoffliner -f articleListToIgnore -p allow_remote_url -v true --type boolean
```

benoit74 (approved):
LGTM, thank you. I will run the maintenance scripts once you've merged this.

PR #1869, opened 2026-04-23
notify CMS when both check result and filename have been set

### LINKED ISSUE #1866: CMS notification often misses zimcheck_url value
CMS notification often misses zimcheck_url value, causing CMS to refuse the notification (this is expected behavior)

See e.g. https://cms.openzim.org/zimfarm-notification/9da5fe75-d718-1fc5-ad01-a608aba5dfbd

Image

Deleting the notification in CMS and update Zimfarm DB file status so that notification is resent causes the notification to contain expected zimcheck_url value.

This is hence only a transitory issue, probably not something linked to a misconfiguration.

I quickly checked code and don't see why we would send the notification too early, but I probably miss something.

elfkuzco:
## Rationale
Worker code updates zim check results and zim upload filename in two requests with the later occassionally taking long (#1456)

This PR enhances the background task to only notify CMS when both have successfully completed as worker sets filename when upload completes.

This fixes #1866

### DIFF
--- a/backend/src/zimfarm_backend/db/files.py
+++ b/backend/src/zimfarm_backend/db/files.py
@@ -1,7 +1,7 @@
 import datetime
 from uuid import UUID
 
-from sqlalchemy import func, or_, select
+from sqlalchemy import and_, func, or_, select
 from sqlalchemy.orm import Session as OrmSession
 
 from zimfarm_backend.common.schemas import BaseModel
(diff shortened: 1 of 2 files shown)


elfkuzco:
@benoit74 , are there any files who have not been notified that have only one of check_result and filename? given the fix for #1456 was implemented later on, some older `File` objects will likely not have the `check_filename`

benoit74 (approved):
The difference between an `OR` and an `AND` ... AND.OR.OR.AND.OR 🤣 

LGTM, thank you

elfkuzco:
@benoit74, I wonder if we should use this opportunity to add conditions to further tighten the query given worker operations happen async. I propose we add conditions for:
- ensuring `uploaded_timestamp` is also not None (which means file has been uploaded)

What do you think?

benoit74:
Indeed, you're right

PR #1871, opened 2026-04-24
hide download button when zim urls is empty

### LINKED ISSUE #1867: Even with CMS deployed, ZIM URLs in tasks do not come from CMS
I've set `INFORM_CMS` in the backend as expected, which causes CMS to be called when getting a task details (as expected), but then the ZIM URLs in the task we receive are still old ones.

Looking quickly at the code it looks like we are setting URLs received from CMS in a temporary object which does not make it to the final API result.

elfkuzco:
## Rationale
This PR implements the logic to hide dowload buttons when zim urls is an empty list
## Changes
- make zim urls nullable
- hide download links when zim urls is an empty list. Empty list indicates call wen through to CMS but returned an empty response
- show default download link when zim urls is null. This happens only when `INFORM_CMS` is set to `false` as the call to CMS API never goes through

This closes #1867


benoit74 (approved):
Oh, indeed, nice catch. LGTM

PR #1872, opened 2026-04-24
allow blobs that cannot be loaded to be deleted

### LINKED ISSUE #1862: Enhance handling of failed image loading
When we open a recipe to edit with an image blob which fails to load, user is kinda stuck:

Image

We should:
- display URL which failed to load for more clarity
- display "delete" button so that user can at least remove the bad URL

This is not supposed to happen, but better to not be stuck in such a situation

elfkuzco:
## Rationale

Screenshot_20260424_094037

Screenshot_20260424_102153

## Changes
- show error message when blob cannot be loaded
- show delete button in preview zone to allow user to delete blob which cannot be loaded
- hide blob comments when blob cannot be loaded

This closes #1862


benoit74 (approved):
LGTM

PR #1879, opened 2026-04-27
add margin to info and detail views

### LINKED ISSUE #1875: Lack of margin
Text should not stick to the border like this
Image

elfkuzco:
## Rationale
Screenshot_20260427_095953

## Changes
- add a margin to the info/details views of tasks/recipes. side margins are 4 units with top and bottom set to 2 as each row within the component adds a `py-2`.
- remove horizontal scrollbar from scraper stderr and stdout zone since text wraps around

This closes #1875


benoit74 (approved):
LGTM

PR #1882, opened 2026-04-28
Fix wrong formatting of insufficient CPU resources

benoit74:
Message displayed when CPU consumption is not OK was not using proper units

### DIFF
--- a/backend/src/zimfarm_backend/db/requested_task.py
+++ b/backend/src/zimfarm_backend/db/requested_task.py
@@ -99,8 +99,7 @@ def _resource_mismatch_message(
     mismatched_message: list[str] = []
     if worker.resources.available.cpu < resource.cpu:
         mismatched_message.append(
-            f"cpu: required={format_size(resource.cpu, binary=True)}, "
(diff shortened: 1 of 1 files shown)


benoit74:
@elfkuzco this is not the first time that I have `black` and `ruff` disagree on what needs to be done. Do you experience the same issue? Should we try to update both tools to a more recent version?

elfkuzco:
Yeah. It happens locally for me and I just pass `--no-verify` when they no longer show unsealed changes

benoit74:
`--no-verify` is what I did. But it does not help here because `black` reformats in one way, and then `ruff` in another one. And hence the CI fails because when it runs `black` it wants to reformat again. Do I miss something?

elfkuzco:
Um, I usually, add the files and then, they conflict with the new file showing as unstaged. Then, I add again and it conflicts but no file is unstaged. That's when I use `--no-verify`. Thought it was a local issue because CI always passes and I blamed it on my environment.

benoit74:
OK, so we have a different issue here.

elfkuzco (approved):
LGTM

PR #1888, opened 2026-04-29
fix populate_zim_urls test

elfkuzco:
## Changes
- fix comparison to empty list in tests for `populate_zim_urls`

This closes #1883

PR #1890, opened 2026-04-30
refactor grapheme string validation for offliner flags

### LINKED ISSUE #1881: Zimfarm complains string is 41 graphemes where it has only 24
I don't get why this recipe (and few others) have an issue with their ZIM title which look good to me (and to the UI way of counting graphemes, this error comes from the backend)

https://farm.openzim.org/recipes/wikipedia_gu_medicine/edit

Image

elfkuzco:
## Rationale
The `NotEmptyString` is the base str class used for building up most of the types of the different offliner fields. However, even though we initially recast it to `GraphemeStr` which has a custom `__len__`, pydantic min/max length validation still do not use this hook. See https://github.com/pydantic/pydantic/discussions/13131#discussioncomment-16785225 for more information. 

The solution is to inject the grapheme string validation function while building the flags and still maintain the base class str annotation for type-checkers. However, min/max length are no longer done by pydantic via the `Field` constructor.

## Changes
- remove different types which were defined when the offliners were still defined in one file per offliner type
- add grapheme string validation as a validator function
- rename max_length, min_length, relaxed_min_length and relaxed_max_length to each end with `_graphemes`
- add maintenace script to rename the fields for the flags in the database

This fixes #1881

### DIFF
--- a/backend/maint-scripts/update_offliner_definition_graphemes.py
+++ b/backend/maint-scripts/update_offliner_definition_graphemes.py
@@ -1,0 +1,91 @@
+#!/usr/bin/env python3
+
+import argparse
+from copy import deepcopy
+from typing import Any
+
+from sqlalchemy import select
+from sqlalchemy.orm.attributes import flag_modified
+
+from zimfarm_backend import logger
+from zimfarm_backend.db import Session
+from zimfarm_backend.db.models import OfflinerDefinition
+from zimfarm_backend.db.offliner_definition import create_offliner_definition_schema
+
+
+def update_flag(
+    flags_dict: dict[str, Any], *, flag_name: str, old_key: str, new_key: str
+) -> bool:
+    flag_changed = False
+
+    if old_key in flags_dict[flag_name]:
+        flags_dict[flag_name][new_key] = flags_dict[flag_name][old_key]
+        del flags_dict[flag_name][old_key]
+        flag_changed = True
+        if flag_changed:
+            logger.info(
+                f"Schema modified for {offliner_definition.offliner} "
+                f"v{offliner_definition.version} flag '{flag_name}'"
+            )
+
+    return flag_changed
+
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser(
+        description="""
+Update offliner definition schemas: minLength -> minGraphemes, maxLength -> maxGraphemes
+""",
+    )
+
+    args = parser.parse_args()
+
+    with Session.begin() as session:
+        definitions = session.execute(select(OfflinerDefinition)).scalars().all()
+        for offliner_definition in definitions:
+            logger.info(
+                f"Processing {offliner_definition.offliner} version "
+                f"{offliner_definition.version}"
+            )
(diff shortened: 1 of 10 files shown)


elfkuzco:
Failing tests have nothing to do with the PR and are fixed in #1888

elfkuzco:
@benoit74 , prior to opening this PR, I created a [discussion on Pydantic](https://github.com/pydantic/pydantic/discussions/13131) on how to go about this cleanly and in the thread, one of my initial hack was to replace the underlying type to `Any` instead of `str` before I tried this approach that is submitted in this PR. 

The strange thing with the PR approach is that while these recipes are validated and everything works in the web and I can see that my `__len__` is being called. Same behaviour when I use the maintenance script to validate all recipes. But when I try to import the `NotEmptyString` in a REPL and define the model in the example discussion, it fails. I can't exactly explain why this happens but I am thinking of reverting to the one in the thread since it's explainable. WDYT?

benoit74:
>  I can't exactly explain why this happens but I am thinking of reverting to the one in the thread since it's explainable. WDYT?

Not sure what is "the one in the thread".

elfkuzco:
> Not sure what is "the one in the thread".

Using `Any` instead of `str` as the underlying type. See https://github.com/pydantic/pydantic/discussions/13131#discussioncomment-16785225

It turns out that the PR works because we use `create_model` to build the models at runtime (see https://github.com/pydantic/pydantic/discussions/13131#discussioncomment-16792457). But given the `NotEmptyString` is used in other places that aren't just offliner flags, I am leaning towards using `Any` since it makes sense that Pydantic hooks into the custom `__len__` and we don't worry about external clients using the openAPI schema

benoit74:
Why not using Option 2 of https://github.com/pydantic/pydantic/discussions/13131#discussioncomment-16792457 ; it is a bit more work but I feel like it would be more stable, or do I miss something? It could even be an opportunity to fix the bug on "list-of-string". I would propose to even stop using "minLength" and "maxLength" since this is misleading, but rather use "minGraphemes" and "maxGraphemes". I'm quite sure it will also allow to customize a better message than "Value should have at most xx items after validation" which is quite "blurry" (saying "Value should have at most xx graphemes"  would be superior).

Or is this just way too much work?

elfkuzco:
> Or is this just way too much work?

No.

> I would propose to even stop using "minLength" and "maxLength" since this is misleading, but rather use "minGraphemes" and "maxGraphemes".

This would apply at the various `offliner-definition.json` files or should it just be an alias to keep existing definitions compatible

benoit74:
> This would apply at the various offliner-definition.json files or should it just be an alias to keep existing definitions compatible

Good point, thank you. I prefer we update all definition versions to use these new words for clarity, even if it comes at the cost of having to prepare PRs on all scrapers repo, sorry for that. Plus a maintenance script to update DB ...

elfkuzco:
Fixed it and also a bit of updating in the UI. This would also require updating zimit-frontend

benoit74 (approved):
LGTM

PR #1891, opened 2026-05-04
show button to configure 2FA on Ory

### LINKED ISSUE #1876: Stuck at Zimfarm level if no 2FA configured
I just have a normal SSO account without MFA.

When I connect to the Zimfarm, I get this:

Image

So far so good, I click on back to sign-in (only choice)

Image

Then I click sign-in, and then I'm back with my error.

We should propose a way to setup the MFA from the error.

elfkuzco:
## Rationale
Screenshot_20260504_101933

## Changes
- add button to configure 2FA on Ory
- hide button to sign in when error is 2FA error

This closes #1876

### DIFF
--- a/frontend-ui/src/views/OAuthCallbackView.vue
+++ b/frontend-ui/src/views/OAuthCallbackView.vue
@@ -26,7 +26,18 @@
               <v-alert type="error" variant="tonal" class="mb-4 text-left">
                 {{ error }}
               </v-alert>
-              <v-btn color="primary" :to="{ name: 'sign-in' }"> Back to Sign In </v-btn>
+              <v-btn v-if="!is2FAError" color="primary" :to="{ name: 'sign-in' }">
+                Back to Sign In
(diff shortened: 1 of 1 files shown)


frontend-ui/src/views/OAuthCallbackView.vue
```diff
@@ -37,15 +48,33 @@
...
+  throw new Error('Config is not defined')
+}
 
 const loading = ref(true)
 const error = ref<string | null>(null)
 
+const is2FAError = computed(() => {
+  return (
+    error.value?.startsWith(
+      '2FA authentication is mandatory on Zimfarm but it looks like you only have one setup on Ory',
```
benoit74: I recommend to rely only on `2FA authentication is mandatory on Zimfarm` (shorter but sufficient and less fragile should we change the error and/or auth provider)

benoit74 (approved):
LGTM. See minimal comment

PR #1893, opened 2026-05-05
Add more disk requirement options

benoit74:
https://farm.openzim.org/recipes/stackoverflow.com_en needs about 800G disk space, this is not configurable through the UI ATM, hence adding new disk options is useful.


elfkuzco (approved):
LGTM

PR #1907, opened 2026-05-18
create default durations for recipes

### LINKED ISSUE #1906: No default duration found for recipe
I have a zimfarm requested task which refuses to start due to `No default duration found for recipe xxxx`:

https://farm.zimit.kiwix.org/pipeline?filter=todo

Image

@elfkuzco is this a bug in Zimfarm recipe itself (these are created by wp1? note this specific recipe is "pretty" old) or on Zimfarm codebase?

elfkuzco:
## Rationale
This PR adds a maintenance script to search for all recipes without a default duration and create one for them

This fixes #1906

### DIFF
--- a/backend/maint-scripts/create_default_recipe_durations.py
+++ b/backend/maint-scripts/create_default_recipe_durations.py
@@ -1,0 +1,48 @@
+#!/usr/bin/env python3
+
+"""
+Create default recipe durations for recipes that do not have a default duration
+
+./create_default_recipe_durations
+"""
+
+import sqlalchemy as sa
+from sqlalchemy.orm import Session as OrmSession
+
+from zimfarm_backend import logger
+from zimfarm_backend.db import Session
+from zimfarm_backend.db.models import Recipe, RecipeDuration
+from zimfarm_backend.db.recipe import DEFAULT_RECIPE_DURATION
+
+
+def create_default_recipe_durations(session: OrmSession):
+    logger.info("Looking for recipes without a default duration")
+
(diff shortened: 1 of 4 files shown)


benoit74 (commented):
Script LGTM but looking again at the situation, I realize that the fact we had a recipe without a default duration caused the pipe to be partially blocked. The worker concerned by the task had this kind of log:

```
[2026-05-18 11:31:16,032: WARNING] poll failed with HTTP 404: {'success': False, 'message': 'No default duration found for recipe wp1_selection_885a94c04b69258c'}
```

I feel like we should fix this, a "bad" recipe / requested task should not block other tasks from starting. I introduced another requested task runnable on same worker (like with a default duration) and it did not started. Other tasks (zimit) not runnable on this worker started correctly. It is ok for the requested task to stay in the "TODO" forever since it is a "bad" one but we should not block other tasks from starting.

WDYT?

elfkuzco:
>  It is ok for the requested task to stay in the "TODO" forever since it is a "bad" one but we should not block other tasks from starting.

> WDYT?

Yeah, certainly should not block others from starting

benoit74:
> Yeah, certainly should not block others from starting

OK, then please fix this in this same PR

elfkuzco:
Added an additional filter function to weed out requested tasks that have a recipe that has no recipe duration. This way, the diagnosis still returns the current error while the worker polling never encounters this.

benoit74 (approved):
LGTM

PR #1908, opened 2026-05-18
Create and maintain all maps recipes automatically

benoit74:
- Create minimal config to maintain existing maps recipes automatically (before we add more)
- Add all missing countries

### DIFF
--- a/recipesauto/.gitignore
+++ b/recipesauto/.gitignore
@@ -1,1 +1,2 @@
 fcc_work
+src/recipesauto/maps_areas.json
--- a/recipesauto/src/recipesauto/entrypoint.py
+++ b/recipesauto/src/recipesauto/entrypoint.py
@@ -67,6 +67,7 @@ def prepare_context(raw_args: list[str]) -> None:
             "phet",
             "gutenberg",
             "stackexchange",
+            "maps",
         ],
     )
 
--- a/recipesauto/src/recipesauto/maps.json
+++ b/recipesauto/src/recipesauto/maps.json
@@ -1,872 +1,92 @@
 {
  "maps_en_all": {
   "resources": {
-   "cpu": 3,
    "disk": 322122547200,
    "memory": 68719476736
   },
-  "title": "World",
-  "description": "Full map, including roads and landmarks"
+  "title": "World"
  },
  "maps_en_africa": {
   "resources": {
-   "cpu": 3,
-   "disk": 241591910400,
-   "memory": 17179869184
-  },
-  "title": "Africa",
-  "description": "Full map, including roads and landmarks"
+   "disk": 241591910400
+  }
  },
  "maps_en_antarctica": {
   "resources": {
-   "cpu": 3,
-   "disk": 214748364800,
-   "memory": 17179869184
-  },
-  "title": "Antarctica",
-  "description": "Full map, including roads and landmarks"
+   "disk": 214748364800
+  }
  },
  "maps_en_asia": {
   "resources": {
-   "cpu": 3,
-   "disk": 241591910400,
-   "memory": 17179869184
-  },
-  "title": "Asia",
-  "description": "Full map, including roads and landmarks"
+   "disk": 241591910400
+  }
  },
  "maps_en_australia-and-oceania": {
-  "name": "maps_en_australia-oceania",
   "resources": {
-   "cpu": 3,
-   "disk": 214748364800,
-   "memory": 17179869184
+   "disk": 214748364800
   },
-  "title": "Australia and Oceania",
-  "description": "Full map, including roads and landmarks"
- },
- "maps_en_central-america": {
-  "resources": {
-   "cpu": 3,
-   "disk": 214748364800,
-   "memory": 17179869184
-  },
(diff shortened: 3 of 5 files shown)


benoit74:
@Popolechien @kelson42 this is the PR containing configuration which will be used to automatically create all `maps` recipes in the Zimfarm

I need your help in reviewing https://github.com/openzim/zimfarm/pull/1908/changes/7caa8e07503a9793b48e252a114d83045afee5eb#diff-2d0142ea6bf15d6deb75908cda5a0b72f154664f14c5bfac78f30b4301fbb076

I've automatically assigned title and description, and ensured title is less than 30 characters and description less than 80 (except for last entry, see below).

I need you to:
- double check title and description are OK for you
- speak-up if the ZIM name needs to be special: the "key" in the JSON file is associated with Geonames DB label ; it is going to be the ZIM name unless we specify a `name` key for the given country
- double check you do not spot a ZIM we do not want to create
- help me solve `maps_en_saint-helena-ascension-and-tristan-da-cunha` situation: title is too long, but it looks like it is very important to give same importance to all three island according to [wikipedia article](https://en.wikipedia.org/wiki/Saint_Helena,_Ascension_and_Tristan_da_Cunha) ; I have no clue how to solve this, suggestions welcomed

Note as well that with this I do not customize the default center of the map for any of the new countries (like I did for india for instance). I assume it is ok to display the center of the area as center of the map by default. Should this prove to be problematic, we can fix this in the aftermath. 

I also did not customized memory and disk usage. Will do it in the aftermath as well (once we have precise measurements about memory and disk usage).

Popolechien:
Few comments here:
- Will these files be flagged as `mul`, or just English?
- We seem to have several variations of the title and description pair (see e.g. Algeria vs. Malaysia). I would go for `Country` and `Full map, including roads and landmarks`
- `Map of GCC states` -> `Gulf States`
- `Maps of Indonesia country (with East Timor), including roads and landmarks` -> `Indonesia and Timor Leste` + `Full map, including roads and landmarks` (this also applies to the United States)
- Not too sure what to make of American Oceania
- `"Map of \u00c3\u008ele de Clipperton"` -> `Clipperton island`
- For Saint Helena etc. I would maybe go with `South Atlantic`? Not ideal, but then not a high-traffic destination I think.
- Speaking of which, I am suprised we do not have the Falklands & South Georgia, nor Madeira, nor Reunion island, nor most of the Carribeans.
- We should check if Lichtenstein is already included in Switzerland, in which case we probably do not need Lichtenstein. Ditto for Monaco/France, Man/UK,

benoit74:
> Map of GCC states -> Gulf States
Maps of Indonesia country (with East Timor), including roads and landmarks -> Indonesia and Timor Leste + Full map, including roads and landmarks (this also applies to the United States)
"Map of \u00c3\u008ele de Clipperton" -> Clipperton island
For Saint Helena etc. I would maybe go with South Atlantic? Not ideal, but then not a high-traffic destination I think.

OK

> Will these files be flagged as mul, or just English?

Same question again, only English for now, see https://github.com/openzim/maps/issues/23 ; will change with https://github.com/openzim/maps/issues/104

> We seem to have several variations of the title and description pair (see e.g. Algeria vs. Malaysia). I would go for Country and Full map, including roads and landmarks

I'm not sure to get. Logic so far has been: 
- use `Map of xxx` as title and `Entire country, including roads and landmarks` as description
- unless it doesn't fit it the title or there are multiple countries, if which case I had to be creative

Do you want to have use `Country` as title and `Full map, including roads and landmarks` as description in all ZIMs?

> Not too sure what to make of American Oceania

Simply not create the ZIM?

> Speaking of which, I am suprised we do not have the Falklands & South Georgia, nor Madeira, nor Reunion island, nor most of the Carribeans.

They are probably included in another ZIM. You can check with https://download.geofabrik.de/index.html (and remember we now select a "rectangular" area including the country, so there is a high chance these can be found in another ZIM (but not advertised and probably over-sized)

> We should check if Lichtenstein is already included in Switzerland, in which case we probably do not need Lichtenstein. Ditto for Monaco/France, Man/UK,

Quite confident it will be. But we can check before pushing to production anyway

Popolechien:
> Do you want to have use Country as title and Full map, including roads and landmarks as description in all ZIMs?

Yes, also because of consistency - seems this wording could apply to all.

I would skip American Oceania (wherever that is), but then what are the odds that someone there would actually need these maps to work offline. Pretty high compared to other places I would say. Is there any way that `Long description` could cover all islands (so that a simple search for "Marshall islands" would be able to connect the dots)?

benoit74:
I've fixed definition file, please double-check again.

> I would skip American Oceania (wherever that is), but then what are the odds that someone there would actually need these maps to work offline. Pretty high compared to other places I would say. 

Indeed ... I don't really have a solution besides "download the whole earth" or "download the whole continent" unless we find a reliable source describing these islands geographic limits.

> Is there any way that Long description could cover all islands (so that a simple search for "Marshall islands" would be able to connect the dots)?

I don't know if long description is used in readers search.

Popolechien:
LGTM

PR #1917, opened 2026-05-22
expose recipe id in response and accept id in route parameters

### LINKED ISSUE #1914: [API] make recipes accessible through their *id*
Currently, the API does not expose at all the recipe *id* (UUID actually), only recipe *name* is exposed. This is a problem because recipe *name* is not stable. It is deemed to be editable, and this happen quite "often".

In order to complete https://github.com/openzim/cms/issues/255, we will need a stable recipe identifier. 

We should hence start preparing the Zimfarm for this change:
- expose recipe *id* in API results
- accept both recipe *name* **and** *id* in API endpoints where currently only recipe name is accepted (we need to continue accepting recipe name at least for backward compatibility with existing "consumers" of Zimfarm APIs)

@elfkuzco this is ready to be developed, this has been discussed live with @kelson42

elfkuzco:
## Rationale
This API enhances the API by returning the recipe id in the endpoints that return a recipe details. Accordingly, it accepts the recipe id as a route parameter and not just the recipe name

This closes #1914

### DIFF
--- a/backend/maint-scripts/relaunch_failed_recipes.py
+++ b/backend/maint-scripts/relaunch_failed_recipes.py
@@ -95,7 +95,7 @@ def relaunch_failed_recipes(session: OrmSession, start_date: str):
 
         result = request_task(
             session=session,
-            recipe_name=recipe_name,
+            recipe_identifier=recipe_name,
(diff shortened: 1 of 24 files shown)


benoit74 (approved):
LGTM

PR #1918, opened 2026-05-22
send recipe id and name in cms notification

### LINKED ISSUE #1915: Send recipe name and ID in CMS notification
In preparation for https://github.com/openzim/cms/issues/255, we need to send recipe name and ID to the CMS.

Recipe name will only be used as a convenience for display since it is not a stable identifier.

Recipe ID will be used for API calls between CMS UI and Zimfarm API.

@elfkuzco please proceed with implementation, this has been discussed live with @kelson42

elfkuzco:
## Changes
- send recipe id and name in cms notification

This closes #1915


benoit74 (approved):
LGTM

Issue #1927, opened 2026-05-27
Unable to go back properly after error


kelson42:
I'm at https://farm.openzim.org/pipeline/704b0631-f4f9-4275-a2bb-34e44d55fbbd

<img width="2475" height="307" alt="Image" src="https://github.com/user-attachments/assets/e6cf43cc-05a1-4a20-ab31-23b44ede768a" />

I want to come back to the pipeline view (where I come from).

I click on the browser back button, but nothing happens, the same URL seems reloaded.

PR #1928, opened 2026-05-27
hide links for requested tasks

### LINKED ISSUE #1926: Task failing to fetch
Image

Clicking on the [first task](https://farm.openzim.org/pipeline/704b0631-f4f9-4275-a2bb-34e44d55fbbd):

Image

Getting error:

Image

elfkuzco:
## Rationale

https://github.com/openzim/zimfarm/pull/1854 introduced a common way for display status and timestamp information. However, it only includes a task link if a task id is supplied. This PR removes links for requested tasks which are shown in the todo tab of the pipeline view as they shouldn't have tasks.

This fixes #1926

### DIFF
--- a/frontend-ui/src/components/PipelineTable.vue
+++ b/frontend-ui/src/components/PipelineTable.vue
@@ -50,7 +50,7 @@
             :status="item.status"
             :timestamp="item.timestamp"
             :updated-at="item.updated_at"
-            :task-id="item.id"
+            :task-id="isRequestedTasks ? null : item.id"
           />
         </template>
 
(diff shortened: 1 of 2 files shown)


elfkuzco:
Added https://github.com/openzim/zimfarm/issues/1927 to the list as I couldn't reproduce by clicking the external link or trying in local environment.

benoit74 (approved):
LGTM ; if you cannot reproduce https://github.com/openzim/zimfarm/issues/1927 (I can't either), then we should not link it to this PR and assume this is solved since we have mostly no real clue since we do not reproduce the issue, even in production. Please rather mark the issue as "awaits-reprocase" and ask for more details about how the issue arise. There is a significant chance not more change is indeed needed, but we can't tell at this stage.

benoit74:
Looks promising, blocked by https://github.com/openzim/cms/pull/298#pullrequestreview-4378784754

PR #1929, opened 2026-05-28
Add maps platform

benoit74:
## Rationale

We need a new `maps` platform so that we do not start more than one `maps` task at a time on the Zimfarm, this causes problems of throttling / overwhelming on upstream.


elfkuzco (approved):
LGTM

PR #1936, opened 2026-06-01
fix issue with periodic scheduler skipping recipes without a most rec…

### LINKED ISSUE #1935: Periodic scheduler is not picking-up never scheduled tasks
See https://farm.openzim.org/recipes/100rabbits_en_all ; recipe is enabled, scheduled for quarterly runs, periodic scheduler should probably pick the recipe and start it. Unless I've forgotten we intentionally do not pick recipes which have never been scheduled for "safety".

@elfkuzco can you please investigate (and fix if it is obviously just a bug)

elfkuzco:
## Changes
- move logic to request tasks outside the block for checks concerning the most recent task of a recipe. This way, it fires unconditionally for recipes that don't have a most recent task and those whose most recent task pass the checks

This fixes #1935

### DIFF
--- a/backend/src/zimfarm_backend/utils/scheduling.py
+++ b/backend/src/zimfarm_backend/utils/scheduling.py
@@ -38,7 +38,7 @@ def request_tasks_using_recipe(session: OrmSession):
             continue  # manually has no data
 
         period_start = getnow() - datetime.timedelta(days=period_data["days"])
-        logger.debug(f"requesting for `{period}` recipes (before {period_start})")
+        logger.info(f"requesting for `{period}` recipes (before {period_start})")
 
         # find non-requested recipes which last run started before our period start
         for recipe in session.execute(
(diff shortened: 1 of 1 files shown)


backend/src/zimfarm_backend/utils/scheduling.py
```diff
@@ -58,40 +58,45 @@ def request_tasks_using_recipe(session: OrmSession):
                     )
                     > period_start
                 ):
+                    logger.warning(
```
benoit74: I don't think this deserve a warning, a debug is probably sufficient (otherwise we will fill background-tasks logs, isn't it?)

backend/src/zimfarm_backend/utils/scheduling.py
```diff
@@ -58,40 +58,45 @@ def request_tasks_using_recipe(session: OrmSession):
...
+                        f"{recipe.name} not requested because most_recent_task "
+                        f"{last_run.id} started recently"
+                    )
                     continue
+
                 # don't request a task if the most_recent_task is still running
                 if last_run and last_run.status not in TaskStatus.complete():
-                    logger.debug(
+                    logger.warning(
```
benoit74: Idem

benoit74 (changes requested):

benoit74 (approved):
LGTM

PR #1942, opened 2026-06-02
handle unicode characters in text to base64 conversion

### LINKED ISSUE #1940: Cannot input txt blob with accented characters correctly
See the article list at https://farm.openzim.org/recipes/wikipedia_fr_test/edit

Multiple issues:
- when I enter accented characters, they are garbage once saved
- if I fix them in the UI, UI refuses to save the change (like there is no change from its perspective)
- and/or if I add more text (like adding another article) while the error is still there, UI refuses to save changes
- when retrieving the file through the API, there is an encoding issue as well (mwoffliner does not achieves to get the characters correctly, see https://farm.openzim.org/pipeline/5f1fa213-8b92-45a9-9b62-f7235c819bf6/debug (`Ignoring articles without revisions: Orbite_h�liosynchrone, Sven-�ke_Johansson`)

elfkuzco:
## Changes
- fix unicode character encoding issue using [TextEncoder](https://developer.mozilla.org/en-US/docs/Web/API/TextEncoder) for utf-8 support
- move textToBase64 to shared module

This fixes #1940


benoit74 (approved):
LGTM

PR #1943, opened 2026-06-02
do not force remote mode if blob fails to load on local

### LINKED ISSUE #1941: It is hard to replace a failing remote blob URL with a local one
https://github.com/user-attachments/assets/f1cd2ba4-bfd2-458b-9961-01935f900638

Issues:
- it is impossible to select "Local" when "Remote" was selected and URL is still there
- if I clear the remote URL then I can switch to "Local" but it still displays an error and I have to manually delete the blob URL

Since anyway all these changes are not saved until recipe configuration is saved, I'm convinced we can simplify the flow here.

elfkuzco:
## Rationale

https://github.com/user-attachments/assets/162985ae-772e-46f6-a40b-94bffc14c0f2

## Changes
- remove logic to force user to remote mode if content failed to load in local mode
- set remote mode initial value based on value of `allowRemoteUrl`

This closes #1941

### DIFF
--- a/frontend-ui/src/components/BlobEditor.vue
+++ b/frontend-ui/src/components/BlobEditor.vue
@@ -80,7 +80,7 @@ import httpRequest from '@/utils/httpRequest'
 import { formattedBytesSize } from '@/utils/format'
 import constants from '@/constants'
 import type { Config } from '@/config'
-import { computed, ref, watch, inject } from 'vue'
+import { computed, ref, watch, inject, onMounted } from 'vue'
 import InlineTextEditor from '@/components/InlineTextEditor.vue'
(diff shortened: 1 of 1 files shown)


benoit74 (commented):
Doesn't this means that `articleList` will always show as "Remote" (because it allows remote URLs) even if user configured a "Local" content last time he configured the recipe? Shouldn't we compute this based on the fact that "Local" content contains the API base URL?

elfkuzco:
> Shouldn't we compute this based on the fact that "Local" content contains the API base URL?

That could work too

benoit74:
> That could work too

Please do so then

benoit74 (approved):
LGTM

PR #1950, opened 2026-06-04
Add dev instructions for restoring a DB dump

benoit74:
Same as CMS, but adapted to Zimfarm. Useful for devs who have access to prod DB dumps (all Zimfarm devs ATM).


elfkuzco (approved):
LGTM

PR #1955, opened 2026-06-11
add zim metadata to offliner spec

### LINKED ISSUE #1953: Associated ZIM metadata with offliner flag in definitions
In order to solve https://github.com/openzim/cms/issues/333, we need to store in the offliner definition which flag is supposed to alter which ZIM metadata.

E.g. as of today, ZIM `Name` comes from mwoffliner `filenamePrefix` flag, sotoki `name` flag, etc ...

This is an issue in Zimfarm but there is maybe little to modify in Zimfarm at this stage besides defining how we model this in the offliner definition schema and ensure it will be correctly reflected in API.

@elfkuzco can you propose something?

elfkuzco:
## Rationale
This PR introduces a way to associate zim metadata entries along with the flag that is used to generate it's value

This closes #1953

### DIFF
--- a/backend/maint-scripts/update_offliner_definitions_zim_metadata.py
+++ b/backend/maint-scripts/update_offliner_definitions_zim_metadata.py
@@ -1,0 +1,168 @@
+#!/usr/bin/env python3
+
+
+from sqlalchemy import select
+
+from zimfarm_backend import logger
+from zimfarm_backend.common.schemas.offliners.models import (
+    OfflinerSpecSchema,
+    ZimMetadata,
+)
+from zimfarm_backend.db import Session
+from zimfarm_backend.db.models import OfflinerDefinition
+from zimfarm_backend.db.offliner_definition import create_offliner_definition_schema
+
+zim_metadata_mapping: dict[str, list[ZimMetadata]] = {
+    "devdocs": [
+        ZimMetadata(metadata="Name", flag="name_format"),
+        ZimMetadata(metadata="Title", flag="title_format"),
+        ZimMetadata(metadata="Description", flag="description_format"),
+        ZimMetadata(metadata="LongDescription", flag="long_description_format"),
+        ZimMetadata(metadata="Creator", flag="creator"),
+        ZimMetadata(metadata="Publisher", flag="publisher"),
+        ZimMetadata(metadata="Illustration", flag="logo_format"),
+    ],
+    "freecodecamp": [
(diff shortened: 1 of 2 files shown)


elfkuzco:
@benoit74 , I just realize we can only use the same flag name as we define in the top level. This is because, for fields that don't define an explicit alias, the flag could be transformed by the base model type which isn't part of the `OfflinerSpec` schema. Either ways, with `getattr`, we can retrieve the value in zimfarm when we create an instance

benoit74 (changes requested):
Code LGTM but we need as well a maintenance script to update existing offliner definitions in DB so they all have this info. I think all versions of a given offliner can have the same zim_metadata definition (we did not introduced much changes to these definitions in scrapers since their introduction), but it is worth to double check.

elfkuzco:
Once this is approved, I will create the PRs to update the offliner definitions across the various repositories

backend/src/zimfarm_backend/common/schemas/offliners/models.py
```diff
@@ -121,6 +121,11 @@ class SimilarityDataSchema(CamelModel):
     transformers: list[TransformerSchema]
 
 
+class ZimMetadata(CamelModel):
+    field: str  # the name of the Metadata entry e.g Name, Language
```
benoit74: Sorry for realizing this a bit late (better late than too late ^^), but `field` is a bit obscure, it is hard to guess this is the metadata without seeing the other is the flag. What about `metadata` and `flag`? I feel like it is more self-explanatory.

backend/maint-scripts/update_offliner_definitions_zim_metadata.py
```diff
@@ -0,0 +1,168 @@
...
+        ZimMetadata(field="Creator", flag="creator"),
+        ZimMetadata(field="Publisher", flag="publisher"),
+        ZimMetadata(field="Language", flag="lang"),
+        ZimMetadata(field="Illustration", flag="favicon"),
+    ],
+    "maps": [
+        ZimMetadata(field="Name", flag="name"),
+        ZimMetadata(field="Title", flag="title"),
+        ZimMetadata(field="Description", flag="description"),
+        ZimMetadata(field="LongDescription", flag="long-description"),
```
benoit74: Shouldn't we fix these bad flag names before running this script which will make things only harder to fix?

benoit74 (approved):
LGTM

PR #1958, opened 2026-06-11
Rename pipeline 'Done' tab to 'Succeeded' for clarity

### LINKED ISSUE #1782: Pipeline Done tab is poorly named
https://farm.openzim.org/pipeline?filter=done is named Done in the UI. It actually lists successful recipe attempts. Failed are also over, and thus, "done".

You might rename it Successes in the UI. You might keep the URL though so that references to it don't break.

nexiouscaliver:
Fixes #1782

  ## Problem
  The pipeline view has a tab labeled `DONE` which is misleading because it only shows `succeeded` tasks. Failed tasks are also
  "done" in a general sense, making the label ambiguous.

  ## Solution
  Rename the tab label from `DONE` to `SUCCEEDED` to accurately reflect that it displays only successfully completed tasks. The
  underlying filter value remains `done` for backward compatibility with URL queries.

  ## Changes
  - `frontend-ui/src/views/PipelineView.vue` — changed label from `'DONE'` to `'SUCCEEDED'` (value unchanged)

  ## Testing
  - ESLint: passed
  - vue-tsc: passed
  - Filter logic: verified `case 'done'` switch and route queries unchanged

### DIFF
--- a/frontend-ui/src/views/PipelineView.vue
+++ b/frontend-ui/src/views/PipelineView.vue
@@ -47,7 +47,7 @@ const loadingStore = useLoadingStore()
 const filterOptions = [
   { value: 'todo', label: 'TODO' },
   { value: 'doing', label: 'DOING' },
-  { value: 'done', label: 'DONE' },
+  { value: 'done', label: 'SUCCEEDED' },
   { value: 'failed', label: 'FAILED' },
 ]
 // determine the headers based on the selected table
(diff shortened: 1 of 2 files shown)


nexiouscaliver:
Hi @benoit74 @elfkuzco — small UX clarity fix, would appreciate a quick review. Thanks!

benoit74:
Still not convinced about this change

kelson42:
Not convinced either. The improvment is not wrong but pretty much pointless IMHO. Closing.

PR #1973, opened 2026-06-26
use zimcheck urls if supplied by CMS

### LINKED ISSUE #1972: UI should retrieve zimcheck URL from CMS
This is the accompanying issue to https://github.com/openzim/cms/issues/377

UI should display zimcheck link coming from the CMS instead of assuming file is always there. And when link is not there, not show download button, just like for view ZIM and download ZIM buttons.

elfkuzco:
## Rationale
CMS is now in charge of deleting zimcheck results from S3 (https://github.com/openzim/cms/pull/355). So, when zimfarm is configured to inform CMS, it needs to rely on CMS to know if zimcheck urls are still valid.

## Changes
- use zimcheck URL provided by CMS if zimfarm is configured to inform CMS, otherwise, default to old behaviour of building the URL using zimcheck filename

This closes #1972

### DIFF
--- a/backend/src/zimfarm_backend/common/schemas/orms.py
+++ b/backend/src/zimfarm_backend/common/schemas/orms.py
@@ -105,7 +105,7 @@ class TaskLightSchema(BaseTaskSchema):
 class ZimUrlSchema(BaseModel):
     """Schema for a single zim URL"""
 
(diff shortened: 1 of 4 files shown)


benoit74 (approved):
LGTM

PR #1979, opened 2026-07-20
use loaded recipe name in view over recipeName props

### LINKED ISSUE #1977: Accessing recipe through their UUID at UI-level is a bit sluggish
See e.g. https://farm.openzim.org/recipes/5c397e42-9611-4c75-9dbb-50bd9cd51d0b

Recipe name should be displayed instead of "5c397e42-9611-4c75-9dbb-50bd9cd51d0b" UUID

Image

elfkuzco:
## Changes
- use loaded recipe name in view instead of `recipeName` passed via props/route as that can be a UUID

This fixes #1977

### DIFF
--- a/frontend-ui/src/views/RecipeDetailView.vue
+++ b/frontend-ui/src/views/RecipeDetailView.vue
@@ -23,10 +23,10 @@
     </v-row>
 
     <!-- Title Row -->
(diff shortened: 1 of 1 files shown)


benoit74 (approved):
LGTM

PR #1980, opened 2026-07-20
exclude archived recipes from scheduling

### LINKED ISSUE #1978: Archived recipes should never be considered for periodic scheduling
Whenever periodic scheduling triggers, it calls the `request_tasks_using_recipe` function which ignores disabled recipes but still considers archived ones.

This causes many time this kind of message to appear:

```
[2026-07-16 14:53:41,363: WARNING] Could not request task due to the following reason: Recipe 'wikinews_sr' is archived 
```

This message is only the consequence of the root issue: archived recipes should not be considered for periodic scheduling.

elfkuzco:
## Changes
- exclude archived recipes from periodic scheduler list of recipes to request tasks for

This fixes #1978


benoit74 (approved):
LGTM

PR #1999, opened 2026-08-21
change message for new users with missing profile scope

### LINKED ISSUE #1986: Why do we claim that `profile` scope is missing?
See https://github.com/openzim/wp1/issues/1228

Looks like when the Zimfarm is presented a JWT which is valid but not tied to a specific user, then for some reason it answers that `profile` scope is missing. 

This is very misleading since AFAIK we do not use scopes at all.

Do we achieve to reproduce the problem? Is this something we can fix? Is it related to the fact we support many auth mode in Zimfarm (session vs full oauth)?

elfkuzco:
## Changes
- change message for new users with missing profile scope

This closes #1986


benoit74 (approved):
LGTM

PR #2001, opened 2026-08-21
omit no-op changes in recipe update confirmation

### LINKED ISSUE #2000: Weird changes displayed in recipe update confirmation
Sample recipe: https://farm.openzim.org/recipes/wikipedia_eu_top

When I do some changes, I'm often presented with unexpected changes like that

Image

While doing the version comparison afterwards there is only one change:

Image

https://farm.openzim.org/recipes/wikipedia_eu_top/history?compare=2e5e2f3a-5f5f-4e60-9f78-76bf3d6c18b3...c3d403a8-2475-4dca-b24b-73d70a5a11d3

This has no concrete impact on ZIM production but makes confirmation way more complex in many cases.

elfkuzco:
## Changes
- extract anonymous function to filter out no-op changes into a named function and reuse in changes computation and diff rendering in update confirmation

This fixes #2000

### DIFF
--- a/frontend-ui/src/components/RecipeEditor.vue
+++ b/frontend-ui/src/components/RecipeEditor.vue
@@ -945,6 +945,32 @@ const areAllFieldsValid = computed(() => {
   return true
 })
 
+// Helper to filter out no-op differences (e.g. null -> "") that shouldn't be
+// considered changes when comparing flags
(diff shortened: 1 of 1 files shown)


benoit74 (approved):
LGTM

PR #2004, opened 2026-08-26
respect http schemes in s3 upload urls

elfkuzco:
## Rationale
CMS dev environment now interacts end to end with zimfarm (https://github.com/openzim/cms/pull/469) but it sets it's s3 urls for zimcheck to start with `s3+http` because it's local minio cannot be accessed via https. This PR enhances the logic to transform upload urls to respect the underlying scheme of the upload URL. This prevents http errors that would arise trying when trying to download zimcheck files

## Changes
- respect http schems in s3 urls
- update zimcheck results upload url env in background tasks service (docker-compose)


benoit74 (approved):
LGTM

PR #2006, opened 2026-08-27
filter out recipes/requested tasks in update_scraper_version maint-script

### LINKED ISSUE #2005: Maintenance script to migrate from mwoffliner 1.x to mwoffliner 2.x
We are about to release mwoffliner 2.x ; while the maintenance script we already have to upgrade offliners is very convenient, it does not handle major upgrades where some parameters are renamed, and others are completely dropped.

For this migration of mwoffliner 1.x to 2.x, we need to:
- transfer `articleList` value to `pageList`
- transfer `articleListToIgnore` value to `pageListToIgnore`
- rework `speed` values: on all wikimedia recipes (we can track them with their context), set `speed` to 10 ; for others, remove `speed` values (tbc, can you please extract the list of recipes with non-default speed values to let me confirm this ; could be a simple SQL query I will run a production)
- transfer `filenamePrefix` value to `customZimName`
- transfer `keepEmptyParagraphs` value to `keepEmptySections`
- drop `mwWikiPath`, `mwRestApiPath`, `mwIndexPhpPath` values: they are not used anymore
- drop `verbose` value: it has been replaced by `logLevel` whose default value is OK
- drop `forceRender` and `osTmpDir` values: their default values are OK

Important: some recipes are already using `dev` offliner-definition, no need to upgrade them again obviously

@elfkuzco can you prepare this? I'm really not sure it is worse to develop a generic format to cope with these changes, maybe something very specific to this move is enough (could even be an "if" branch in current maint script where some more things are done)

elfkuzco:
## Rationale
This PR enhances the maint-script to update scraper versions for offliner/image tag by filtering out recipes/requested tasks based on contexts, offliner version. In addition, it supports a dry run mode that writes the output to `update_scraper_version.md`

## Changes
- manually call `session.commit` for each updated recipe/requested task
- added flags to include/exclude contexts and offliner definitions.
- added `-s` flag to set the value of flag(s)
This closes #2005

### DIFF
--- a/backend/maint-scripts/update_scraper_version.py
+++ b/backend/maint-scripts/update_scraper_version.py
@@ -1,8 +1,8 @@
 #!/usr/bin/env python3
 
 """
-Script to update recipes and requested tasks with new image tags or offliner
-definitions.
+Script to update recipes and requested tasks. Use the -h flag to see the script
+description.
 
 EXAMPLES:
 
@@ -22,7 +22,7 @@ top level key that houses it's specification. For example, if the offliner defin
 contains a new field like:
 
     "article_list_to_ignore": {
-        "type": "string",
+    "type": "string",
         "required": false,
         "title": "Article List to ignore",
         "description": "List of articles to ignore"
@@ -43,27 +43,41 @@ database. If you must use the -n flag, here are some examples:
 5. Rename a field in recipe and requested tasks flags:
    ./update_scraper_version.py -o mwoffliner -d 2.1.0 -n "old_name:new_name"
 
-6. Multiple name mappings (add, remove, and rename in one command):
+6. Change flags names(add, remove, and rename in one command):
    ./update_scraper_version.py -o mwoffliner -d 2.1.0 \
-     -n "+:new_field,old_field:-,old_name:new_name"
+     -m "+:new_field,old_field:-,old_name:new_name"
+
+7. Set the value of an existing flag:
+   ./update_scraper_version.py -o mwoffliner \
+     -s "articleList=https://example.com/list.tsv"
 """
 
 import argparse
-from typing import cast
+import json
+from copy import deepcopy
+from pathlib import Path
+from typing import Any, cast
 
 import sqlalchemy as sa
+from pydantic.alias_generators import to_camel
+from sqlalchemy import and_
 from sqlalchemy.orm import Session as OrmSession
 from sqlalchemy.orm.attributes import flag_modified
 
 from zimfarm_backend import logger
+from zimfarm_backend.common.constants import parse_bool
 from zimfarm_backend.common.schemas.models import (
     DockerImageSchema,
     RecipeConfigSchema,
 )
+from zimfarm_backend.common.schemas.offliners.models import (
+    FlagSchema,
+    OfflinerSpecSchema,
+)
 from zimfarm_backend.common.schemas.orms import OfflinerDefinitionSchema, OfflinerSchema
 from zimfarm_backend.db import Session
 from zimfarm_backend.db.account import get_account_by_username
-from zimfarm_backend.db.models import Recipe, RequestedTask
+from zimfarm_backend.db.models import OfflinerDefinition, Recipe, RequestedTask
 from zimfarm_backend.db.offliner import get_offliner
 from zimfarm_backend.db.offliner_definition import (
     create_offliner_instance,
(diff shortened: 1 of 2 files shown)


elfkuzco:
I added the mwoffliner 2.0.0 version to my DB and ran the following command:

```sh
./update_scraper_version.py -o mwoffliner -d 2.0.0 -n "articleList:pageList,articleListToIgnore:pageListToIgnore,filenamePrefix:customZimName,keepEmptyParagraphs:keepEmtpySections,forceRender:-,osTmpDir:-,verbose:-,mwWikiPath:-,mwRestApiPath:-,mwIndexPhpPath:-" --dry-run -e dev
```

This drops some fields and renames others according to https://github.com/openzim/zimfarm/issues/2005#issue-5267816565

Here's a truncated output generated in dry run:
# Scraper version update report (dry run)

- **Offliner:** `mwoffliner`
- **Offliner definition version:** `2.0.0`
- **Name mappings:** `articleList` → `pageList`, `articleListToIgnore` → `pageListToIgnore`, `filenamePrefix` → `customZimName`, `keepEmptyParagraphs` → `keepEmtpySections`, remove `forceRender`, remove `osTmpDir`, remove `verbose`, remove `mwWikiPath`, remove `mwRestApiPath`, remove `mwIndexPhpPath`

## Summary

- **Recipes affected:** 1194
- **Requested tasks affected:** 0

## Recipes

### `wikipedia_ar_medicine_app`

- `offliner_definition.version`: `1.17.5` → `2.0.0`
- `offliner.addContentModels`: added `None`
- `offliner.addModules`: added `None`
- `offliner.articleList`: removed `http://download.openzim.org/wp1/arwiki/customs/medicine.tsv`
- `offliner.articleListToIgnore`: removed `None`
- `offliner.categoriesPageSize`: added `None`
- `offliner.customCss`: added `None`
- `offliner.customJs`: added `None`
- `offliner.customZimFilename`: added `None`
- `offliner.customZimName`: added `wikipedia_ar_medicine-app`
- `offliner.filenamePrefix`: removed `wikipedia_ar_medicine-app`
- `offliner.forceRender`: `ActionParse` → `None`
- `offliner.javaScript`: added `None`
- `offliner.keepEmptyParagraphs`: removed `None`
- `offliner.keepEmptySections`: added `None`
- `offliner.keepEmtpySections`: added `None`
- `offliner.logLevel`: added `None`
- `offliner.mathJaxAllPages`: added `None`
- `offliner.mathJaxConfig`: added `None`
- `offliner.mathJaxEntryPoint`: added `None`
- `offliner.mathJaxSource`: added `None`
- `offliner.mwIndexPhpPath`: removed `None`
- `offliner.mwRestApiPath`: removed `None`
- `offliner.mwWikiPath`: removed `None`
- `offliner.onlyNamespaces`: added `None`
- `offliner.osTmpDir`: `/dev/shm` → `None`
- `offliner.pageList`: added `http://download.openzim.org/wp1/arwiki/customs/medicine.tsv`
- `offliner.pageListToIgnore`: added `None`
- `offliner.userAgent`: added `None`
- `offliner.verbose`: removed `log`

elfkuzco:
I just realize I forget flag to update an existing flag's value needed to rework the speed

elfkuzco:
added flag `-s` to set a flag value and updated readme description

benoit74 (commented):
Do I get it correctly that `--offliner-definition-version` is used to update recipes to use this offliner definition while `--exclude-offliner-definition-versions` is used to select which recipes we wanna update?

If yes, then this situation is confusing ; both names are too close IMHO, or it's the fact we cannot decide which recipes to include by offliner definition version.

I feel like we should rename long-form arguments to something clearer in all arguments: use `--set-xxx-to` form for arguments which decide what config to set, `--include-xxx` form for arguments which decide which recipes to include and `--exclude-xxx` form for arguments which decide which recipes to not include?

What if we add support for `--include-offliner-definition-versions` and rename `--contexts` to `--include-contexts`? This would be significantly clearer IMHO.

backend/maint-scripts/update_scraper_version.py
```diff
@@ -244,9 +659,39 @@ def update_requested_tasks(
...
-    with Session.begin() as session:
+    # Ensure there's no overlap between include and exclude options
+    if (
+        args.offliner_definition_version
+        and args.exclude_offliner_definition_versions
+        and args.offliner_definition_version
+        in args.exclude_offliner_definition_versions
+    ):
+        raise ValueError(
+            "Overlap between offliner definition and excluded offliner definition "
```
benoit74: I don't get why this is a problem. Why can't I say that I want to update all recipes which are not using `x.y.z` offliner versions to use this offliner version for instance and changes their config to something else? This look like quite a typical use-case IMHO.

elfkuzco: If you say you want to exclude `x.y.z`, then, it wouldn't be the same version that you are trying to update it's recipes. For example, it would be contradictory to run
```sh
./update_scraper_version.py -o mwoffliner -d 2.0.0 -e 2.0.0,dev
```
You can't exclude the same version you are trying to apply the updates to. The `-d` flag supports only one version and it unconditionally updates all recipes to that version but if you say you want to exclude a specific version, then, recipes matching that version won't be updated. For example, if i want to update all recipes to `2.0.0` except for those running on `dev`, I would use
```sh
./update_scraper_version.py -o mwoffliner -d 2.0.0 -e dev
```

The `-e` flag for exclusion supports multiple versions to exclude

elfkuzco:
> Do I get it correctly that --offliner-definition-version is used to update recipes to use this offliner definition while --exclude-offliner-definition-versions is used to select which recipes we wanna update?

No. The behaviour of the script without these exclusion flags is to change the flags and offliner version of  existing recipes and any tasks yet to be processed. The `--offliner-definition-version` is where you specify the version you want to change the recipes to. If you don't specify any, then, any action would default to the recipe's own definition.

What these `--excude-offliner-definition-versions` does is that says to apply the original logic but not to any recipes belonging in the excluded versions. For example, if i want to apply actions to change recipes to `2.0.0` but exclude recipes using the `dev` version

```sh
./update_scraper_version.py -o mwoffliner --offliner-definition-version=2.0.0 --exclude-offliner-definition-versions=dev
```

Without the `--excluded-offliner-definition-versions=dev`, all `mowffliner` recipes would be unconditionally updated to `2.0.0`, even those running on `dev`

elfkuzco:
> I feel like we should rename long-form arguments to something clearer in all arguments: use --set-xxx-to form for arguments which decide what config to set, --include-xxx form for arguments which decide which recipes to include and --exclude-xxx form for arguments which decide which recipes to not include?

> What if we add support for --include-offliner-definition-versions and rename --contexts to --include-contexts? This would be significantly clearer IMHO.

Yeah, I agree with you that they are clearer

elfkuzco:
Renamed the flags to the forms requested and updated description

backend/maint-scripts/update_scraper_version.py
benoit74: What if if want to update only recipes not yet on `2.0.0` to use `2.0.0` (rather than "forcing" all recipes to `2.0.0` even if they are already using `2.0.0`)?

benoit74 (commented):
LGTM except the pending conversation I've just updated

backend/maint-scripts/update_scraper_version.py
elfkuzco: We could pass the other versions in the `-e` or `--exclude-offliner-definition-versions` flag

```sh
./update_scraper_version.py -o mwoffliner -d 2.0.0 -e dev,2.0.1,2.02
```

Admittedly, it might be long to pass all the versions. Unless we want to support regex-like expressions in exclusion flags

benoit74: I admit we rarely have more than 2 versions in use, you are right. 

Correct command for my use case is rather using `-i`, not `-e` ; and it typically consist in updating previous versions, not next ones:

```
./update_scraper_version.py -o mwoffliner -d 2.0.0 -i dev,1.17.5
```

But is there a fundamental problem if we remove the check all let users call the tool with `-d 2.0.0 -e 2.0.0` (which will do the same than `-d 2.0.0 -i dev,1.17.5` on current DB)?

elfkuzco: The check for overlap no longer happens between the version we want to update to and the exclusion flags. It now works between the `--include-xxx` and `--exclude-xxx` flags. So, `-d 2.0.0 -e 2.0.0` should still work

https://github.com/openzim/zimfarm/blob/098aa47dfae93f36873a7a8a9e265263757d0d22/backend/maint-scripts/update_scraper_version.py#L745-L763

benoit74: Oh indeed, sorry, I missed that. And this check indeed makes full sense

benoit74 (approved):

PR #2011, opened 2026-08-31
add documentation for offliner-definition.json file

### LINKED ISSUE #2009: Document `offliner-definition.json` format
Unless I forgot about it, we do not have a "formal" documentation around `offliner-definition.json` file format. 

Now that the format has mostly stabilized, because it is also getting more complex (e.g with new "zimMetadata" field) and also because we have more scraper contributors needing to modify it, I feel like we should have such a documentation.

Recording this issue as an enhancement because it was not asked for in the first place and we could live without it technically speaking, but this is mostly a bug from organizational perspective ^^

elfkuzco:
## Changes
- add documentation in integrators' guide on the offliner defintion file from scraper repositories

This closes #2009


benoit74 (approved):
LGTM ; this makes me realize there are already a ton of features supported ^^

PR #2013, opened 2026-09-01
Rework slightly OAuth OIDC

benoit74:
## Rationale

Adapt many details to our new understanding and configuration of OAuth OIDC flows.

## Changes

- see https://github.com/kiwix/overview/issues/175
- backend:
  - always create an account when a JWT arrives with an unknown `sub` (even for machines): when `name` is unknown, simply use `sub` as display name until it gets updated by a Zimfarm admin
  - clarify variable namings
    - `OAUTH_OIDC_CLIENT_ID` => `OAUTH_OIDC_AUDIENCE` (we use it only to check audience in the JWT)
    - `OAUTH_SESSION_AUDIENCE_ID` => `OAUTH_SESSION_AUDIENCE` (audience ID is meaningless, it is an audience and a client ID ; choosing audience for clarity)
    - `CREATE_NEW_OAUTH_ACCOUNT` => `OAUTH_CREATE_NEW_ACCOUNT`
  - share code in both OAuth kinds to detect humans vs machine when receiving a JWT
  - OIDC: source aal and name from new `ext.kiwix-aal` and `ext.kiwix-name` claims
- frontend: 
  - OIDC: 
    - use the `access_token` instead of the `id_token` (save it in local storage + send it to the API)
    - when initiating the auth flow at `/oauth2/auth`, specify the `audience` (backend client ID) the access_token is intended for
    - add a new setting to allow the distinction between `OAUTH_CLIENT_ID`(the UI client ID) from `OAUTH_AUDIENCE` (the backend client ID)
- update integrator guide to clarify what needs to be configured in backend/frontend regarding OAuth

### DIFF
--- a/INTEGRATORS_GUIDE.md
+++ b/INTEGRATORS_GUIDE.md
@@ -113,7 +113,6 @@ be injected at runtime so the browser can read it. This is typically done via th
 - `ZIMFARM_ZIM_DOWNLOAD_URL`: The URL of the server where the ZIMs are stored
 - `MONITORING_URL`: The URL of the [Netdata server](https://www.netdata.cloud/) where
   monitoring statistics are stored
-- `OAUTH_CLIENT_ID`: Oauth client ID for authentication (currently backed by [Ory.sh](https://www.ory.com/))
 
 See the `dev/frontend-ui/public/config.json` for some reasonable defaults of these variables.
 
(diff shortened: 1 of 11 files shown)


benoit74:
@elfkuzco please have a look and do not hesitate to speak-up if something worries you.

The root of this change comes from reasons exposed in https://github.com/kiwix/overview/issues/175

I made the modifications myself because there was too many moving pieces and uncertainties, needing quick loops of modifications between bits you did not had access to (the new token webhook and Ory config).

I will push token webhook code and update the wiki in coming hour

elfkuzco (approved):
LGTM

PR #2018, opened 2026-09-03
pass token to logout call in oidc auth provider

### LINKED ISSUE #2016: Frontend sign-out is not working when using OIDC mode
Currently farm.zimit.kiwix.org and farm.openzim.org are configured to use OIDC mode.

The sign-out is not really working, because as soon as I refresh the page I'm signed-in again.

This is very probably linked to the fact that Ory session is not invalidated.

I do not recall the discussions we had on this matter, but this is at least very confusing for the user and we need to enhance the situation

@elfkuzco can you help me find (or simply recall) the discussions we had on that matter?

elfkuzco:
## Changes
- pass stored token to logout call to OAuthOIDCProvider class

This fixes #2016

### DIFF
--- a/frontend-ui/src/App.vue
+++ b/frontend-ui/src/App.vue
@@ -78,8 +78,8 @@ const navigationItems = computed<NavigationItem[]>(() => [
   },
 ])
 
-const handleSignOut = () => {
-  authStore.logout()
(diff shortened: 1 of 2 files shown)


benoit74 (approved):
LGTM

PR #2022, opened 2026-09-04
fix issue with OIDC token refresh

### LINKED ISSUE #2020: OIDC: Refreshing access token with refresh token is broken
Refreshing acces token is currently broken in Zimfarms.

Fixing this issue is a top priority.

I'm probably partially the culprit (due to my recent changes to OIDC flows), but I seek @elfkuzco help to reproduce the issue in staging and fix it. 

It looks like:
- the UI correctly detects the access token is about to expire
- it calls Ory to renew the token and Ory issues a new one
- it correctly stores and uses this access token in subsequent calls
- but the API refuses this new token
- looking inside the token, I see that
   - it misses custom extensions "kiwix-aal" and "kiwix-name" (the code populating these lives here: https://github.com/kiwix/container-images/tree/main/login-token-webhook ; I suspect it just misses sufficient info)
   - the audience is incorrect (it is frontend audience instead of backend API audience)

I strongly suspect that the call to refresh the token should at least have:
- proper audience passed
- proper scope passed

FYI, I've switched **staging** token webhook into DEBUG log level so that we can have details about the incoming payloads received ; do not hesitate to ask for these details if passing proper audience and scope are not sufficient.

elfkuzco:
## Changes
- use config audience ID while refreshing token
- store access_token instead of id_token while saving token

This fixes #2020

### DIFF
--- a/frontend-ui/src/services/auth/OAuthOIDCProvider.ts
+++ b/frontend-ui/src/services/auth/OAuthOIDCProvider.ts
@@ -96,6 +96,8 @@ export class OAuthOIDCProvider extends AuthProvider {
       grant_type: 'refresh_token',
       client_id: this.config.clientId,
       refresh_token: refreshToken,
+      audience: this.config.audience,
+      scope: 'openid profile offline_access',
     })
 
     const response = await service.post<URLSearchParams, OAuthOIDCTokenResponse>('', params, {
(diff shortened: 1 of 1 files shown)


benoit74 (approved):
LGTM, thank you (and nice catches the remaining id token ...)

PR #2025, opened 2026-09-11
make recipe flags dependent on enum choice

### LINKED ISSUE #2010: Add support for "conditional recipe flags"
For now, all recipe flags defined in `offliner-definition.json` are always shown in the UI.

We do have scrapers where some flag are meaningless when another flag is set to some value.

For instance in gutenberg we have (or will have soon more exactly) the `source` flag which is an enum specifying which kind of source the scraper is processing, and some other flags are specific to each source.

It would help Zimfarm editor if unrelevant flags could be automatically hidden and cleared when the `source` flag is changed.

This dependency must of-course be defined in `offliner-definition`.

elfkuzco:
## Rationale
This PR enhances the offliner definition to make some flags be conditionally shown when an enum has a specific value

https://github.com/user-attachments/assets/74280566-be71-445f-a609-c1df8ae644cf

## Changes
- add support for enum types to optionally control visibility and value of other flags. only available for `string-enum` types as it would be ambiguous to enforce for `list-of-string-enum` types
- restrict choices to be only set for `enum` and `list-of-string-enum` types
- when an enum with dependents is selected, unset values of other dependents

This closes #2010


benoit74 (changes requested):
LGTM, please update integrator documentation

benoit74 (approved):
LGTM

Issue #2026, opened 2026-09-13
Missing icons (for Apple WebApp)


kelson42:
I have created a shortcut on homescreen of my iPad.

But the Zimfarm icon does not display properly

<img width="390" height="571" alt="Image" src="https://github.com/user-attachments/assets/5da8e658-7623-406d-bdc4-799ccc56684f" />

PR #2030, opened 2026-09-18
fix: remove invalid Apple touch icon reference

Dhritimann:
## Rationale

The Apple WebApp icon configuration included an unsized `apple-touch-icon.png` reference pointing to a 57×57 image.

The page already provides explicit size-specific Apple touch icons, including the 180×180 icon. Since iOS can select the most appropriate icon when multiple sizes are provided, the legacy unsized reference to the 57×57 asset is redundant.

This change removes the legacy fallback while keeping the existing size-specific Apple touch icon declarations unchanged.

Issue: #2026

## Changes

- Removed the unsized `apple-touch-icon.png` reference from `frontend-ui/index.html`.
- Kept the existing size-specific Apple touch icon declarations, including the 180×180 icon.


benoit74:
@Dhritimann can you explain why this unsized icon is problematic with more sources? 

This was needed for iOS 6 and before according to https://realfavicongenerator.net ; and I don't get how this prevent Firefox on iOS to find an icon ; at worst it would cause it to select wrong resolution, but I would be surprised it causes it to just use no icon at all.

Dhritimann:
@benoit74  Thanks for pointing this out. I think you're right that my original rationale was a bit too strong.

The 57×57 icon was used for older iOS versions, and the apple-touch-icon.png fallback itself is valid. My intention here was mainly to remove the old unsized reference since the page already has the explicit size-specific icons, including the 180×180 one.
I've updated the PR description to clarify this and removed the claim about it causing the icon to disappear or be selected incorrectly. Thanks for catching that.

benoit74:
@Dhritimann then it is very likely it will not fix https://github.com/openzim/zimfarm/issues/2026 right?

If you just wanna clean things up, I recommend you run https://realfavicongenerator.net/ again on the SVG icon we have. There are many other files to remove to follow 2026 baseline.

